Issue #26, opened 2016-09-09
PEP8 Compliance


t8y8:
We're actually pretty close!

```
./setup.py:23:2: W292 no newline at end of file
./samples/explore_datasource.py:34:1: E266 too many leading '#' for block comment
./samples/explore_datasource.py:76:1: W391 blank line at end of file
./samples/explore_workbook.py:35:1: E266 too many leading '#' for block comment
./tableauserverclient/namespace.py:1:44: W292 no newline at end of file
./tableauserverclient/models/datasource_item.py:116:29: E128 continuation line under-indented for visual indent
./tableauserverclient/models/exceptions.py:2:9: W292 no newline at end of file
./tableauserverclient/models/user_item.py:147:29: E128 continuation line under-indented for visual indent
./tableauserverclient/models/workbook_item.py:166:29: E128 continuation line under-indented for visual indent
./tableauserverclient/server/exceptions.py:2:9: W292 no newline at end of file
./tableauserverclient/server/endpoint/projects_endpoint.py:55:1: W391 blank line at end of file
./test/test_request_option.py:99:1: W391 blank line at end of file
./test/test_view.py:65:112: W292 no newline at end of file
```

PR #27, opened 2016-09-10
PEP8 Compliance and TravisCI Updates

t8y8:
A few very small tweaks to bring us into pep8 compliance, and then enable pycodestyle in travisci.

This addresses #26


RussTheAerialist:
:rocket:

Issue #25, opened 2016-09-09
Design Proposal: Change pagination_items in returns from lists


t8y8:
All of the `model.get()` calls return `PaginationItem, List[<model>]` which means there's a lot of 

```
_, all_workbooks  = server.workbooks.get()
```

I propose that we either:
1) Don't return the PaginationItem because you, the caller, already know your pagination parameters, since you applied them when you issued the 'get' AND you can't do anything interesting with them after the fact (the most interesting item, the total number, is available idiomatically as `len(all_workbooks)`
2) Switch the order of the items in the return to that it's (List[<model>], PaginationItem)

LGraber:
You can't remove them. By default, you get 100 items per page and without this return value you won't know if you got all of the items / how many items there are. The only way to "get rid of it" would be if we didn't return a list but instead some more complicated object which, as you called get-next eventually asked for the next page as needed. This, however, is something I would build on top of this low level api.

LGraber:
to clarify, total is total_available, not how many were returned so it does not necessarily match len(all_workbooks)

t8y8:
Thanks for your reply.

I didn't realize that .get doesn't get all resources in the collection by default!

I still think it's very weird to expose this implementation detail do the high-level get call. 
When 99% of the time folks are ignoring a return value maybe we don't need to return it.

But your comments make me wonder if there's a different route for me to propose -- should `.get` by default get all resources. That's a different issue that I can file.

For now, what about just reversing the return order?

`all_workbooks, _  = server.workbooks.get()`

t8y8:
I had an idea for a decorator @paginated that we wrap the `get` calls with, it'll check for `get_all=True` and just loop through collecting all results and returning them. 

It would leverage the pagination_item that's currently being returned.

t8y8:
And I've got a working decorator.
It obeys the request options if you specify them, else it just loops at the default size.

I can figure out how to pass in a flag to do this by default but disabled if you'd like in another revision. I'll get some working code up for a PR with a test but I need to learn how to use the test framework first.

Here's a preview

```
def paginate(get_call):
    from math import ceil
    from ..request_options import RequestOptions

    def wrapper(self, opts):

        all_things = []

        # First call
        print("FIRST CALL")
        res, PI = get_call(self, opts)
        all_things.extend(res)
        print("ALL THINGS HAS: " + str(len(all_things)))

        # Build loop Conditions
        number_of_workbooks = PI.total_available
        pageSize = PI.page_size
[...]

PR #28, opened 2016-09-10
PROPOSAL: Remove PaginationItem from model.get() calls.

t8y8:
Updated tests to match. 

I had to skip 3 that specifically test the pagination item and I'm not sure how to verify that information now. I'm also not sure it matters at this level -- it's mostly an implementation detail of how many API requests get made and that's it.

Looking for discussion.

One approach to addressing #25


t8y8:
Killing this one dead

PR #29, opened 2016-09-12
reversed the order of return variables for all .get() calls

shinchris:
Prepping for release: Pagination item is now returned as the second variable and Travis passes


RussTheAerialist:
:rocket:

PR #30, opened 2016-09-13
added support for schedules

shinchris:


test/test_schedule.py
```diff
@@ -0,0 +1,63 @@
...
+        self.assertEqual("Saturday night", all_schedules[1].name)
+        self.assertEqual("Active", all_schedules[1].state)
+        self.assertEqual(80, all_schedules[1].priority)
+        self.assertEqual("2016-07-07T20:19:00Z", all_schedules[1].created_at)
+        self.assertEqual("2016-09-12T16:39:38Z", all_schedules[1].updated_at)
+        self.assertEqual("Subscription", all_schedules[1].type)
+        self.assertEqual("Weekly", all_schedules[1].frequency)
+        self.assertEqual("2016-09-18T06:00:00Z", all_schedules[1].next_run_at)
+
+    def test_get_empty(self):
```
t8y8: Do we test this for every content type? 

This is harmless -- but ultimately what's it testing? 

I guess that the Schedule.from_response doesn't choke on an empty input?

t8y8:
:rocket:

RussTheAerialist:
:rocket:

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self):
+        self._created_at = None
+        self._end_schedule_at = None
+        self._frequency = None
+        self._id = None
+        self._name = None
+        self._next_run_at = None
+        self._priority = None
+        self._state = None
+        self._type = None
```
LGraber: What is type. This doesn't feel like it should be just an arbitrary string

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,79 @@
...
+class ScheduleItem(object):
+    def __init__(self):
+        self._created_at = None
+        self._end_schedule_at = None
+        self._frequency = None
+        self._id = None
+        self._name = None
+        self._next_run_at = None
+        self._priority = None
+        self._state = None
```
LGraber: Same here (what is state)

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,79 @@
+import xml.etree.ElementTree as ET
+from .. import NAMESPACE
+
+
+class ScheduleItem(object):
+    def __init__(self):
+        self._created_at = None
+        self._end_schedule_at = None
+        self._frequency = None
```
LGraber: Same here (what is frequency)

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,79 @@
...
+            schedule_item = cls()
+            schedule_item._id = schedule_xml.get('id', None)
+            schedule_item._name = schedule_xml.get('name', None)
+            schedule_item._state = schedule_xml.get('state', None)
+            schedule_item._created_at = schedule_xml.get('createdAt', None)
+            schedule_item._updated_at = schedule_xml.get('updatedAt', None)
+            schedule_item._type = schedule_xml.get('type', None)
+            schedule_item._frequency = schedule_xml.get('frequency', None)
+            schedule_item._next_run_at = schedule_xml.get('nextRunAt', None)
+            schedule_item._end_schedule_at = schedule_xml.get('endScheduleAt', None)
```
LGraber: Is this supposed to be a date?

tableauserverclient/server/endpoint/schedules_endpoint.py
```diff
@@ -0,0 +1,31 @@
+from .endpoint import Endpoint
+from .. import PaginationItem, ScheduleItem
+import logging
+
+logger = logging.getLogger('tableau.endpoint.schedules')
+
+
+class Schedules(Endpoint):
```
LGraber: You can't update a schedule right now?

tableauserverclient/models/schedule_item.py
t8y8: Frequency is a string (or, I guess, an enum of a few potential values)

"Weekly", "Daily", "Hourly" there might be another one

tableauserverclient/models/schedule_item.py
t8y8: If the schedule is 'on' or not.

"Active" or "Suspended"

tableauserverclient/models/schedule_item.py
t8y8: "Extract" or "Subscription" -- maybe more in the future

tableauserverclient/models/schedule_item.py
t8y8: It's a date in the format "YYYY-MM-DDTHH:MM:SSZ"

We could make these native datetime objects, that seems reasonable

tableauserverclient/server/endpoint/schedules_endpoint.py
t8y8: No :'(

t8y8:
Whoa, this got even bigger :)

It'll take me some time to take a look at it, but I've got a long flight tomorrow...

samples/create_schedules.py
```diff
@@ -0,0 +1,57 @@
...
+import getpass
+import logging
+from datetime import time
+
+parser = argparse.ArgumentParser(description='Creates sample schedules for each type of frequency.')
+parser.add_argument('--server', '-s', required=True, help='server address')
+parser.add_argument('--username', '-u', required=True, help='username to sign into server')
+parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
+                    help='desired logging level (set to error by default)')
+args = parser.parse_args()
```
t8y8: The other samples probably need this as well -- but this should be in an `if __name__ == '__main__'` block so it only runs as a script, and so that we can check for no arguments and print the help.

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,99 @@
+import xml.etree.ElementTree as ET
+from datetime import datetime, time
+from .. import NAMESPACE
+
+
+class IntervalItem(object):
```
t8y8: I don't have an alternate suggestion, so feel free to ignore, but is IntervalItem the best name we've got?

The nested classes are fine for the enum-y purpose you're using it for (man, in py3 we could use real enums) -- but we might be hitting a point where we just needs a constants or common module that holds all these so the actual class with logic doesn't get muddled down.

I think this is also the case for RequestOptions?

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,99 @@
...
+        elif interval_occurrence == IntervalItem.Occurrence.Hours and interval_value not in [1, 2, 4, 6, 8, 12]:
+            error = "Invalid hour value defined: {}.".format(interval_value)
+            raise ValueError(error)
+        elif interval_occurrence == IntervalItem.Occurrence.Minutes and interval_value not in [15, 30]:
+            error = "Invalid minute value defined: {}".format(interval_value)
+            raise ValueError(error)
+
+        cls._validate_time(start_time)
+        cls._validate_time(end_time)
+        interval = [(interval_occurrence.lower(), str(interval_value))]
```
t8y8: If this doesn't need to be mutable can just return a tuple here

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,99 @@
...
+        LastDay = "LastDay"
+
+    def __init__(self, frequency, interval, start_time, end_time=None):
+        self.end_time = end_time
+        self.frequency = frequency
+        self.interval = interval
+        self.start_time = start_time
+
+    @staticmethod
+    def _validate_time(t):
```
t8y8: This might be useful just as a function at the module level, and returns true or false, then the classes that use it can decide what exception to throw, or to handle some other way (like using a default time)

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,203 @@
...
+        return self._updated_at
+
+    def _parse_common_tags(self, schedule_xml):
+        if not isinstance(schedule_xml, ET.Element):
+            schedule_xml = ET.fromstring(schedule_xml).find('.//t:schedule', namespaces=NAMESPACE)
+        if schedule_xml is not None:
+            (_, name, _, _, updated_at, _, frequency, next_run_at, end_schedule_at, execution_order,
+             priority, interval_item) = self._parse_element(schedule_xml)
+
+            self._set_values(None, name, None, None, updated_at, None, frequency, next_run_at, end_schedule_at,
```
t8y8: Can you use keyword arguments for all of these -- else we're doomed to be lost at sea :)

(Side note, can any of these be moved out of the function, it seems crazy big)

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,203 @@
...
+        if execution_order:
+            self._execution_order = execution_order
+        if priority:
+            self._priority = priority
+        if interval_item:
+            self._interval_item = interval_item
+
+    @classmethod
+    def from_response(cls, resp):
+        all_schedule_items = list()
```
t8y8: Nit: We seem to use the literal in most places `[]`

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,203 @@
...
+        parsed_response = ET.fromstring(resp)
+        all_schedule_xml = parsed_response.findall('.//t:schedule', namespaces=NAMESPACE)
+        for schedule_xml in all_schedule_xml:
+            (id, name, state, created_at, updated_at, schedule_type, frequency, next_run_at,
+             end_schedule_at, execution_order, priority, interval_item) = cls._parse_element(schedule_xml)
+
+            schedule_item = cls(name, priority, schedule_type, execution_order, interval_item)
+            schedule_item._set_values(id, None, state, created_at, updated_at, None, frequency, next_run_at,
```
t8y8: Same comment about keyword arguments

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,203 @@
...
+        updated_at = schedule_xml.get('updatedAt', None)
+        schedule_type = schedule_xml.get('type', None)
+        frequency = schedule_xml.get('frequency', None)
+        next_run_at = schedule_xml.get('nextRunAt', None)
+        end_schedule_at = schedule_xml.get('endScheduleAt', None)
+        execution_order = schedule_xml.get('executionOrder', None)
+
+        priority = schedule_xml.get('priority', None)
+        if priority:
+            priority = int(priority)
```
t8y8: This will fail if priority is a non empty string that isn't an integer. Do we know that's always safe to do this conversion?

tableauserverclient/server/endpoint/schedules_endpoint.py
```diff
@@ -0,0 +1,60 @@
...
+        url = self._construct_url()
+        server_response = self.get_request(url, req_options)
+        pagination_item = PaginationItem.from_response(server_response.content)
+        all_schedule_items = ScheduleItem.from_response(server_response.content)
+        return all_schedule_items, pagination_item
+
+    def delete(self, schedule_id):
+        if not schedule_id:
+            error = "Schedule ID undefined"
+            raise ValueError(error)
```
t8y8: The exception should get logged as an error -- or do we not do that on other endpoints?

samples/create_schedules.py
t8y8: Addressing the other samples can be a different Issue#

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,203 @@
...
+        if schedule_xml is not None:
+            (_, name, _, _, updated_at, _, frequency, next_run_at, end_schedule_at, execution_order,
+             priority, interval_item) = self._parse_element(schedule_xml)
+
+            self._set_values(None, name, None, None, updated_at, None, frequency, next_run_at, end_schedule_at,
+                             execution_order, priority, interval_item)
+
+        return self
+
+    def _set_values(self, id, name, state, created_at, updated_at, schedule_type, frequency,
```
t8y8: @RussTheAerialist I took a local copy of this PR to try and address my own feedback and put it up for review, and this method made me wonder if there was a better way.

I'm thinking of instead of having a giant setter that does nothing but check that attributes are not false -- what if something like this:

1. The ScheduleItem constructor has everything initialized to None (or has default args)
2. _parse_element returns only non-null/None attributes it gets back as a dict
3. We can loop over the dict to apply the attributes in _set_values (and remove the giant function signature) or I think we can do away with the method entirely, but the idea was on the tip of my tongue and it left me.

I can dig in farther, but for now I've at least done this:
```
            schedule_item._set_values(id=id,
                                      name=None,
                                      state=state,
                                      created_at=created_at,
                                      updated_at=updated_at,
                                      schedule_type=None,
                                      frequency=frequency,
                                      next_run_at=next_run_at,
                                      end_schedule_at=end_schedule_at,
                                      execution_order=None,
                                      priority=None,
                                      interval_item=None)
```

Thoughts? (I can do that in a separate PR)

samples/create_schedules.py
t8y8: Fixed this in all samples

tableauserverclient/models/interval_item.py
t8y8: Ignored for now

tableauserverclient/models/interval_item.py
t8y8: I looked at it for a while and made a small refactor, but pulling it out can happen in another PR and after more conversation.

tableauserverclient/models/interval_item.py
t8y8: Not sure why this is a list, but it's consistent so I left it

tableauserverclient/models/schedule_item.py
t8y8: Fixed

tableauserverclient/models/schedule_item.py
t8y8: fixed

tableauserverclient/models/schedule_item.py
t8y8: fixed

tableauserverclient/server/endpoint/schedules_endpoint.py
t8y8: Confirmed the XSD schema only allows nonNegInts

RussTheAerialist:
#48 supersedes this change

PR #32, opened 2016-09-14
Fix PEP440 warnings in TravisCI 

t8y8:
This was happening due to bad normalization of the package name in setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,6 @@ setup(
         'requests>=2.11,<2.12.0a0'
     ],
     tests_require=[
-        'requests-mock>=1.0,<1.1a0'
+        'requests_mock>=1.0,<1.1a0'
     ]
 )


t8y8:
Nope, that didn't fix it :(

RussTheAerialist:
@t8y8 are you referring to these errors?

```text
/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/pkg_resources/__init__.py:2510:
PEP440Warning: 'pbr (0.5.2.5.g5b3e942)' is being parsed as a legacy, non PEP 440, version. You
may find odd behavior and sort order. In particular it will be sorted as less than 0.0. It is
recommend to migrate to PEP 440 compatible versions.

  PEP440Warning,
```

These are coming from dependencies being installed by requests-mock, not from the way we are referring to it, from what I can tell.  In this example, it's complaining about pbr being versioned as 0.5.2.5.g5b3e942, which, arguably, is a not so nice version number.

t8y8:
Since this cannot be fixed by us, I shall squint when I see those warnings.

Goodbye PR, I hardly knew ye.

Issue #33, opened 2016-09-15
Update publish sample to use publish mode enum


benlower:
Currently we are using server.PublishMode.Overwrite.

We should hang the enum off the Server so it should be TSC.Server.PublishMode.Overwrite

t8y8:
This was fixed in #35 and extra fixed in #47

PR #35, opened 2016-09-16
#33 fixed publish mode enums and typos in samples

shinchris:


samples/explore_datasource.py
```diff
@@ -1,5 +1,5 @@
 ####
-# This script demonstrates how to use the Tableau Server API
+# This script demonstrates how to use the Tableau Server Client
```
RussTheAerialist: Good Catch

PR #46, opened 2016-09-29
Added decorators for checking values of input to property setters (Issue-38)

LGraber:


tableauserverclient/models/datasource_item.py
```diff
@@ -1,5 +1,6 @@
 import xml.etree.ElementTree as ET
 from .exceptions import UnpopulatedPropertyError
+from .property_decorators import property_not_nullable
```
t8y8: At first glance the code feels ok, but I think there might be a better name than "property_decorator" which both have meanings of their own.

I'll do a more detailed review in a bit

t8y8:
I prefer the decorator having a name that indicates it's intent and not the fact that it happens to be testing properties or is a decorator :) I'm not a stickler but it'll make it easier to understand down the road.

@validate_schema or @validate_model or something like that feels much closer to what it does.

The second question is can we do a single decorator that takes in parameters on what it validates (Django Forms, I believe, is similar to this)

@validate_model(enum=Auth, is_null=False)

It makes the decorator explosion a little more manageable.

RussTheAerialist:
I'd much rather the individual decorators than one über one.  If we have a single one, then we have to account for all possible combinations.  That's error prone.  If we have multiple decorators, we can compose them into whatever combination we need.

LGraber:
I also prefer the individual decorators. I tried writing one but as Russell pointed out it ended up being an explosion and not nearly as easy to read.

t8y8:
Fair point on wanting them to be composable -- I think that can be done without having an uber method though.

Django does it by having one decorator that takes a list of callbacks called 'validator' and that does the validation for the model.

So for us it could be:

```
@validate_model(validators=[is_not_nullable, is_right_type(Auth)])
def function():
    pass
``` 

It encapsulates the "Validation" stuff in on method without making all our functions look like this:
```
@a
@b
@c
@d.setter
def happy_days():
    pass
````

There's also an advantage in making the order more easily specified, I forget if decorators are bottom up or top down in their order, a list could make it explicit.

LGraber:
Order is not so important except to make sure the setter is first (still an issue in yours).

PR #47, opened 2016-10-02
Cleaning up samples

t8y8:
Though not required, I moved all of the samples code to be wrapped in `if __name__ == "__main__"` blocks since that's best practice for command line scripts.

I also cleaned up imports to match pep8, and removed one or two that were unused.

I think we can refactor some of the really long lines a bit more to make them more readable -- especially with respect to the Filter/Sort options, but that can come in a later checkin.

### DIFF
--- a/samples/explore_datasource.py
+++ b/samples/explore_datasource.py
@@ -9,12 +9,12 @@
 # on top of the general operations.
 ####
 
-
-import tableauserverclient as TSC
-import os.path
 import argparse
 import getpass
 import logging
+
+import tableauserverclient as TSC
+
 
 parser = argparse.ArgumentParser(description='Explore datasource functions supported by the Server API.')
 parser.add_argument('--server', '-s', required=True, help='server address')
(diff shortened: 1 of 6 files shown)


samples/explore_datasource.py
```diff
@@ -9,56 +9,61 @@
...
-import tableauserverclient as TSC
-import os.path
 import argparse
 import getpass
 import logging
 
+import tableauserverclient as TSC
+
+
 parser = argparse.ArgumentParser(description='Explore datasource functions supported by the Server API.')
```
RussTheAerialist: Should probably move these into the __main__ conditional also

samples/explore_datasource.py
```diff
@@ -9,56 +9,61 @@
...
+    logging.basicConfig(level=logging_level)
 
-# SIGN IN
-tableau_auth = TSC.TableauAuth(args.username, password)
-server = TSC.Server(args.server)
-with server.auth.sign_in(tableau_auth):
-    # Query projects for use when demonstrating publishing and updating
-    all_projects, pagination_item = server.projects.get()
-    default_project = next((project for project in all_projects if project.is_default()), None)
+    # SIGN IN
```
RussTheAerialist: I'm not sure if this is just a product of github, but the spaces here are showing different than then ones below.  Make sure they are all spaces, not a mix of tabs and spaces.

samples/explore_workbook.py
```diff
@@ -25,72 +26,78 @@
...
+            print([view.name for view in sample_workbook.views])
+
+            # Populate connections
+            server.workbooks.populate_connections(sample_workbook)
+            print("\nConnections for {}: ".format(sample_workbook.name))
+            print(["{0}({1})".format(connection.id, connection.datasource_name)
+                   for connection in sample_workbook.connections])
+
+            # Update tags and show_tabs flag
+            original_tag_set = set(sample_workbook.tags)
```
t8y8: This is a simpler way to copy a set and doesn't require copy.copy

samples/explore_datasource.py
t8y8: Yeah I noticed that when it posted, weird, I'll make sure they're all spaces.

samples/explore_datasource.py
t8y8: I debated that too -- I wonder if I should wrap everything in a main(), and then just have:

```
if __name__ == "__main__":
    main()
```

So it's easier to tell runtime from any setup?

PR #48, opened 2016-10-03
Adding Schedules Support Round 2

t8y8:
Resubmitted @shinchris 's PR #30 with a few small changes -- I updated the old PR with comments on my own review.

### DIFF
--- a/samples/create_schedules.py
+++ b/samples/create_schedules.py
@@ -34,29 +34,29 @@ def main():
     server = TSC.Server(args.server)
     with server.auth.sign_in(tableau_auth):
         # Hourly Schedule
-        hourly_interval = TSC.IntervalItem.create_hourly(time(2, 30), time(23, 0), TSC.IntervalItem.Occurrence.Hours, 2)
+        hourly_interval = TSC.HourlyInterval(time(2, 30), time(23, 0), TSC.IntervalItem.Occurrence.Hours, 2)
         hourly_schedule = TSC.ScheduleItem("Hourly-Schedule", 50, TSC.ScheduleItem.Type.Extract,
                                            TSC.ScheduleItem.ExecutionOrder.Parallel, hourly_interval)
         hourly_schedule = server.schedules.create(hourly_schedule)
         print("Hourly schedule created (ID: {}).".format(hourly_schedule.id))
 
         # Daily Schedule
-        daily_interval = TSC.IntervalItem.create_daily(time(5))
+        daily_interval = TSC.DailyInterval(time(5))
         daily_schedule = TSC.ScheduleItem("Daily-Schedule", 60, TSC.ScheduleItem.Type.Subscription,
                                           TSC.ScheduleItem.ExecutionOrder.Serial, daily_interval)
         daily_schedule = server.schedules.create(daily_schedule)
         print("Daily schedule created (ID: {}).".format(daily_schedule.id))
 
         # Weekly Schedule
-        weekly_interval = TSC.IntervalItem.create_weekly(time(19, 15), TSC.IntervalItem.Day.Monday,
-                                                         TSC.IntervalItem.Day.Wednesday, TSC.IntervalItem.Day.Friday)
+        weekly_interval = TSC.WeeklyInterval(time(19, 15), TSC.IntervalItem.Day.Monday,
+                                             TSC.IntervalItem.Day.Wednesday, TSC.IntervalItem.Day.Friday)
         weekly_schedule = TSC.ScheduleItem("Weekly-Schedule", 70, TSC.ScheduleItem.Type.Extract,
                                            TSC.ScheduleItem.ExecutionOrder.Serial, weekly_interval)
         weekly_schedule = server.schedules.create(weekly_schedule)
         print("Weekly schedule created (ID: {}).".format(weekly_schedule.id))
 
         # Monthly Schedule
-        monthly_interval = TSC.IntervalItem.create_monthly(time(23, 30), 15)
+        monthly_interval = TSC.MonthlyInterval(time(23, 30), 15)
         monthly_schedule = TSC.ScheduleItem("Monthly-Schedule", 80, TSC.ScheduleItem.Type.Subscription,
                                             TSC.ScheduleItem.ExecutionOrder.Parallel, monthly_interval)
         monthly_schedule = server.schedules.create(monthly_schedule)
--- a/samples/create_schedules.py
+++ b/samples/create_schedules.py
@@ -34,29 +34,40 @@ def main():
     server = TSC.Server(args.server)
     with server.auth.sign_in(tableau_auth):
         # Hourly Schedule
-        hourly_interval = TSC.HourlyInterval(time(2, 30), time(23, 0), TSC.IntervalItem.Occurrence.Hours, 2)
+        # This schedule will run every 2 hours between 2:30AM and 11:00PM
+        hourly_interval = TSC.HourlyInterval(start_time=time(2, 30),
+                                             end_time=time(23, 0),
+                                             interval_occurrence=TSC.IntervalItem.Occurrence.Hours,
+                                             interval_value=2)
+
         hourly_schedule = TSC.ScheduleItem("Hourly-Schedule", 50, TSC.ScheduleItem.Type.Extract,
                                            TSC.ScheduleItem.ExecutionOrder.Parallel, hourly_interval)
         hourly_schedule = server.schedules.create(hourly_schedule)
         print("Hourly schedule created (ID: {}).".format(hourly_schedule.id))
 
         # Daily Schedule
-        daily_interval = TSC.DailyInterval(time(5))
+        # This schedule will run every day at 5AM
+        daily_interval = TSC.DailyInterval(start_time=time(5))
         daily_schedule = TSC.ScheduleItem("Daily-Schedule", 60, TSC.ScheduleItem.Type.Subscription,
                                           TSC.ScheduleItem.ExecutionOrder.Serial, daily_interval)
         daily_schedule = server.schedules.create(daily_schedule)
         print("Daily schedule created (ID: {}).".format(daily_schedule.id))
 
         # Weekly Schedule
-        weekly_interval = TSC.WeeklyInterval(time(19, 15), TSC.IntervalItem.Day.Monday,
-                                             TSC.IntervalItem.Day.Wednesday, TSC.IntervalItem.Day.Friday)
+        # This schedule will wun every Monday, Wednesday, and Friday at 7:15PM
+        weekly_interval = TSC.WeeklyInterval(time(19, 15),
+                                             TSC.IntervalItem.Day.Monday,
+                                             TSC.IntervalItem.Day.Wednesday,
+                                             TSC.IntervalItem.Day.Friday)
         weekly_schedule = TSC.ScheduleItem("Weekly-Schedule", 70, TSC.ScheduleItem.Type.Extract,
                                            TSC.ScheduleItem.ExecutionOrder.Serial, weekly_interval)
         weekly_schedule = server.schedules.create(weekly_schedule)
         print("Weekly schedule created (ID: {}).".format(weekly_schedule.id))
 
         # Monthly Schedule
-        monthly_interval = TSC.MonthlyInterval(time(23, 30), 15)
+        # This schedule will run on the 15th of every month at 11:30PM
+        monthly_interval = TSC.MonthlyInterval(start_time=time(23, 30),
+                                               interval_value=15)
         monthly_schedule = TSC.ScheduleItem("Monthly-Schedule", 80, TSC.ScheduleItem.Type.Subscription,
                                             TSC.ScheduleItem.ExecutionOrder.Parallel, monthly_interval)
         monthly_schedule = server.schedules.create(monthly_schedule)
--- a/samples/create_schedules.py
+++ b/samples/create_schedules.py
@@ -5,53 +5,63 @@
 # To run the script, you must have installed Python 2.7.9 or later.
 ####
 
-import tableauserverclient as TSC
+
 import argparse
 import getpass
 import logging
+
 from datetime import time
 
-parser = argparse.ArgumentParser(description='Creates sample schedules for each type of frequency.')
-parser.add_argument('--server', '-s', required=True, help='server address')
-parser.add_argument('--username', '-u', required=True, help='username to sign into server')
-parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
-                    help='desired logging level (set to error by default)')
-args = parser.parse_args()
+import tableauserverclient as TSC
 
-password = getpass.getpass("Password: ")
 
-# Set logging level based on user input, or error by default
-logging_level = getattr(logging, args.logging_level.upper())
-logging.basicConfig(level=logging_level)
+def main():
 
-tableau_auth = TSC.TableauAuth(args.username, password)
-server = TSC.Server(args.server)
-with server.auth.sign_in(tableau_auth):
-    # Hourly Schedule
-    hourly_interval = TSC.IntervalItem.create_hourly(time(2, 30), time(23, 0), TSC.IntervalItem.Occurrence.Hours, 2)
-    hourly_schedule = TSC.ScheduleItem("Hourly-Schedule", 50, TSC.ScheduleItem.Type.Extract,
-                                       TSC.ScheduleItem.ExecutionOrder.Parallel, hourly_interval)
-    hourly_schedule = server.schedules.create(hourly_schedule)
-    print("Hourly schedule created (ID: {}).".format(hourly_schedule.id))
+    parser = argparse.ArgumentParser(description='Creates sample schedules for each type of frequency.')
+    parser.add_argument('--server', '-s', required=True, help='server address')
+    parser.add_argument('--username', '-u', required=True, help='username to sign into server')
+    parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
+                        help='desired logging level (set to error by default)')
+    args = parser.parse_args()
 
-    # Daily Schedule
-    daily_interval = TSC.IntervalItem.create_daily(time(5))
-    daily_schedule = TSC.ScheduleItem("Daily-Schedule", 60, TSC.ScheduleItem.Type.Subscription,
-                                      TSC.ScheduleItem.ExecutionOrder.Serial, daily_interval)
-    daily_schedule = server.schedules.create(daily_schedule)
-    print("Daily schedule created (ID: {}).".format(daily_schedule.id))
+    password = getpass.getpass("Password: ")
 
-    # Weekly Schedule
-    weekly_interval = TSC.IntervalItem.create_weekly(time(19, 15), TSC.IntervalItem.Day.Monday,
-                                                     TSC.IntervalItem.Day.Wednesday, TSC.IntervalItem.Day.Friday)
-    weekly_schedule = TSC.ScheduleItem("Weekly-Schedule", 70, TSC.ScheduleItem.Type.Extract,
-                                       TSC.ScheduleItem.ExecutionOrder.Serial, weekly_interval)
-    weekly_schedule = server.schedules.create(weekly_schedule)
-    print("Weekly schedule created (ID: {}).".format(weekly_schedule.id))
+    # Set logging level based on user input, or error by default
+    logging_level = getattr(logging, args.logging_level.upper())
+    logging.basicConfig(level=logging_level)
 
-    # Monthly Schedule
-    monthly_interval = TSC.IntervalItem.create_monthly(time(23, 30), 15)
-    monthly_schedule = TSC.ScheduleItem("Monthly-Schedule", 80, TSC.ScheduleItem.Type.Subscription,
-                                        TSC.ScheduleItem.ExecutionOrder.Parallel, monthly_interval)
-    monthly_schedule = server.schedules.create(monthly_schedule)
-    print("Monthly schedule created (ID: {}).".format(monthly_schedule.id))
+    tableau_auth = TSC.TableauAuth(args.username, password)
+    server = TSC.Server(args.server)
+    with server.auth.sign_in(tableau_auth):
+        # Hourly Schedule
+        hourly_interval = TSC.IntervalItem.create_hourly(time(2, 30), time(23, 0), TSC.IntervalItem.Occurrence.Hours, 2)
+        hourly_schedule = TSC.ScheduleItem("Hourly-Schedule", 50, TSC.ScheduleItem.Type.Extract,
+                                           TSC.ScheduleItem.ExecutionOrder.Parallel, hourly_interval)
+        hourly_schedule = server.schedules.create(hourly_schedule)
+        print("Hourly schedule created (ID: {}).".format(hourly_schedule.id))
+
+        # Daily Schedule
+        daily_interval = TSC.IntervalItem.create_daily(time(5))
+        daily_schedule = TSC.ScheduleItem("Daily-Schedule", 60, TSC.ScheduleItem.Type.Subscription,
+                                          TSC.ScheduleItem.ExecutionOrder.Serial, daily_interval)
+        daily_schedule = server.schedules.create(daily_schedule)
+        print("Daily schedule created (ID: {}).".format(daily_schedule.id))
+
+        # Weekly Schedule
+        weekly_interval = TSC.IntervalItem.create_weekly(time(19, 15), TSC.IntervalItem.Day.Monday,
+                                                         TSC.IntervalItem.Day.Wednesday, TSC.IntervalItem.Day.Friday)
+        weekly_schedule = TSC.ScheduleItem("Weekly-Schedule", 70, TSC.ScheduleItem.Type.Extract,
+                                           TSC.ScheduleItem.ExecutionOrder.Serial, weekly_interval)
+        weekly_schedule = server.schedules.create(weekly_schedule)
+        print("Weekly schedule created (ID: {}).".format(weekly_schedule.id))
+
+        # Monthly Schedule
+        monthly_interval = TSC.IntervalItem.create_monthly(time(23, 30), 15)
+        monthly_schedule = TSC.ScheduleItem("Monthly-Schedule", 80, TSC.ScheduleItem.Type.Subscription,
+                                            TSC.ScheduleItem.ExecutionOrder.Parallel, monthly_interval)
+        monthly_schedule = server.schedules.create(monthly_schedule)
+        print("Monthly schedule created (ID: {}).".format(monthly_schedule.id))
+
+
+if __name__ == '__main__':
+    main()
--- a/samples/create_schedules.py
+++ b/samples/create_schedules.py
@@ -1,0 +1,57 @@
+####
+# This script demonstrates how to create schedules using the Tableau
+# Server Client.
+#
+# To run the script, you must have installed Python 2.7.9 or later.
+####
+
+import tableauserverclient as TSC
+import argparse
+import getpass
+import logging
+from datetime import time
+
+parser = argparse.ArgumentParser(description='Creates sample schedules for each type of frequency.')
+parser.add_argument('--server', '-s', required=True, help='server address')
+parser.add_argument('--username', '-u', required=True, help='username to sign into server')
+parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
+                    help='desired logging level (set to error by default)')
+args = parser.parse_args()
+
+password = getpass.getpass("Password: ")
+
+# Set logging level based on user input, or error by default
+logging_level = getattr(logging, args.logging_level.upper())
+logging.basicConfig(level=logging_level)
+
+tableau_auth = TSC.TableauAuth(args.username, password)
+server = TSC.Server(args.server)
+with server.auth.sign_in(tableau_auth):
+    # Hourly Schedule
+    hourly_interval = TSC.IntervalItem.create_hourly(time(2, 30), time(23, 0), TSC.IntervalItem.Occurrence.Hours, 2)
+    hourly_schedule = TSC.ScheduleItem("Hourly-Schedule", 50, TSC.ScheduleItem.Type.Extract,
+                                       TSC.ScheduleItem.ExecutionOrder.Parallel, hourly_interval)
+    hourly_schedule = server.schedules.create(hourly_schedule)
+    print("Hourly schedule created (ID: {}).".format(hourly_schedule.id))
+
+    # Daily Schedule
+    daily_interval = TSC.IntervalItem.create_daily(time(5))
+    daily_schedule = TSC.ScheduleItem("Daily-Schedule", 60, TSC.ScheduleItem.Type.Subscription,
+                                      TSC.ScheduleItem.ExecutionOrder.Serial, daily_interval)
+    daily_schedule = server.schedules.create(daily_schedule)
+    print("Daily schedule created (ID: {}).".format(daily_schedule.id))
+
+    # Weekly Schedule
+    weekly_interval = TSC.IntervalItem.create_weekly(time(19, 15), TSC.IntervalItem.Day.Monday,
+                                                     TSC.IntervalItem.Day.Wednesday, TSC.IntervalItem.Day.Friday)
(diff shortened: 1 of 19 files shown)


tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,225 @@
...
+    @property
+    def end_schedule_at(self):
+        return self._end_schedule_at
+
+    @property
+    def execution_order(self):
+        return self._execution_order
+
+    @execution_order.setter
+    def execution_order(self, value):
```
t8y8: Need to update this with @LGraber 's decorators still

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,225 @@
...
+                             end_schedule_at=end_schedule_at,
+                             execution_order=execution_order,
+                             priority=priority,
+                             interval_item=interval_item)
+
+        return self
+
+    def _set_values(self, id, name, state, created_at, updated_at, schedule_type, frequency,
+                    next_run_at, end_schedule_at, execution_order, priority, interval_item):
+        if id is not None:
```
RussTheAerialist: I feel like accepting **kwargs and then having a list of accepted attributes and iterating over them might be cleaner than all of these.

```python
# Assume some static value called "ALLOWED_ATTRIBUTES"
def _set_values(self, **kwargs):
    for attr_name in self.ALLLOWED_ATTRIBUTES:
        if attr_name in kwargs and kwargs[attr_name]:
            setattr(self, '_{}'.format(attr_name), kwargs[attr_name]
```

Since this is mostly a helper method, I think it's okay to use the kwargs

tableauserverclient/server/request_factory.py
```diff
@@ -129,6 +129,54 @@ def create_req(self, project_item):
...
+            for interval in interval_item.interval:
+                (expression, value) = interval
+                single_interval_element = ET.SubElement(intervals_element, 'interval')
+                single_interval_element.attrib[expression] = value
+        return ET.tostring(xml_request)
+
+    def update_req(self, schedule_item):
+        xml_request = ET.Element('tsRequest')
+        schedule_element = ET.SubElement(xml_request, 'schedule')
+        if schedule_item.name:
```
RussTheAerialist: This could also be turned in a loop across all the different attributes.  since it's pretty prescriptive, the only tricky-ish part is the conversion of things like executionOrder to execution_order, but I think that's well understood.

tableauserverclient/models/schedule_item.py
t8y8: I'm on board -- there are several items where we should do that (site_item is also hilariously large). I'll do that in another PR -- I'll pick one and try it, then apply it to all of them

tableauserverclient/models/schedule_item.py
```diff
@@ -1,5 +1,6 @@
 import xml.etree.ElementTree as ET
 from .interval_item import IntervalItem
+from .property_decorators import property_is_enum, property_is_boolean, property_not_empty, property_not_nullable
```
t8y8: I will remove the imports I'm not using, copy/paste error

LGraber:
I would like to talk about this. I have a lot of trouble following the "IntervalItem" logic. It seems like we are trying to force 4 classes into one and the meaning of various fields changes based on one field. Given the complexity that ends up being introduced, I guess I don't see why we didn't just create appropriate classes for the different types of intervals. The result of the above code is that after I create an item, I can't change it. I have to recreate one. If we think of how a user flow might work here, that does not feel like a good experience. Lets talk about this.

t8y8:
I don't think having to recreate an IntervalItem is a big deal -- these are basically thin wrappers over combining some enums together into a usable blob to pass to a schedule. I don't think it's going to be common to reuse an interval instance.

Actually, this follows the stdlib pattern for datetime pretty closely.

```
important_schedule = IntervalItem.create_hourly(start, end, occurrence, value)
less_important_schedule = IntervalItem.create_daily(start_time)

for extract in extracts:
    if extract_is_important(extract):
        extract.update_schedule(important_schedule)
    else:
        extract.update_schedule(less_important_schedule)
```

I do agree that the *class itself* is pretty large and hairy and could be broken down into smaller classes for readability. 

The above code could then be something like:
```
important_schedule = HourlyInterval(start, end, occurrence, value)
less_important_schedule = DailyInterval(start_time)

for extract in extracts:
    if extract_is_important(extract):
        extract.update_schedule(important_schedule)
    else:
        extract.update_schedule(less_important_schedule)
```

A little nicer, but feels about the same -- I think the bigger wins are for code maintainability.

I'm guessing Chris was just following the patterns of similar our APIs (with the nested enum classes) and staying a fairly thin wrapper over our XML schema. Admittedly this is also the most complicated type in our schema :/

RussTheAerialist:
From my viewpoint, the second code block that you provided is better.  4 separate classes is more clear to the user, and more clear on being able to make modifications to the schedule/interval.

RE: easier to parse the xml vs. easier for the user to grok.  I'm going to go with easier for the user to grok 100% of the time.  The little extra work we need to do in parsing out to different classes is minimal and worth the effort for clarity

t8y8:
@LGraber @RussTheAerialist here's a first stab at it -- there's a lot of validation logic I haven't ported to the individual Interval classes yet

tableauserverclient/models/interval_item.py
```diff
@@ -75,6 +75,20 @@ def from_xml_element(cls, parsed_response, frequency):
...
 class HourlyInterval(IntervalItem):
     def __init__(self, start_time, end_time, interval_occurrence, interval_value):
+        self._validate_time(start_time)
+        self._validate_time(end_time)
+
+        if interval_occurrence != IntervalItem.Occurrence.Hours and \
+                interval_occurrence != IntervalItem.Occurrence.Minutes:
+            error = "Invalid interval type defined: {}.".format(interval_occurrence)
+            raise ValueError(error)
+        elif interval_occurrence == IntervalItem.Occurrence.Hours and int(interval_value) not in [1, 2, 4, 6, 8, 12]:
```
t8y8: Similarly I'm not sure how this worked before, parsed responses will be strings, found via test failures.

tableauserverclient/models/interval_item.py
```diff
@@ -102,44 +117,10 @@ class MonthlyInterval(IntervalItem):
     def __init__(self, start_time, interval_value):
         self._validate_time(start_time)
 
+        if (int(interval_value) < 1 or int(interval_value) > 31) and interval_value != IntervalItem.Day.LastDay:
```
t8y8: Again, not sure how this worked without converting to an int before

LGraber:
Are you reviewing 'your' changes now?

t8y8:
Just leaving comments for you guys :) -- it's not 100% ready yet but wanted it up for eyes on it

LGraber:
It looks like you are mid change now. I am expecting their will be 4 "Interval" classes which can be modified after creation with validation. You cannot change hourly to monthly by simply changing a field but that is fine. They are different classes with very different settings. The IntervalItem class can go away. Let me know when I should take a look again or if you want me to take a stab.

t8y8:
@LGraber I still don't think that use case will be common. People aren't going to be likely to create `morhping_interval` and constantly update it.

```
every_tuesday = WeeklyInterval(start_time, Interval.Occurance.Day.Tuesday)
awesome_schedule = Schedule(name='The Best', interval=every_tuesday, *args)
for wb in wbs:
    add_to_schedule(wb, awesome_schedule)
```
It's easy enough to attach setters that trigger validation logic so you can do that if we really need to -- but I'm not sure it's worth it at all. (@benlower, any use cases come to mind?)

For the breakdown of the classes: I've got a question

As currently implemented I have the Daily/Weekly/Hourly/Monthly Intervals inheriting from IntervalItem. IntervalItem owns the parsing logic and building the more specific IntervalItem types, as well as holding their shared enums and default instance variables (start_time, end_time, frequency). 

It also lets us keep the parsing logic centralized in IntervalItem which acts as a proxy and generates the more specific IntervalType. I liked not having to change all the schedule_endpoints so I kept it, but if there's a suggestion for an alternative I'm all ears (/cc @RussTheAerialist )

LGraber:
The scenario is that I query the rest endpoint to get the schedule and then I want to change it.

LGraber:
I think you are going to like what the code looks like.

t8y8:
@LGraber Ok fine, maybe it looks a little cleaner.

Maybe 💛

samples/create_schedules.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Set logging level based on user input, or error by default
+    logging_level = getattr(logging, args.logging_level.upper())
+    logging.basicConfig(level=logging_level)
+
+    tableau_auth = TSC.TableauAuth(args.username, password)
+    server = TSC.Server(args.server)
+    with server.auth.sign_in(tableau_auth):
+        # Hourly Schedule
+        hourly_interval = TSC.IntervalItem.create_hourly(time(2, 30), time(23, 0), TSC.IntervalItem.Occurrence.Hours, 2)
```
t8y8: Yikes, I need to update these now :)

(I wonder if we can find a way to run these, either in a local task, or in travis somehow)

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,175 @@
...
+    def interval(self, intervals):
+        interval_occurrence, interval_value = intervals
+
+        Hours = IntervalItem.Occurrence.Hours
+        Minutes = IntervalItem.Occurrence.Minutes
+
+        VALID_HOUR_INTERVALS = {1, 2, 4, 6, 8, 12}
+        VALID_MIN_INTERVALS = {15, 30}
+
+        if interval_occurrence not in {Hours, Minutes}:
```
LGraber: Ideally our error messages could list the possible valid values.

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,175 @@
...
+            error = "Invalid interval type defined: {}.".format(interval_occurrence)
+            raise ValueError(error)
+        elif interval_occurrence == Hours and int(interval_value) not in VALID_HOUR_INTERVALS:
+            error = "Invalid hour value defined: {}.".format(interval_value)
+            raise ValueError(error)
+        elif interval_occurrence == Minutes and int(interval_value) not in VALID_MIN_INTERVALS:
+            error = "Invalid minute value defined: {}".format(interval_value)
+            raise ValueError(error)
+
+        self._interval = [(interval_occurrence.lower(), str(interval_value))]
```
LGraber: why do you do "lower"? Shouldn't the enum's value be correct already. This seems silly. Also, why are we storing it as a string. This means that when the user asks for it back, they won't get the same type they passed in?

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,175 @@
...
+        Hours = IntervalItem.Occurrence.Hours
+        Minutes = IntervalItem.Occurrence.Minutes
+
+        VALID_HOUR_INTERVALS = {1, 2, 4, 6, 8, 12}
+        VALID_MIN_INTERVALS = {15, 30}
+
+        if interval_occurrence not in {Hours, Minutes}:
+            error = "Invalid interval type defined: {}.".format(interval_occurrence)
+            raise ValueError(error)
+        elif interval_occurrence == Hours and int(interval_value) not in VALID_HOUR_INTERVALS:
```
LGraber: why are we casting to int? The user should pass in an int. I would rather have clear types instead of let's see what happens. If they didn't give us an int ... error

RussTheAerialist:
Iirc, There are some properties that aren't returned on the update that are
returned on the query, so we create a copy and update the new copy we
return. This also keeps in line with the "we don't modify existing objects
on a result, but return a new copy"

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,175 @@
...
+        _validate_time(value)
+        self._start_time = value
+
+    @property
+    def interval(self):
+        return self._interval
+
+    @interval.setter
+    def interval(self, interval_values):
+        if not all(hasattr(IntervalItem.Day, day) for day in interval_values):
```
LGraber: I am not totally gungho on decorators but if we see this pattern again of a list of enums instead of just an enum, I would move this to a decorator to be consistent

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,175 @@
...
+        Thursday = "Thursday"
+        Friday = "Friday"
+        Saturday = "Saturday"
+        LastDay = "LastDay"
+
+
+class HourlyInterval(object):
+    def __init__(self, start_time, end_time, interval_occurrence, interval_value):
+
+        self.start_time = start_time
```
LGraber: For other code, it seems we start by setting the internal variable to None before calling the setter. Any reason for using that pattern versus not?

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,239 @@
...
+        self.interval_item = interval_item
+
+        # Invoke setter
+        self.execution_order = execution_order
+        self.name = name
+        self.priority = priority
+        self.schedule_type = schedule_type
+
+    @property
+    def created_at(self):
```
LGraber: This has to be a datetime, right?

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,239 @@
...
+
+    class State:
+        Active = "Active"
+        Suspended = "Suspended"
+
+    def __init__(self, name, priority, schedule_type, execution_order, interval_item):
+        self._created_at = None
+        self._end_schedule_at = None
+        self._execution_order = None
+        self._frequency = None
```
LGraber: what is frequency if I have an interval associated with this? That object contains a frequency. What is this? Does this just reference the "frequency" in the interval? If so, why do we have an actual property here instead of just accessing the frequency member of the Interval?

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,239 @@
...
+    def execution_order(self):
+        return self._execution_order
+
+    @execution_order.setter
+    @property_is_enum(ExecutionOrder)
+    def execution_order(self, value):
+        self._execution_order = value
+
+    @property
+    def frequency(self):
```
LGraber: I think this is supposed to access the interval's frequency

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,239 @@
...
+    def next_run_at(self):
+        return self._next_run_at
+
+    @property
+    def priority(self):
+        return self._priority
+
+    @priority.setter
+    def priority(self, value):
+        if value < 1 or value > 100:
```
LGraber: feels like "@property_is_int(min,max)" to be consistent with our other handling. min = None and/or max = None meaning don't check that boundary

tableauserverclient/server/endpoint/schedules_endpoint.py
```diff
@@ -0,0 +1,60 @@
...
+logger = logging.getLogger('tableau.endpoint.schedules')
+
+
+class Schedules(Endpoint):
+    def __init__(self, parent_srv):
+        super(Endpoint, self).__init__()
+        self.baseurl = "{0}/schedules"
+        self.parent_srv = parent_srv
+
+    def _construct_url(self):
```
LGraber: We changed the pattern here so that it was in sync with the root server class. We don't use _construct_url. We just made base_url a property and build the string when you call it. Let's be consistent. It was done after Chris wrote the original pass on this but we should update.

tableauserverclient/server/request_factory.py
```diff
@@ -129,6 +129,54 @@ def create_req(self, project_item):
         return ET.tostring(xml_request)
 
 
+class ScheduleRequest(object):
+    def create_req(self, schedule_item):
```
LGraber: These are getting so hard to read. I want to either put a comment above them which shows what the output of what we are building looks like or something.

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,101 @@
...
+        Monday = "Monday"
+        Tuesday = "Tuesday"
+        Wednesday = "Wednesday"
+        Thursday = "Thursday"
+        Friday = "Friday"
+        Saturday = "Saturday"
+        LastDay = "LastDay"
+
+    def __init__(self, frequency, interval, start_time, end_time=None):
+        self.end_time = end_time
```
LGraber: This is old. I know it is a property now. :(

tableauserverclient/server/endpoint/schedules_endpoint.py
t8y8: Yup, noticed this too, I've got this queued up for the next rev

samples/create_schedules.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Set logging level based on user input, or error by default
+    logging_level = getattr(logging, args.logging_level.upper())
+    logging.basicConfig(level=logging_level)
+
+    tableau_auth = TSC.TableauAuth(args.username, password)
+    server = TSC.Server(args.server)
+    with server.auth.sign_in(tableau_auth):
+        # Hourly Schedule
+        hourly_interval = TSC.HourlyInterval(time(2, 30), time(23, 0), TSC.IntervalItem.Occurrence.Hours, 2)
```
t8y8: Done. Added keyword arguments to several of them as well so the values are clearer

samples/create_schedules.py
```diff
@@ -0,0 +1,67 @@
...
+        hourly_schedule = TSC.ScheduleItem("Hourly-Schedule", 50, TSC.ScheduleItem.Type.Extract,
+                                           TSC.ScheduleItem.ExecutionOrder.Parallel, hourly_interval)
+        hourly_schedule = server.schedules.create(hourly_schedule)
+        print("Hourly schedule created (ID: {}).".format(hourly_schedule.id))
+
+        # Daily Schedule
+        daily_interval = TSC.DailyInterval(time(5))
```
t8y8: Done

samples/create_schedules.py
```diff
@@ -0,0 +1,67 @@
...
+        # Weekly Schedule
+        weekly_interval = TSC.WeeklyInterval(time(19, 15), TSC.IntervalItem.Day.Monday,
+                                             TSC.IntervalItem.Day.Wednesday, TSC.IntervalItem.Day.Friday)
+        weekly_schedule = TSC.ScheduleItem("Weekly-Schedule", 70, TSC.ScheduleItem.Type.Extract,
+                                           TSC.ScheduleItem.ExecutionOrder.Serial, weekly_interval)
+        weekly_schedule = server.schedules.create(weekly_schedule)
+        print("Weekly schedule created (ID: {}).".format(weekly_schedule.id))
+
+        # Monthly Schedule
+        monthly_interval = TSC.MonthlyInterval(time(23, 30), 15)
```
t8y8: Done

tableauserverclient/__init__.py
```diff
@@ -1,9 +1,10 @@
 from .namespace import NAMESPACE
 from .models import ConnectionItem, DatasourceItem,\
-    GroupItem, PaginationItem, ProjectItem, \
+    GroupItem, PaginationItem, ProjectItem, ScheduleItem, \
     SiteItem, TableauAuth, UserItem, ViewItem, WorkbookItem, UnpopulatedPropertyError
 from .server import RequestOptions, Filter, Sort, Server, ServerResponseError,\
     MissingRequiredFieldError, NotSignedInError
+from .models.interval_item import *
```
t8y8: aww, but I'm lazy. ok done.

tableauserverclient/models/interval_item.py
t8y8: This feels cleaner to me because it hide the _blah versions as the implementation detail that they are and keeps the `__init__` function clean.

I then don't need to have a bunch of these in `__init__`:
```
        # Invoke setter
        self.name = name
```

tableauserverclient/models/interval_item.py
t8y8: done

tableauserverclient/models/interval_item.py
t8y8: I'd like to clean this up in a separate PR, it needs deeper digging

The problem is that the schedule_item parser has to deal with lots of values that could be ints or strings (eg [1,2,3] or "LastDay"), so it returns them all as strings from the XML. The user can still pass in an int, it's harmless to cast an int to an int -- this deals with input from the parser

tableauserverclient/models/interval_item.py
t8y8: I don't know... Chris did and I didn't question it, let me take a look

tableauserverclient/models/schedule_item.py
t8y8: You'll need to take a look at the XSD (and grab a glass of wine). On the schedule level there's a frequency attribute (Daily, Hourly, Monthly) and then the interval is a frequenyDetail

tableauserverclient/models/schedule_item.py
t8y8: We currently parse it as a string from the XML -- it's read only so I think that's fine

tableauserverclient/models/schedule_item.py
t8y8: Working on this now, the min or max=None is a little tricky.

tableauserverclient/models/interval_item.py
LGraber: Russell?

tableauserverclient/models/schedule_item.py
t8y8: Ok, done.

```
def property_is_int(min, max=2**31):
    def property_type_decorator(func):
        @wraps(func)
        def wrapper(self, value):
            if min is None:
                return func(self, value)

            if value < min or value > max:
                error = "Invalid priority defined: {}.".format(value)
                raise ValueError(error)

            return func(self, value)

        return wrapper

    return property_type_decorator
```

This means that passing just None will skip validation, otherwise it's min/max.

Technically to make this signature work, Max has a default value of 2**31

tableauserverclient/models/schedule_item.py
t8y8: There isn't a setter for frequency on scheduleitem

t8y8: I just confirmed I can remove that attribute entirely, tests pass (other than the ones that check for it explicitly), and the sample still runs... should I nix it entirely?

tableauserverclient/models/schedule_item.py
```diff
@@ -0,0 +1,236 @@
+import xml.etree.ElementTree as ET
+from .interval_item import *
```
t8y8: I'll fix this to not be a *

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,168 @@
...
+class IntervalItem(object):
+    class Frequency:
+        Hourly = "Hourly"
+        Daily = "Daily"
+        Weekly = "Weekly"
+        Monthly = "Monthly"
+
+    class Occurrence:
+        Hours = "hours"
+        Minutes = "minutes"
```
LGraber: Minutes should come before hours.

tableauserverclient/models/interval_item.py
```diff
@@ -58,26 +58,22 @@ def interval(self):
...
+        # We use fractional hours for the two minute-based intervals.
+        # Need to convert to minutes from hours here
+        if interval in {.25, .5}:
+            interval_value = int(interval * 60)
+            interval_occurrence = IntervalItem.Occurrence.Minutes
+        else:
+            interval_value = interval
+            interval_occurrence = IntervalItem.Occurrence.Hours
+
         self._interval = [(interval_occurrence, str(interval_value))]
```
t8y8: Intervals are tuples of (Occurence, Value), so in this case it would return
```
('hours', 5) or ('minutes', 15) or ('weekDay', 'Monday')
```

It's consistent across the different interval values -- which is how the serializer wants it:
```
        if interval_item.interval:
            intervals_element = ET.SubElement(frequency_element, 'intervals')
            for interval in interval_item.interval:
                expression, value = interval
                single_interval_element = ET.SubElement(intervals_element, 'interval')
                single_interval_element.attrib[expression] = value
```

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,166 @@
...
+            interval_occurrence = IntervalItem.Occurrence.Hours
+
+        self._interval = [(interval_occurrence, str(interval_value))]
+
+
+class DailyInterval(object):
+    def __init__(self, start_time):
+        self.start_time = start_time
+        self._frequency = IntervalItem.Frequency.Daily
+        self.end_time = None
```
LGraber: why do you have this (besides the issue with serialization)? I would rather have the serializer clean it than have properties which are meaningless but visible.

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,166 @@
...
+        self.interval = None
```
LGraber: Same for this

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,166 @@
...
+    def start_time(self, value):
+        self._start_time = value
+
+
+class WeeklyInterval(object):
+    def __init__(self, start_time, *interval_values):
+        self.start_time = start_time
+        self._frequency = IntervalItem.Frequency.Weekly
+        self.interval = interval_values
+        self.end_time = None
```
LGraber: Same here

tableauserverclient/models/interval_item.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self._interval = [(IntervalItem.Occurrence.WeekDay, day) for day in interval_values]
+
+
+class MonthlyInterval(object):
+    def __init__(self, start_time, interval_value):
+        self.start_time = start_time
+        self._frequency = IntervalItem.Frequency.Monthly
+        self.interval = str(interval_value)
+        self.end_time = None
```
LGraber: And here. No?

tableauserverclient/models/interval_item.py
t8y8: Are you talking about the `end_time` arritbutes?

Yeah, that was so I didn't have to re-write the serializer, which I consider to be the riskiest part.

We'd have to add a bunch of `hasattr` or similar checks into the serializer, I think it's the lesser of two evils to have the empty attributes here

tableauserverclient/models/property_decorators.py
```diff
@@ -46,3 +46,34 @@ def wrapper(self, value):
...
+
+    return wrapper
+
+
+def property_is_int(min, max=2**31):
+    def property_type_decorator(func):
+        @wraps(func)
+        def wrapper(self, value):
+            if min is None:
+                return func(self, value)
```
LGraber: I don't get this? Why can't I set the min to None and the max to some number (like 100). Not saying you are doing this right now but this code would break. Just do the if "min is not None and value < min" and similarly "if max is not None and value > max". No?

tableauserverclient/server/request_factory.py
```diff
@@ -129,6 +129,54 @@ def create_req(self, project_item):
...
+        schedule_element = ET.SubElement(xml_request, 'schedule')
+        schedule_element.attrib['name'] = schedule_item.name
+        schedule_element.attrib['priority'] = str(schedule_item.priority)
+        schedule_element.attrib['type'] = schedule_item.schedule_type
+        schedule_element.attrib['executionOrder'] = schedule_item.execution_order
+        interval_item = schedule_item.interval_item
+        schedule_element.attrib['frequency'] = interval_item._frequency
+        frequency_element = ET.SubElement(schedule_element, 'frequencyDetails')
+        frequency_element.attrib['start'] = str(interval_item.start_time)
+        if interval_item.end_time:
```
LGraber: add hasattr(interval_item, end_time)

LGraber:
I gave 4-5 more comments but they should be small and easy fixes. Then I will be ready to signoff. There might be more to cleanup but I really like the .5 and .25 as it just really cleans up some of the code and makes this a bit more sensible for me. Thanks!

tableauserverclient/models/property_decorators.py
t8y8: I could enable that, but then how would we say "this should be an int but not checked within a range"

`property_is_int(None)` is nice and clean to say that.
`property_is_int(None, 100)` is ambiguous with the above call because both min and max are required positional arguments, so no nice signature :(

I could:
1. Make a keyword argument that is `property_is_int(range=False)` or some similar keyword.

LGraber: I am fine with accepting a pair using range=(None,100). It is more readable. I am good with that.

tableauserverclient/models/interval_item.py
LGraber: I want to return what we set. We have made the "Occurence" disappear. Why are we putting it back? if I write:
foo.interval = .25
and then do
x = foo.interval

x should equal .25

We can make the serialization work without exposing this to users.

tableauserverclient/models/property_decorators.py
t8y8: Done.

```
def property_is_int(range):
    def property_type_decorator(func):
        @wraps(func)
        def wrapper(self, value):
            if range is None:
                if isinstance(value, int):
                    return func(self, value)

            min, max = range

            if value < min or value > max:
                error = "Invalid priority defined: {}.".format(value)
                raise ValueError(error)

            return func(self, value)

        return wrapper

    return property_type_decorator
```

`property_is_int(range=None)` means "int but it can be whatevs"
`property_is_int(range=(1,100))` means "int between 1 and 100"

LGraber: Okay. I am fine with this. If it is ever actually a need, we can debate a case with only a max or only a min but we have no case so I am not going to debate just for the sake of debating. I like the syntax. Thanks!

tableauserverclient/models/interval_item.py
t8y8: I think that's going to be a fairly large change.

We'd need a serializer for each IntervalType maybe? 

Or... I'd have to think more about it, when it's not so late.

LGraber: We could do it in the serializer or you can have an internal function on each of them which returns this pair. We can call internal functions from our serializer ... this is already a pattern we use. But the value I set for interval should be the value I get back for it (and I should be able to directly set it, ie not just in the constructor).

t8y8: I like the latter

`interval_item.interval_pairs()` or whatever.

One-line change in the serializer, and then the internal function is easy to write as well. 
I'll give it a go tomorrow.

tableauserverclient/server/request_factory.py
```diff
@@ -141,14 +141,16 @@ def create_req(self, schedule_item):
...
-        if interval_item.end_time:
-            frequency_element.attrib['end'] = str(interval_item.end_time)
-        if interval_item.interval:
-            intervals_element = ET.SubElement(frequency_element, 'intervals')
-            for interval in interval_item.interval:
-                expression, value = interval
-                single_interval_element = ET.SubElement(intervals_element, 'interval')
-                single_interval_element.attrib[expression] = value
+        if hasattr(interval_item, 'end_time'):
+            if interval_item.end_time:
```
LGraber: you could put this on the hasattr line with an "and". I assume python does not evaluate the second expression if the first expression is false. Don't need to fix it now but seems more concise

tableauserverclient/server/request_factory.py
```diff
@@ -141,14 +141,16 @@ def create_req(self, schedule_item):
...
+                frequency_element.attrib['end'] = str(interval_item.end_time)
+        if hasattr(interval_item, 'interval'):
+            if interval_item.interval:
```
LGraber: same here

tableauserverclient/server/request_factory.py
```diff
@@ -167,13 +169,15 @@ def update_req(self, schedule_item):
             schedule_element.attrib['frequency'] = interval_item._frequency
         frequency_element = ET.SubElement(schedule_element, 'frequencyDetails')
         frequency_element.attrib['start'] = str(interval_item.start_time)
-        if interval_item.end_time:
-            frequency_element.attrib['end'] = str(interval_item.end_time)
+        if hasattr(interval_item, 'end_time'):
+            if interval_item.end_time:
```
LGraber: here

LGraber:
I think there is probably some more cleanup we can do but I am happy with where we have ended up. :rocket:

t8y8:
:tada:

PR #59, opened 2016-10-12
Prototype attribute refactor

t8y8:
Here's a proposal for how to make the _set_values implementation a bit cleaner.

Now all we have to do to get new attributes on a model are to add it to the parser and in the whitelist

Yeah, I know, locals is gross, but it's just a prototype :)

### DIFF
--- a/tableauserverclient/models/datasource_item.py
+++ b/tableauserverclient/models/datasource_item.py
@@ -13,6 +13,24 @@ def _get_tags(datasource_xml):
     return tags
 
 
+def _get_project_id(datasource_xml):
+    project_elem = datasource_xml.find('.//t:project', namespaces=NAMESPACE)
+    if project_elem is not None:
+        return project_elem.get('id', None)
+
+
+def _get_project_name(datasource_xml):
+    project_elem = datasource_xml.find('.//t:project', namespaces=NAMESPACE)
+    if project_elem is not None:
+        return project_elem.get('name', None)
+
+
+def _get_owner_id(datasource_xml):
+    owner_elem = datasource_xml.find('.//t:owner', namespaces=NAMESPACE)
+    if owner_elem is not None:
+        return owner_elem.get('id', None)
+
+
 class DatasourceItem(object):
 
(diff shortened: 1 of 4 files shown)


t8y8:
OH I WENT CRAZY.

There's a test failure because this will re-parse from the XML but it lets us specify all the attributes for set_values and in the parse all in one place:
```
    SCHEMA = {'_id': lambda x: x.get('id', None),
              'name': lambda x: x.get('name', None),
              '_datasource_type': lambda x: x.get('type', None),
              '_content_url': lambda x: x.get('contentUrl', None),
              '_created_at': lambda x: x.get('createdAt', None),
              '_updated_at': lambda x: x.get('updatedAt', None),
              '_tags': _get_tags,
              'project_id': lambda x: x.find('.//t:project', namespaces=NAMESPACE).get('id', None),
              '_project_name': lambda x: x.find('.//t:project', namespaces=NAMESPACE).get('name', None),
               'owner_id': lambda x: x.find('.//t:owner', namespaces=NAMESPACE).get('id', None),
              }
```

And then in _parse_element becomes:
```
    @staticmethod
    def _parse_element(datasource_xml):
        attribs = {}
        for attribute, getter in DatasourceItem.SCHEMA.items():
            attribs[attribute] = getter(datasource_xml)
        return attribs
```

and _set_values becomes:
```
    def _set_values(self, ds_attributes):
        print(ds_attributes)
        for attribute in DatasourceItem.SCHEMA:
            setattr(self, attribute, ds_attributes[attribute])
```

t8y8:
But can I collapse the parser and serializer into one place so when we add attributes it's all automatic?

That might be more trouble than it's worth, but hey, prototypes are prototyptes

t8y8:
Where would you put it? We keep it there today inside of the staticmethod `from_response`

I flirted with the idea of a parser class for each but I was lazy

t8y8:
Retiring this PR, is was an interesting exploration and we may revisit some day, but I don't intend to push this any farther at the moment

Issue #60, opened 2016-10-13
Deleting a site results in a signout error


mpope-tableau:
I'm deleting a site using code like this:

```
tableau_auth = TSC.TableauAuth(args.username, password, site=args.contentUrl)
with server.auth.sign_in(tableau_auth):
    server.sites.delete(site_id)
    print("\nSite deleted.")
```
Note that in order to delete a site, I have to sign into that site by including the **contentUrl**. (The **site_id** value was fetched earlier in the program from which I take this snippet.) This code works--the site is deleted. 

_However_, at the termination of the **with** block, the code implicitly attempts to sign out from the site. But the site doesn't exist any more, so the signout fails and throws all sorts of errors.

LGraber:
we will have to take a look at this. The server should allow you to signout after you delete the site. If you need to work around this ... don't use the "with" statement (and thus don't call signout). Thanks. Let's keep this open to track.

t8y8:
This is one edge case that results from our REST API design, it would repro without the library.

After you delete the site your session is invalid (because sessions are tied to a site), and so every subsequent call will throw a 401.

We've known about it, and I swear documented it before.

RussTheAerialist:
Even still, we should handle this situation better than throwing an exception out of the context handler.

t8y8:
Yeah it's not a great experience... calls throw a generic 401.

I haven't looked at the __exit__ but we could try and catch it there... except it could be throwing a 401 for a valid reason. Actually, if you call sign out in the with block I wonder if this repros.

Or, the __exit__ could swallow all 401s because it's assumed you're signing out and don't care?

LGraber:
I understand we may chose to "hide" this in the client code but this is a bug in our server code. I would agree that all other actions on the site should fail, but signout should not fail just because I deleted the site. That call should succeed because there is no reason it can't. I do not want to hide this in our client code until we have investigated fixing this behavior in our server. I feel confident that we can fix this in our server and do not want any change to our library yet.

t8y8:
How? The site gets deleted the session is invalid.

The 401 is real -- what would you have Server do instead?

LGraber:
Let's talk about this offline first.

t8y8:
@LGraber had an idea on how to make sign_out a no-op without missing valid exceptions.

@mpope-tableau can I swing by and try a monkey-patch in your code?

LGraber:
[...]

PR #61, opened 2016-10-14
Fix for issue #60

t8y8:
Make Sign Out a no-op when you are already signed out.

Here is a repro script that passes after and fails before the change:
```
import argparse
import getpass
import logging

def main():

    parser = argparse.ArgumentParser(description='Publish a workbook to server.')
    parser.add_argument('--server', '-s', required=True, help='server address')
    parser.add_argument('--username', '-u', required=True, help='username to sign into server')
    parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
                        help='desired logging level (set to error by default)')

    args = parser.parse_args()

    password = getpass.getpass("Password: ")

    # Set logging level based on user input, or error by default
    logging_level = getattr(logging, args.logging_level.upper())
    logging.basicConfig(level=logging_level)

    # Step 1: Sign in to server.
    tableau_auth = TSC.TableauAuth(args.username, password)

        new_site = TSC.SiteItem(name='blorp', content_url='blorp')
        yay = server.sites.create(new_site)

    tableau_auth = TSC.TableauAuth(args.username, password, site='blorp')

        sites = server.sites.get()
        print([i for i in sites])
        server.sites.delete(yay.id)

if __name__ == '__main__':
    main()
```

### DIFF
--- a/tableauserverclient/server/endpoint/auth_endpoint.py
+++ b/tableauserverclient/server/endpoint/auth_endpoint.py
@@ -42,7 +42,7 @@ class Auth(Endpoint):
     def sign_out(self):
         url = "{0}/{1}".format(self.baseurl, 'signout')
         # If there are no auth tokens you're already signed out. No-op
-        if not self.parent_srv._auth_token:
+        if not self.parent_srv.is_signed_in():
             return
         self.post_request(url, '')
         self.parent_srv._clear_auth()
--- a/tableauserverclient/server/endpoint/auth_endpoint.py
+++ b/tableauserverclient/server/endpoint/auth_endpoint.py
@@ -41,6 +41,9 @@ class Auth(Endpoint):
 
     def sign_out(self):
         url = "{0}/{1}".format(self.baseurl, 'signout')
+        # If there are no auth tokens you're already signed out. No-op
+        if not self.parent_srv._auth_token:
+            return
         self.post_request(url, '')
         self.parent_srv._clear_auth()
         logger.info('Signed out')
--- a/tableauserverclient/server/endpoint/sites_endpoint.py
+++ b/tableauserverclient/server/endpoint/sites_endpoint.py
@@ -59,7 +59,12 @@ class Sites(Endpoint):
             raise ValueError(error)
         url = "{0}/{1}".format(self.baseurl, site_id)
         self.delete_request(url)
-        logger.info('Deleted single site (ID: {0})'.format(site_id))
+        # Deleting the site also signs you out on the Server.
+        # Here we check that you're deleting the site you are signed in to
+        # (Which is true now but may not always be) and then clear auth tokens
+        if site_id == self.parent_srv._site_id:
+            self.parent_srv._clear_auth()
+        logger.info('Deleted single site (ID: {0}) and signed out'.format(site_id))
 
     # Create new site
     def create(self, site_item):
--- a/tableauserverclient/server/endpoint/sites_endpoint.py
+++ b/tableauserverclient/server/endpoint/sites_endpoint.py
@@ -62,6 +62,7 @@ class Sites(Endpoint):
         # If we deleted the site we are logged into
         # then we are automatically logged out
         if site_id == self.parent_srv.site_id:
+            logger.info('Deleting current site and clearing auth tokens')
             self.parent_srv._clear_auth()
         logger.info('Deleted single site (ID: {0}) and signed out'.format(site_id))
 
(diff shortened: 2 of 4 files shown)


tableauserverclient/server/endpoint/sites_endpoint.py
```diff
@@ -59,7 +59,12 @@ def delete(self, site_id):
             raise ValueError(error)
         url = "{0}/{1}".format(self.baseurl, site_id)
         self.delete_request(url)
-        logger.info('Deleted single site (ID: {0})'.format(site_id))
+        # Deleting the site also signs you out on the Server.
+        # Here we check that you're deleting the site you are signed in to
+        # (Which is true now but may not always be) and then clear auth tokens
+        if site_id == self.parent_srv._site_id:
```
LGraber: we have a property getter on the parent_srv object. Why not use that instead of the internal member?

tableauserverclient/server/endpoint/auth_endpoint.py
```diff
@@ -41,6 +41,9 @@ def sign_in(self, auth_req):
 
     def sign_out(self):
         url = "{0}/{1}".format(self.baseurl, 'signout')
+        # If there are no auth tokens you're already signed out. No-op
+        if not self.parent_srv._auth_token:
```
LGraber: This feels like it should be a method on the server object to indicate if it is signed in. Looking at what you wrote, I understand that there is no "private" in python but we should honor things when they make sense. A method on the server liked signed_in() which returns a boolean would feel cleaner.

tableauserverclient/server/endpoint/sites_endpoint.py
```diff
@@ -59,7 +59,12 @@ def delete(self, site_id):
             raise ValueError(error)
         url = "{0}/{1}".format(self.baseurl, site_id)
         self.delete_request(url)
-        logger.info('Deleted single site (ID: {0})'.format(site_id))
+        # Deleting the site also signs you out on the Server.
+        # Here we check that you're deleting the site you are signed in to
+        # (Which is true now but may not always be) and then clear auth tokens
```
LGraber: This comment doesn't make sense, "Which is true now". Our code does not enforce this. It might be true that this only works in this case but I think you can leave that out. Simply put if you delete the site you are signed into, you are automatically signed out.

t8y8: Tableau Server enforces this, you MUST be signed into the site you are deleting or you get a 403 that says "You can't delete a resource on another site"

tableauserverclient/server/endpoint/auth_endpoint.py
t8y8: When you say "signed in" are you just defining that as having a token? 
The token isn't guaranteed to be valid just because it exists (the crux of this issue) which means you aren't actually 'signed in' -- but if that's ok I can have something like that

tableauserverclient/server/endpoint/sites_endpoint.py
t8y8: The getter checks if you have a site_id, which we don't because we remove it in the DELETE call when we make it, and then complains that you aren't signed in.

```
    @property
    def site_id(self):
        if self._site_id is None:
            error = 'Missing site ID. You must sign in first.'
            raise NotSignedInError(error)
        return self._site_id
```

tableauserverclient/server/endpoint/sites_endpoint.py
t8y8: Got it, will update

tableauserverclient/server/endpoint/auth_endpoint.py
LGraber: That is what I am saying. Our logic right now is that we assume if we have a token we are signed in.  You still might fail because the token expired or something else, but from the client's standpoint, we think we are signed in. The exact same check you are doing ... just move it so we are not accessing private member variables.

tableauserverclient/server/endpoint/sites_endpoint.py
LGraber: We clear the auth info (token and site_id) after this check. This code is in the delete method.

tableauserverclient/server/endpoint/sites_endpoint.py
```diff
@@ -59,10 +59,9 @@ def delete(self, site_id):
...
         url = "{0}/{1}".format(self.baseurl, site_id)
         self.delete_request(url)
-        # Deleting the site also signs you out on the Server.
-        # Here we check that you're deleting the site you are signed in to
-        # (Which is true now but may not always be) and then clear auth tokens
-        if site_id == self.parent_srv._site_id:
+        # If we deleted the site we are logged into
+        # then we are automatically logged out
+        if site_id == self.parent_srv.site_id:
             self.parent_srv._clear_auth()
```
LGraber: Add a log statement here.

LGraber:
🚀 Just the one request for an extra line of logging. Looks good to me

Issue #50, opened 2016-10-05
Site creation fails when using Japanese characters for site name


benlower:
Trying to create a site with the following:
```python
    sites = "遊び場"
    new_site = TSC.SiteItem(name=site, content_url=site.replace(" ", ""), admin_mode=TSC.SiteItem.AdminMode.ContentAndUsers)
    server.sites.create(new_site)
```

This errors with:

	409002: Resource Conflict
	A site already exists with the content URL '遊び場'

The site doesn't exist on my server.

RussTheAerialist:
@benlower Did you try this using postman directly?  That error is coming directly from the REST API, so I don't think this is a bug in the server client but in the REST api.  Can you verify via postman against the rest api directly, and update this issue once you have?

t8y8:
@benlower specify a different content_url, that content_url is becoming "" (because you can't have Japanese characters as the content_url) and "" is the ID if the default site.

t8y8:
Idea: we could validate that content_url only contains valid ascii characters and throws before making the call to Server

```
    @content_url.setter
    @property_not_nullable
    def content_url(self, value):
        if not is_ascii(value): raise ValueError
        self._content_url = value
```

And/or make it a new decorator

RussTheAerialist:
I think making it a decorator would be better.  Something like "OnlyAscii" or something.

t8y8:
@RussTheAerialist, I think according to SO this is the best py2/py3 way to do it:

```
def property_is_url_safe(func):
    @wraps(func)
    def wrapper(self, value):
        try:
            value.decode('ascii')
        except UnicodeDecodeError:
            raise ValueError("content_url must be ascii only and url safe")
        return func(self, value)

    return wrapper
```

LGraber:
Why do you call is "is_url_safe"? That doesn't really describe what it is doing (I still have to go to the code to figure out that safe == ascii). Are you planning on adding other things here? It is also an odd property name given that it will be on a property which is not an url.  Besides that I am fine with it after talking with Russell

RussTheAerialist:
Maybe have it be a property that takes an argument and call it something like "allowed_characters"?  And then define "allowed_content_url_characters" somewhere which includes a string (or list/tuple) of all of the allowed characters in a content url?

t8y8:
Resolved in #64

PR #64, opened 2016-10-18
Fix for issue #50 Add regex validator to content_url

t8y8:
Content_urls can only be alphanumeric or underscores and dashes. Fix is to validate it client side because the Server error is confusing and the client should just prevent invalid urls from going through.

* Added a generic regex validator for properties and apply it to the site_model
* Updated site_item_tests with some unicode and valid/invalid cases

### DIFF
--- a/tableauserverclient/models/property_decorators.py
+++ b/tableauserverclient/models/property_decorators.py
@@ -87,15 +87,15 @@ def property_is_int(range):
     return property_type_decorator
 
 
-def property_matches(regex_to_match):
+def property_matches(regex_to_match, error):
 
-    COMPILED_RE = re.compile(regex_to_match)
+    compiled_re = re.compile(regex_to_match)
 
     def wrapper(func):
         @wraps(func)
         def validate_regex_decorator(self, value):
-            if not COMPILED_RE.match(value):
-                raise ValueError("content_url can contain only ascii letters, numbers, dashes, and underscores")
+            if not compiled_re.match(value):
+                raise ValueError(error)
             return func(self, value)
         return validate_regex_decorator
     return wrapper
--- a/tableauserverclient/models/property_decorators.py
+++ b/tableauserverclient/models/property_decorators.py
@@ -1,3 +1,4 @@
+import re
 from functools import wraps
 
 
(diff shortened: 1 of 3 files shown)


tableauserverclient/models/property_decorators.py
```diff
@@ -84,3 +85,17 @@ def wrapper(self, value):
         return wrapper
 
     return property_type_decorator
+
+
+def property_matches(regex_to_match):
+
+    COMPILED_RE = re.compile(regex_to_match)
```
RussTheAerialist: This isn't a constant, so it should be compiled_re

tableauserverclient/models/property_decorators.py
```diff
@@ -84,3 +85,17 @@ def wrapper(self, value):
...
+
+    def wrapper(func):
+        @wraps(func)
+        def validate_regex_decorator(self, value):
+            if not COMPILED_RE.match(value):
+                raise ValueError("content_url can contain only ascii letters, numbers, dashes, and underscores")
```
RussTheAerialist: Make the message an argument that gets passed in because if we want to reuse this, then the error message would be wrong.

tableauserverclient/models/site_item.py
```diff
@@ -1,8 +1,12 @@
 import xml.etree.ElementTree as ET
-from .property_decorators import property_is_enum, property_is_boolean, property_not_empty, property_not_nullable
+from .property_decorators import (property_is_enum, property_is_boolean, property_matches,
+                                  property_not_empty, property_not_nullable)
 from .. import NAMESPACE
 
 
+VALID_CONTENT_URL_RE = r"^[a-zA-Z0-9_\-]*$"
```
RussTheAerialist: Do we ever want the content url to be empty?   I guess it is when we have the default site, right?

test/test_site_model.py
```diff
@@ -1,3 +1,5 @@
+# coding=utf-8
```
RussTheAerialist: This should be part of our template for new files!

tableauserverclient/models/site_item.py
t8y8: Correct, it needs to validate "" or all of our models fail when reading responses from the default site

Issue #67, opened 2016-10-20
Why do I need to populate a Group with Users to add a new user?


tableauembeddingteam:
After about an hour of work, I am finally able to successfully add a new user and then add them to an existing group. However, I was perplexed by receiving an error about the Group not being populated:

My initial code:

with server.auth.sign_in(auth):
    user = TSC.UserItem('stone_cold', 'Publisher')
    new_user = server.users.add(user)
    gs, pi = server.groups.get()
    for g in gs:
        if g.name == 'Ballers':
            server.groups.add_user(g, new_user.id) 

Working code:
with server.auth.sign_in(auth):
    user = TSC.UserItem('stone_cold', 'Publisher')
    new_user = server.users.add(user)
    gs, pi = server.groups.get()
    for g in gs:
        if g.name == 'Ballers':
            server.groups.populate_users(g)
            server.groups.add_user(g, new_user.id)

Why do I need to use the populate_users() method on the Group just to add a user to it? As a second issue, why is populate_users() a Groups endpoint method and not a method of the GroupItem itself, where it could be called automatically prior to the attempt to add?

RussTheAerialist:
I'll have to investigate why we need the populate_users call (not familiar with this code)
but I can answer the part about why we do not do it automatically prior to attempting to add.

This is a conscious design choice.  We do not want to make calls that you do not specifically ask for.  Yes it means there is a little bit of extra code, but we feel like it's better to be specific about why we are making calls to the api.

RussTheAerialist:
Okay quickly looking over the code, this is a side-effect of trying to keep the user list accurate, so it made an assumption that the user list was already populated.  This is a bug.

PR #68, opened 2016-10-20
Making the user list operations a no-op userlist not populated

RussTheAerialist:
Refactored the calls to the endpoint and added the user list manipulation into a try/catch for the unpopulated error.

### DIFF
--- a/tableauserverclient/server/endpoint/groups_endpoint.py
+++ b/tableauserverclient/server/endpoint/groups_endpoint.py
@@ -1,5 +1,6 @@
 from .endpoint import Endpoint
 from .exceptions import MissingRequiredFieldError
+from ...models.exceptions import UnpopulatedPropertyError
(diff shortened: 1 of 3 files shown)


tableauserverclient/server/endpoint/groups_endpoint.py
```diff
@@ -1,5 +1,6 @@
 from .endpoint import Endpoint
 from .exceptions import MissingRequiredFieldError
+from ...models.exceptions import UnpopulatedPropertyError
```
t8y8: I'm not even sure the import rules for triple . -- is this above or below the .. imports? :(

RussTheAerialist: ... is the same as ../.. on the command line.  I know it's a bit annoying, but we need relative imports.

RussTheAerialist:
fixes #67

Issue #66, opened 2016-10-20
Groups endpoint missing Create method


tableauembeddingteam:
Doesn't appear to be any way to create a new Group.

RussTheAerialist:
Looks like it's a feature we missed on the initial implementation.

RussTheAerialist:
The full implementation which includes importing from AD will not be in the 0.2 release due to time.

irwando:
Group creation was added some time ago, not sure why this is still open. Closing.

PR #69, opened 2016-10-20
Bugfix 66 add create group basic

RussTheAerialist:
@bryantbhowell - I think you are looking for some ability to create groups.  Groups are handled in two separate ways even though it's the same create endpoint in the server.

This handles creating basic groups, but AD group imports are a bit more complicated (because they are done in the background).  Do you need AD imports immediately?


test/assets/group_create_async.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<tsResponse xmlns="http://tableau.com/api"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://tableau.com/api http://tableau.com/api/ts-api-2.3.xsd">
+    <job id="job-id" mode="Asynchronous" type="GroupImport"
```
LGraber: This is there for future tests? We don't currently let people trigger an async import.

tableauserverclient/server/endpoint/groups_endpoint.py
```diff
@@ -45,6 +45,14 @@ def delete(self, group_id):
         self.delete_request(url)
         logger.info('Deleted single group (ID: {0})'.format(group_id))
 
+    def create(self, group_item):
+        url = self.baseurl
+        create_req = RequestFactory.Group.create_req(group_item)
+        server_response = self.post_request(url, create_req)
+        new_group = GroupItem.from_response(server_response.content)[0]
```
LGraber: Not sure what our standard is but we could probably just put "return" here.

LGraber:
We should have a quick chat about what the api should look like for importing an AD group. I don't want to ship something too quickly and then have the signature not work for the future. Besides that ... it looks good.

test/assets/group_create_async.xml
RussTheAerialist: Yup, that's for a future test when I implement import() (see comment above about AD import)

tableauserverclient/server/endpoint/groups_endpoint.py
RussTheAerialist: fixed

test/test_group.py
```diff
@@ -140,3 +142,13 @@ def test_remove_user_missing_group_id(self):
         single_group._users = []
         self.assertRaises(TSC.MissingRequiredFieldError, self.server.groups.remove_user, single_group,
                           '5de011f8-5aa9-4d5b-b991-f462c8dd6bb7')
+
+    def test_create_group(self):
+        with open(CREATE_GROUP, 'rb') as f:
+            response_xml = f.read().decode('utf-8')
+        with requests_mock.mock() as m:
+            m.post(self.baseurl, text=response_xml)
+            group_to_create = TSC.GroupItem('test')
```
t8y8: Can I be high maintenance and request a non-ascii name in a test as well?

I just remember some bugs in early sample code and libs prototypes around this.

LGraber:
:rocket:

test/test_group.py
RussTheAerialist: Sure, any suggestions on appropriate words or should I just start finding interesting looking utf-8 characters?

t8y8: Got a favorite foreign language?

I used "蚵仔煎" because it's my favorite street food in the entire world...

t8y8:
🚀 🚀

RussTheAerialist:
Addresses creation of groups for #66 .  Importing is not currently supported.

PR #72, opened 2016-10-21
Adding pagination sample

RussTheAerialist:
This sample has a generator class that makes it easy to iterate over all of the pages of one of the `.get()` methods.  We might want to consider making this class (or an updated version of it) part of the library.

### DIFF
--- a/samples/pagination_sample.py
+++ b/samples/pagination_sample.py
@@ -1,6 +1,6 @@
 ####
 # This script demonstrates how to use pagination item that is returned as part
-# of mayn of the .get() method calls.
+# of many of the .get() method calls.
 #
 # This script will iterate over every workbook that exists on the server using the
 # pagination item to fetch additional pages as needed.
@@ -51,7 +51,7 @@ class pagination_generator(object):
 
 def main():
 
-    parser = argparse.ArgumentParser(description='Explore workbook functions supported by the Server API.')
+    parser = argparse.ArgumentParser(description='Return a list of all of the workbooks on your server')
     parser.add_argument('--server', '-s', required=True, help='server address')
     parser.add_argument('--username', '-u', required=True, help='username to sign into server')
     parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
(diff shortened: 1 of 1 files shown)


samples/pagination_sample.py
```diff
@@ -0,0 +1,79 @@
+####
+# This script demonstrates how to use pagination item that is returned as part
+# of mayn of the .get() method calls.
```
benlower: mayn?

samples/pagination_sample.py
```diff
@@ -0,0 +1,79 @@
...
+    def _load_next_page(self, current_item_list, last_pagination_item):
+        next_page = last_pagination_item.page_number + 1
+        opts = TSC.RequestOptions(pagenumber=next_page, pagesize=last_pagination_item.page_size)
+        current_item_list, last_pagination_item = self._callback(self._server, opts)
+        return current_item_list, last_pagination_item
+
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Explore workbook functions supported by the Server API.')
```
benlower: description needs updated for this sample

samples/pagination_sample.py
```diff
@@ -0,0 +1,79 @@
...
+    logging_level = getattr(logging, args.logging_level.upper())
+    logging.basicConfig(level=logging_level)
+
+    # SIGN IN
+
+    tableau_auth = TSC.TableauAuth(args.username, password)
+    server = TSC.Server(args.server)
+
+    with server.auth.sign_in(tableau_auth):
+        generator = pagination_generator(server.workbooks.get)
```
LGraber: Like we talked about ... this should take the options object so that we can support filters. I know I mentioned the optional additional arguments in case a get method takes more than just options but for now, you probably don't need to add this. Let's put the "paginator" in our server directory so that is it is a supported utility and then that will simplify our code. Would like Ben to comment on the name.

LGraber:
Something is better than nothing. Would have liked to plumb through the options but I am good putting this in and then fixing it with the one we will officially support (as well as dealing with Tyler's comment)

Issue #73, opened 2016-10-25
Can't set password for new user?


mpope-tableau:
Sorry if this is an old issue. (I searched but didn't find it.) I can create a new user this way:

`new_user = server.users.add(name, siteRole)`

The **add** function has this signature:

`def add(self, user_item)`

The **UserItem** constructor explicitly sets the password property like this:

```
def __init__(self, name, site_role, auth_setting=None):
        self._auth_setting = None
        self._domain_name = None
        self._external_auth_user_id = None
        self._id = None
        self._last_login = None
        self._name = None
        self._site_role = None
        self._workbooks = None
        self.email = None
        self.fullname = None
        self.password = None
```

And as near as I can tell (??) there is no setter in the **UserItem** class for a password property. 

I tried doing this, which does compile and run:

```
new_user.password = args.newpassword
print("\nPassword changed to {0}".format(new_user.password))
new_user = server.users.update(new_user)
```

The `server.users.update(new_user)` call did result in a **PUT** request, but per Fiddler, the **PUT** request had only this in it (no password):

`<tsRequest><user authSetting="ServerDefault" siteRole="SiteAdministrator" /></tsRequest>`

In the **UserRequest** class (`request_factory.py`), there's a definition for an **update_req** method that would seem to take a password:

```
def update_req(self, user_item, password=''):
...
    if password:
        user_element.attrib['password'] = password
```    

But I can't tell how to reach this method directly. So I'm stuck with being able to create a user but not assign a password. (And can't sign in as that user.)

RussTheAerialist:
Can you try to do the following to the UserItem before you create it on the server:

```python
new_user_item.password = 'foo'
```

The only ones with setters are the ones that begin with `_`.  Otherwise, you can just set the attribute directly.

I'll take a look at why the update isn't working, also

t8y8:
Update doesn't work because it's looking at the password element passed in via the function signature, not from the UserItem.

If we want to pull from the model it should be:
```python
...
if user_item.password:
    user_element.attrib['password'] = password
return ET.tostring(xml_request)
```

And/or we'd need to update the user endpoint `.update()` signature to take a password and pass it down.
```python
def update(self, user_item, password=None):
```

I'm not sure I like us storing a password on the users model unless we set it to `None` after every access or something like that, it is plain text after all.

@RussTheAerialist -- I'm thinking to just plumb the password through the `update` method and into the RequestFactory. Thoughts?

RussTheAerialist:
[...]

PR #75, opened 2016-10-25
Fix Issue #73: Plumb Password to Update User

t8y8:
We were checking against the password parameter that wasn't accessible via the update endpoint.

This plumbs password through and optionally serializes if if it's present in the update request.

I removed all traces of password from the user model to remove any temptation to store a long lived plain text password there.

[Manual Test Screenshot Proof](https://www.dropbox.com/s/9h72u1bx2wvunns/Screenshot%202016-10-25%2009.45.50.png?dl=0)

The new signature would be:
```python
user = TSC.UserItem(name='mikepope', site_role='Publisher')
server.users.update(user, password='secret')
```
/cc @RussTheAerialist @mpope-tableau

### DIFF
--- a/tableauserverclient/models/user_item.py
+++ b/tableauserverclient/models/user_item.py
@@ -28,7 +28,6 @@ class UserItem(object):
         self._workbooks = None
         self.email = None
         self.fullname = None
-        self.password = None
         self.name = name
         self.site_role = site_role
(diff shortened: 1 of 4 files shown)


t8y8:
@mpope-tableau totally tested this locally too.

PR #84, opened 2016-10-28
Implement get server info

RussTheAerialist:
Quickly implemented serverInfo endpoint.

### DIFF
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -1,6 +1,7 @@
 from .exceptions import ServerResponseError
 import logging
 
+
 logger = logging.getLogger('tableau.endpoint')
 
 Success_codes = [200, 201, 204]
@@ -14,6 +15,16 @@ class Endpoint(object):
     def _check_status(server_response):
         if server_response.status_code not in Success_codes:
             raise ServerResponseError.from_response(server_response.content)
+
+    def get_unauthenticated_request(self, url, request_object=None):
+        if request_object is not None:
+            url = request_object.apply_query_params(url)
+        server_response = self.parent_srv.session.get(url, **self.parent_srv.http_options)
+        self._check_status(server_response)
+        if server_response.encoding:
+            logger.debug(u'Server response from {0}:\n\t{1}'.format(
+                url, server_response.content.decode(server_response.encoding)))
+        return server_response
 
     def get_request(self, url, request_object=None):
         if request_object is not None:
--- a/tableauserverclient/server/endpoint/server_info_endpoint.py
+++ b/tableauserverclient/server/endpoint/server_info_endpoint.py
@@ -1,0 +1,21 @@
+from .endpoint import Endpoint
+from ...models import ServerInfoItem
+import logging
+
+logger = logging.getLogger('tableau.endpoint.server_info')
+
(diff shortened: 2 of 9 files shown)


tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -15,6 +16,16 @@ def _check_status(server_response):
         if server_response.status_code not in Success_codes:
             raise ServerResponseError.from_response(server_response.content)
 
+    def get_unauthenticated_request(self, url, request_object=None):
+        if request_object is not None:
+            url = request_object.apply_query_params(url)
+        server_response = self.parent_srv.session.get(url, **self.parent_srv.http_options)
+        self._check_status(server_response)
+        if server_response.encoding:
```
t8y8: Question -- why check for encoding? Will it ever not be there?

Does this mean if there isn't an encoding there is nothing to log?

RussTheAerialist: I don't remember why we needed to do this, but it's the pattern that all of the others follow.

tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -15,6 +16,16 @@ def _check_status(server_response):
         if server_response.status_code not in Success_codes:
             raise ServerResponseError.from_response(server_response.content)
 
+    def get_unauthenticated_request(self, url, request_object=None):
```
LGraber: Feels like we should probably have a "_get_request" method and these two just call the underlying one. Unauthenticated can call directly. Normal can create the headers list and pass it in. The rest of the code is identical.

tableauserverclient/server/endpoint/server_info_endpoint.py
```diff
@@ -0,0 +1,21 @@
+from .endpoint import Endpoint
+from ...models import ServerInfoItem
+import logging
+
+logger = logging.getLogger('tableau.endpoint.server_info')
+
+
+class ServerInfo(Endpoint):
+    def __init__(self, parent_srv):
+        super(Endpoint, self).__init__()
```
LGraber: We need to go through and update all of these. This is not the correct way to do it. "Endpoint" should be "ServerInfo". It doesn't matter since Endpoint and object both have empty constrctors. Which leaves us with ... we should just delete all of these.

LGraber:
Couple of comments for cleanup later. You already submitted so they can be dealt with the next time

t8y8:
@LGraber I think I noticed another weird thing (unrelated to his change), I might be able to incorporate your feedback into that

tableauserverclient/server/endpoint/endpoint.py
RussTheAerialist: I was planning to do that in a separate checkin anyway.  There is a bunch of overlap that I think I can reduce, but it would have been noise in this change.

tableauserverclient/server/endpoint/server_info_endpoint.py
RussTheAerialist: This was another thing that I noticed while doing this change.  EVERY Endpoint takes parent_srv, so I'm going to rip all of the custom ctors out and just allow the base object's ctor which doesn't currently take a parent_srv take the parent_srv

PR #90, opened 2016-10-31
Implement Pager for auto-paging requests

t8y8:
Here's a first draft based on @RussTheAerialist's code in `pagination_sample.py`.

Currently it:
- Takes an endpoint as its argument instead of a `get` function, this is because these calls only apply to `get` anyway, so let's keep it clean.
- Takes an options object and seeds the filters/sorts with it, as well as the initial page size
- Can be used as an iterator

I would like it to:

- Not do everything in the `__call__` method. I _think_ we can move that to a helper or to `__init__`
    - We would make the first page call on instantiation, which populates some properties that could be useful (like `__len__`)
- Since we know the `total_avaliable` we might be able to implement `__len__`. This gives users the ability to check the length to see if they really want to iterate over it or not.
- I need to read more to see if I need to worry about `__next__` or not.
- Add tests

This script works (tested manually):
```python

auth_stuff = TSC.TableauAuth(username='USERNAME', password='PASSWORD')

server = TSC.Server("http://SERVER")
server.auth.sign_in(auth_stuff)

options = TSC.RequestOptions()
options.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
                                         "WORKBOOKNAME"))

# Call with a filter
wbs = TSC.server.Pager(server.workbooks, options)
print(len([i for i in wbs[)) # prints 1

# Get them all!
wbs = TSC.server.Pager(server.workbooks)
print(len([i for i in wbs])) # prints 1086 on my test server, that's all of em!
server.auth.sign_out()
```

Open Questions:

1. Where should it live? I'm not sure Server is the right place
2. Thoughts on implementing `__len__`?
3. Thoughts on implementing the primary logic somewhere other than `__call__`? It might just be me who thinks it's weird.

/cc @RussTheAerialist @LGraber

### DIFF
--- a/tableauserverclient/server/server.py
+++ b/tableauserverclient/server/server.py
@@ -1,45 +1,7 @@
 from .exceptions import NotSignedInError
 from .endpoint import Sites, Views, Users, Groups, Workbooks, Datasources, Projects, Auth, Schedules, ServerInfo
-from . import RequestOptions
 
 import requests
-
-
-class Pager(object):
-    """ This class returns a generator that will iterate over all of the results.
-
-    server is the server object that will be used when calling the callback.  It will be passed
-    to the callback on each iteration
-
-    Callback is expected to take a server object and a request options and return two values, an array of results,
-    and the pagination item from the current call.  This will be used to build subsequent requests.
-    """
-
-    def __init__(self, fetcher, opts=None):
-        self._fetcher = fetcher.get
-        self._options = opts
-
-    def __call__(self):
-        current_item_list, last_pagination_item = self._fetcher(self._options)
-        count = 0
-
-        while count < last_pagination_item.total_available:
-            if len(current_item_list) == 0:
-                current_item_list, last_pagination_item = self._load_next_page(current_item_list, last_pagination_item)
-
-            yield current_item_list.pop(0)
-            count += 1
-
-    def __iter__(self):
-        return self()
-
-    def _load_next_page(self, current_item_list, last_pagination_item):
-        next_page = last_pagination_item.page_number + 1
-        opts = RequestOptions(pagenumber=next_page, pagesize=last_pagination_item.page_size)
-        if self._options is not None:
-            opts.sort, opts.filter = self._options.sort, self._options.filter
-        current_item_list, last_pagination_item = self._fetcher(opts)
-        return current_item_list, last_pagination_item
 
 
 class Server(object):
--- a/tableauserverclient/server/server.py
+++ b/tableauserverclient/server/server.py
@@ -1,6 +1,45 @@
 from .exceptions import NotSignedInError
 from .endpoint import Sites, Views, Users, Groups, Workbooks, Datasources, Projects, Auth, Schedules, ServerInfo
+from . import RequestOptions
+
 import requests
+
+
+class Pager(object):
+    """ This class returns a generator that will iterate over all of the results.
+
+    server is the server object that will be used when calling the callback.  It will be passed
+    to the callback on each iteration
+
+    Callback is expected to take a server object and a request options and return two values, an array of results,
+    and the pagination item from the current call.  This will be used to build subsequent requests.
+    """
+
+    def __init__(self, fetcher, opts=None):
+        self._fetcher = fetcher.get
+        self._options = opts
+
+    def __call__(self):
+        current_item_list, last_pagination_item = self._fetcher(self._options)
+        count = 0
+
+        while count < last_pagination_item.total_available:
+            if len(current_item_list) == 0:
+                current_item_list, last_pagination_item = self._load_next_page(current_item_list, last_pagination_item)
+
+            yield current_item_list.pop(0)
+            count += 1
+
+    def __iter__(self):
+        return self()
+
+    def _load_next_page(self, current_item_list, last_pagination_item):
+        next_page = last_pagination_item.page_number + 1
+        opts = RequestOptions(pagenumber=next_page, pagesize=last_pagination_item.page_size)
+        if self._options is not None:
+            opts.sort, opts.filter = self._options.sort, self._options.filter
+        current_item_list, last_pagination_item = self._fetcher(opts)
+        return current_item_list, last_pagination_item
 
 
 class Server(object):
--- a/test/test_pager.py
+++ b/test/test_pager.py
@@ -37,11 +37,10 @@ class PagerTests(unittest.TestCase):
             m.get(self.baseurl + "?pageNumber=3&pageSize=1", text=page_3)
 
             # No options should get all 3
-            workbooks = TSC.Pager(self.server.workbooks)
-            self.assertTrue(len(list(workbooks)) == 3)
+            workbooks = list(TSC.Pager(self.server.workbooks))
+            self.assertTrue(len(workbooks) == 3)
 
             # Let's check that workbook items aren't duplicates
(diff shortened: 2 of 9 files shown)


tableauserverclient/server/server.py
```diff
@@ -1,8 +1,47 @@
...
+
+    server is the server object that will be used when calling the callback.  It will be passed
+    to the callback on each iteration
+
+    Callback is expected to take a server object and a request options and return two values, an array of results,
+    and the pagination item from the current call.  This will be used to build subsequent requests.
+    """
+
+    def __init__(self, fetcher, opts=None):
+        self._fetcher = fetcher.get
```
RussTheAerialist: probably name the argument that is passed in `endpoint`

tableauserverclient/server/server.py
```diff
@@ -1,8 +1,47 @@
...
+        self._options = opts
+
+    def __call__(self):
```
RussTheAerialist: yeah this is not ideal.  It should probably just live in `__iter__` and we get rid of `__call__`

tableauserverclient/server/server.py
```diff
@@ -1,8 +1,47 @@
...
+        current_item_list, last_pagination_item = self._fetcher(self._options)
```
RussTheAerialist: this won't work if they set a page number in the options because count will never be greater than total available until we hit the end and throw an error.

We should check that a page number hasn't been set in the options, and if it has output a warning or throw an exception.

t8y8: There's always a page number (it's defaulted to 1).

Should the logic be:

1. On first iteration, run the generator
2. Store some state that says "we're working on it now"
3. Then just exhaust the generator starting from that point, then they could, if they want, start on page 3

tableauserverclient/server/server.py
```diff
@@ -1,8 +1,47 @@
...
 from .endpoint import Sites, Views, Users, Groups, Workbooks, Datasources, Projects, Auth, Schedules, ServerInfo
+from . import RequestOptions
+
 import requests
 
 
+class Pager(object):
+    """ This class returns a generator that will iterate over all of the results.
+
+    server is the server object that will be used when calling the callback.  It will be passed
```
LGraber: This comment does not appear to be relevant anymore

LGraber: why doesn't it work if we simply change how we initialize count?

tableauserverclient/server/server.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def __call__(self):
+        current_item_list, last_pagination_item = self._fetcher(self._options)
+        count = 0
+
+        while count < last_pagination_item.total_available:
+            if len(current_item_list) == 0:
+                current_item_list, last_pagination_item = self._load_next_page(current_item_list, last_pagination_item)
+
+            yield current_item_list.pop(0)
```
LGraber: This needs a bit more protection. Technically _load_next_page can return an empty list ... even after we fix our counting issue. This could happen if someone deletes an item while we are iterating. We should make sure the list is not empty before popping.

LGraber:
Oh yeah ... why did you put this in the server.py file and not its own file?

t8y8:
Alright! Addressed feedback (in person, and on here) tests added, and tested manually with this code:
```python


auth_stuff = TSC.TableauAuth(username='USER', password='password')

server = TSC.Server("http://server")

server.auth.sign_in(auth_stuff)

print(server.auth_token)

options = TSC.RequestOptions(pagenumber=2, pagesize=300)

wbs = TSC.Pager(server.workbooks, options)
print("GOT WBS")
print(len({i.id for i in wbs}))
server.auth.sign_out()
```

test/test_pager.py
```diff
@@ -0,0 +1,90 @@
...
+            m.get(self.baseurl + "?pageNumber=1&pageSize=1", text=page_1)
+            m.get(self.baseurl + "?pageNumber=2&pageSize=1", text=page_2)
+            m.get(self.baseurl + "?pageNumber=3&pageSize=1", text=page_3)
+
+            # No options should get all 3
+            workbooks = TSC.Pager(self.server.workbooks)
+            self.assertTrue(len(list(workbooks)) == 3)
+
+            # Let's check that workbook items aren't duplicates
+            workbooks = TSC.Pager(self.server.workbooks)
```
LGraber: why did you need to call this again?

test/test_pager.py
```diff
@@ -0,0 +1,90 @@
...
+        with open(GET_XML_PAGE3, 'rb') as f:
+            page_3 = f.read().decode('utf-8')
+        with requests_mock.mock() as m:
+            # Register Pager with default request options
+            m.get(self.baseurl, text=page_1)
+
+            # Register Pager with some pages
+            m.get(self.baseurl + "?pageNumber=1&pageSize=1", text=page_1)
+            m.get(self.baseurl + "?pageNumber=2&pageSize=1", text=page_2)
+            m.get(self.baseurl + "?pageNumber=3&pageSize=1", text=page_3)
```
LGraber: why are you registering these if we are not supposed to call them?

test/test_pager.py
```diff
@@ -0,0 +1,90 @@
...
+    def test_pager_with_options(self):
+        with open(GET_XML_PAGE1, 'rb') as f:
+            page_1 = f.read().decode('utf-8')
+        with open(GET_XML_PAGE2, 'rb') as f:
+            page_2 = f.read().decode('utf-8')
+        with open(GET_XML_PAGE3, 'rb') as f:
+            page_3 = f.read().decode('utf-8')
+        with requests_mock.mock() as m:
+            # Register Pager with default request options
+            m.get(self.baseurl, text=page_1)
```
LGraber: Why are you registering this if the test is not supposed to call it?

LGraber:
I am confused because the tests don't represent how we expect the users to use this class (ie you keep casting to list and doing len) and you don't have a sample of how to use it. Can you please add a sample

test/test_pager.py
RussTheAerialist: generators are one time use.  Once you iterate over them, you cannot go back.  He could have stored the output of `list(workbooks)` and then he wouldn't have had to call it twice.

test/test_pager.py
LGraber: Hahahaha. :) I know what it does. My comment was that we don't call these urls in this test so why do we need to register them. :)

test/test_pager.py
RussTheAerialist: though, it might be better (and more instructive) to use

```python
self.assertEquals(3, itertools.count(workbooks)
```

test/test_pager.py
LGraber: My bad ... it does get called (except maybe the first one). OOps

test/test_pager.py
```diff
@@ -37,11 +37,10 @@ def test_pager_with_no_options(self):
             m.get(self.baseurl + "?pageNumber=3&pageSize=1", text=page_3)
 
             # No options should get all 3
-            workbooks = TSC.Pager(self.server.workbooks)
-            self.assertTrue(len(list(workbooks)) == 3)
+            workbooks = list(TSC.Pager(self.server.workbooks))
+            self.assertTrue(len(workbooks) == 3)
```
RussTheAerialist: nitpick
```python
self.assertEquals(len(workbooks), 3)
```

PR #96, opened 2016-11-02
Deprecate site parameter on setter too

t8y8:
/cc @RussTheAerialist


t8y8:
Redoing this because git is acting funny

PR #100, opened 2016-11-07
Implement call to move to highest supported REST API version

t8y8:
Implemented during TC16 Hackathon!

Today, by default we sign in as '2.3' hard coded as our API version.

Now users can call `server.use_highest_version()` and it will negotiate the highest-supported API version for the current Server.

### DIFF
--- a/tableauserverclient/server/server.py
+++ b/tableauserverclient/server/server.py
@@ -1,7 +1,18 @@
+import xml.etree.ElementTree as ET
+
 from .exceptions import NotSignedInError
-from .endpoint import Sites, Views, Users, Groups, Workbooks, Datasources, Projects, Auth, Schedules, ServerInfo
+from .endpoint import Sites, Views, Users, Groups, Workbooks, Datasources, Projects, Auth, \
+    Schedules, ServerInfo, ServerInfoEndpointNotFoundError
 
 import requests
+
+_PRODUCT_TO_REST_VERSION = {
+    '10.0': '2.3',
+    '9.3': '2.2',
+    '9.2': '2.1',
+    '9.1': '2.0',
+    '9.0': '2.0'
+}
 
 
 class Server(object):
@@ -46,6 +57,29 @@ class Server(object):
         self._site_id = site_id
         self._user_id = user_id
         self._auth_token = auth_token
+
+    def _get_legacy_version(self):
+        response = self._session.get(self.server_address + "/auth?format=xml")
+        info_xml = ET.fromstring(response.content)
+        prod_version = info_xml.find('.//product_version').text
+        version = _PRODUCT_TO_REST_VERSION.get(prod_version, '2.1')  # 2.1
+        return version
+
+    def _determine_highest_version(self):
+        try:
+            old_version = self.version
+            self.version = "2.4"
+            version = self.server_info.get().rest_api_version
+        except ServerInfoEndpointNotFoundError:
+            version = self._get_legacy_version()
+
+        finally:
+            self.version = old_version
+
+        return version
+
+    def use_highest_version(self):
+        self.version = self._determine_highest_version()
 
     @property
     def baseurl(self):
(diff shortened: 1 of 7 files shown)


tableauserverclient/server/server.py
```diff
@@ -47,6 +58,29 @@ def _set_auth(self, site_id, user_id, auth_token):
         self._user_id = user_id
         self._auth_token = auth_token
 
+    def _get_legacy_version(self):
+        response = self._session.get(self.server_address + "/auth?format=xml")
+        info_xml = ET.fromstring(response.content)
+        prod_version = info_xml.find('.//product_version').text
+        version = _PRODUCT_TO_REST_VERSION.get(prod_version, '2.1')  # 2.1
```
t8y8: I'll make that comment not a random repetition: `2.1 is the version with permissions changes`

tableauserverclient/server/server.py
```diff
@@ -47,6 +58,29 @@ def _set_auth(self, site_id, user_id, auth_token):
...
+
+    def _determine_highest_version(self):
+        try:
+            old_version = self.version
+            self.version = "2.4"
+            version = self.server_info.get().rest_api_version
+        except ServerInfoEndpointNotFoundError:
+            version = self._get_legacy_version()
+
+        finally:
```
RussTheAerialist: This seems a bit weird.  We are setting the version, determining the new version, and then resetting the old version.  I think with the way this is written, we probably don't need _determine_highest_version to be it's own function.  In this case, we could do something like the following pseudo-code

```
save old version
set version to 2.4
Try to call get rest api version
If exception, try to get legacy version
If no legacy version found, restore old value.
```

I'm not sure which version would be clearer to understand.

t8y8: I currently have it this way so that this function simply *retrieves* the highest supported version, and the caller is responsible for setting it wherever it wants to be set -- this lets us reuse the function elsewhere (I don't know a use case yet, but I don't think the premature refactor makes it harder to follow).

I'd prefer the separation, if there's a strong objection let me know and I can combine it.

I'm pro-flag that does this automatically!!! I'll let Lee weigh in and can take a look

RussTheAerialist: I do not have a strong opinion either way, like I said, I wasn't sure which direction was clearer, and you make a good point about reuse.

divyavanmahajan: Prefer the API is set to the highest level (since that is usually what most people expect). If someone wants to fix the version - they can set the version after sign in.

tableauserverclient/server/server.py
LGraber: Personally, I am not sure I would include this code. The implication of adding this is that we actually test against some of these older versions, which we don't. While believe we should work with most of it, my inclination would be that if the user asks us to pick the latest version, and we can't find it (ie there is no serverInfo endpoint), we set it to 9.3. I am open to discuss, though. I am sure it is helpful, however we do need to be clear about what we actually test against.

tableauserverclient/server/server.py
LGraber: Why do you think if you put this in the use_highest_version() function it could not be used elsewhere? Just because it is a public function doesn't mean you can't call it from other parts of the code. Seems like we don't need two functions.

tableauserverclient/server/server.py
t8y8: We could just default to 9.3 -- this was recommended by @RussTheAerialist since I think he plans to decorate versions back farther than 9.3

t8y8:
@LGraber I don't follow your comments on the Server constructor -- I didn't change the behavior of `__init__`. We can chat more when I'm back to I can figure out what oyu mean (but remember I'm in 🗾 and :taiwan: for a while!)

LGraber:
I was commenting on the question of whether we should add a flag to the constructor to do this versus having them always call the method. It is not actually in your code.

t8y8:
I just rebased on top of the latest changes -- I didn't make any code changes to the PR.

I think I saw these issues still open:
1. Default to API v2.2 instead of v2.1 per @LGraber 
2. Why don't I put it all in `use_highest_version` <- I much prefer the code this way because it keeps the logic around version-detecting clean and keeps the try/except blocks out of the publish function. I can move this if there's strong objection but I really don't want to
3. Should we have a flag somewhere to do this automatically? @LGraber and @RussTheAerialist both commented, no decisions.

Issue #102, opened 2016-11-16
Convert dates to py datetime


reevery:
Is there any reason why this module doesn't convert dates to python datetime objects? Things like last logged in / last updated etc.

RussTheAerialist:
I think this falls into the bucket of "things we haven't done yet".  Definitely something we want to implement sooner rather than later.

RussTheAerialist:
This morning, I checked in support for this into development, which will be included in the next release (yay!) and thanks for pointing this out!

PR #103, opened 2016-11-16
Initial implementation to address #102 and provide datetime objects

RussTheAerialist:
I'm a little conflicted on this implementation.  created_at and updated_at are not user settable, but I went down the "use it as a property" route for coercing to the datetime object, so I wanted to get other opinions on it.  The other option would be to change the _set_values functions to call the parse_datetime function instead of using the property.

Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,4 @@ script:
   # Tests
   - python setup.py test
   # pep8 - disabled for now until we can scrub the files to make sure we pass before turning it on
-  - pycodestyle .
+  - pycodestyle tableauserverclient test
--- a/tableauserverclient/models/property_decorators.py
+++ b/tableauserverclient/models/property_decorators.py
@@ -1,5 +1,12 @@
+import datetime
 import re
 from functools import wraps
+from ..datetime_helpers import parse_datetime
+try:
+    basestring
+except NameError:
+    # In case we are in python 3 the string check is different
+    basestring = str
 
 
 def property_is_enum(enum_type):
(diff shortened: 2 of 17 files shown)


LGraber:
I don't like exposing setters that users should not call. That doesn't feel good right now. Maybe one day we will have a reason to expose this but we don't right now. Our public api set should be clear. Also not sure how it is that you didn't have to change the parse code to use the setter. Did I miss something there? It feels like we should either use parse_datetime  function when building the object from the response or we should have an "internal setter".

RussTheAerialist:
@LGraber that was pretty much the same conclusion I was coming to, but I figured I'd throw out what I had to make sure others felt the same way.  I only have unit tests written (this wasn't intended to be a complete PR) so the parsing code doesn't have tests yet, it was just where I got by the time I had to leave for meetings.

I'll continue this tomorrow morning to finish it out.  My approach now will be to call parse_datetime in the appropriate parsing functions where we parse out created_at and updated_at

tableauserverclient/models/property_decorators.py
```diff
@@ -99,3 +106,25 @@ def validate_regex_decorator(self, value):
             return func(self, value)
         return validate_regex_decorator
     return wrapper
+
+
+def property_is_datetime(func):
```
RussTheAerialist: Even though this isn't being used right now, I left it in.  If you feel like I should delete it, I can.

.travis.yml
```diff
@@ -14,4 +14,4 @@ script:
   # Tests
   - python setup.py test
   # pep8 - disabled for now until we can scrub the files to make sure we pass before turning it on
-  - pycodestyle .
+  - pycodestyle tableauserverclient test
```
LGraber: Why are you doing this?

tableauserverclient/models/property_decorators.py
LGraber: I am happy to leave this

.travis.yml
RussTheAerialist: A recent update to pycodestyle (since we always install the latest in the travis run) is picking up files that are outside of our control (libraries installed), so now we want to limit to specifically code we have control over.

PR #110, opened 2016-12-08
PROTOTYPE: Running smoketests

t8y8:
Here's a quick prototype for one way we could add a smoketest command (to run locally, or even on some subset of travisci runs someday).

For simplicity's stake I have the command living in `setup.py` and it just calls a `smoke.py` script vis subprocessing -- we could make modules and things and have it be a little integrated, but I wanted something quick to explore the idea.

This does run locally against "http://qa-server"!

@RussTheAerialist @LGraber what do you think of the idea?


LGraber:
I have to read up a bit on this but I have been wanting to do this. We need to port our smoketests to this. This is great!

t8y8:
Just so I don't forget, apparently another way people add custom commands to `setup.py` is to be lazy and do this:
```python
if sys.argv[-1] == "command":
    do_stuff()
```

Because as you can see, subclassing the command is kind of annoying

setup.py
```diff
@@ -1,8 +1,43 @@
...
+
+        import getpass
+        self.password = getpass.getpass()
+
+    def run(self):
+        import subprocess as sp
+        import os
+
+        with open(os.devnull, 'wb') as silent:
+            sp.check_call(['python', 'smoke.py', self.server, self.username, self.password, self.site], stdout=silent)
```
RussTheAerialist: Rather than do it this way, could we do a directory of smoke tests and then run it with `python -m unittest smoke`?  This would mean we wouldn't need to write our own test runner just to run the smoke tests, then we can have a Base Smoke Test class that reads the arguments from the command line, and all tests derive from that test class.

smoke.py
```diff
@@ -0,0 +1,24 @@
+import sys
+import tableauserverclient as TSC
+
+server, username, password, site = sys.argv[1:]
```
RussTheAerialist: also, ew.

```python
def main():
   # ...

if __name__ == '__main__':
    main()
```

setup.py
t8y8: Yes, that is possible, but I wasn't sure from online docs how to:

1. Have a SmokeTestBase that gets the command line args
2. Doesn't run under `python setup.py test` -- by default it runs everything named `*test*.py`

t8y8:
Retiring this PR as it doesn't seem to want to run travisci

Issue #105, opened 2016-11-20
Memory Error downloading large extract data sources


TCMurph:
Hi,

I am using the module to download data sources, and publish them to another server instance. Most data sources download fine, but larger extracts give the error below. Have you seen this before?
I can provide the code if you need it.
There is no memory spike on either the Tableau Server, or the machine I am running the code.

Thanks.

Traceback (most recent call last):
  File "tableauserver.py", line 44, in <module>
    prod.datasources.download(datasource.id)
  File "C:\Python27\lib\site-packages\tableauserverclient\server\endpoint\dataso
urces_endpoint.py", line 67, in download
    server_response = self.get_request(url)
  File "C:\Python27\lib\site-packages\tableauserverclient\server\endpoint\endpoi
nt.py", line 54, in get_request
    request_object=request_object)
  File "C:\Python27\lib\site-packages\tableauserverclient\server\endpoint\endpoi
nt.py", line 34, in _make_request
    server_response = method(url, **parameters)
  File "C:\Python27\lib\site-packages\requests\sessions.py", line 488, in get
    return self.request('GET', url, **kwargs)
  File "C:\Python27\lib\site-packages\requests\sessions.py", line 475, in reques
t
    resp = self.send(prep, **send_kwargs)
  File "C:\Python27\lib\site-packages\requests\sessions.py", line 628, in send
    r.content
  File "C:\Python27\lib\site-packages\requests\models.py", line 755, in content
    self._content = bytes().join(self.iter_content(CONTENT_CHUNK_SIZE)) or bytes
()
MemoryError

RussTheAerialist:
Based on the error, the python process is running out of memory because we are trying to hold the extract in memory.

A quick fix for you to try to check to make sure you are running the 64-bit version of python.  The 32-bit version has necessary upper memory limits because of the address space.  64-bit has significantly more address space.

The long term fix will be that we need to provide a way to stream the extract to disk rather than try to hold it all in memory.

TCMurph:
Thanks for such a quick reply Russell.
I was running python 32-bit version (just an oversight on my part), so I downloaded the 64-bit version and tested it out on a 10gb extract, and it ran with no issues!! Thank you very much.

If you feel like the stream to disk is something that would be a good feature enhancement, I look forward to testing that out in the future.

Paul

RussTheAerialist:
[...]

PR #111, opened 2016-12-12
Fix large downloads (#105)

t8y8:
I've got a working solution for #105 

Requests provides a way to just get the response headers back and then iterating over the content in chunks. This lets us stream to disk instead of trying to hold large workbooks in memory.

(See http://docs.python-requests.org/en/master/api/#requests.Response.iter_content and http://docs.python-requests.org/en/master/user/advanced/#body-content-workflow)

As a proof of concept I extended our wrapper methods to allow per-call http options (because we don't need to stream every request) and then tested it on workbook download with the script below.

If this approach works, I can clean it up and add it to data source download as well.

But a few questions:
1. Are we ok with the way I extended our request-get-wrapper function? An alternative is to have a special `get_download_request` instead
2. I need to make sure the response always gets closed once we use `stream=True` I think I'm right but may want to just wrap it all in a context manager or try/except
3. Preferences on a chunk size? (I did 4MB because it's usually safe)

```python

authz = TSC.TableauAuth('user', 'pass')
server = TSC.Server("http://server")

server.auth.sign_in(authz)
print("Signed in: " + server.auth_token)

#workbooks = TSC.Pager(server.workbooks)

big_wb = "FIND ONE OR USE CODE TO FIND ONE"

print("Downloading workbook: " + big_wb)

server.workbooks.download(big_wb)

print("DONE")
```

### DIFF
--- a/tableauserverclient/server/endpoint/datasources_endpoint.py
+++ b/tableauserverclient/server/endpoint/datasources_endpoint.py
@@ -6,6 +6,7 @@ import os
 import logging
 import copy
 import cgi
+from contextlib import closing
 
 # The maximum size of a file that can be published in a single request is 64MB
 FILESIZE_LIMIT = 1024 * 1024 * 64   # 64MB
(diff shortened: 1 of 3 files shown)


t8y8:
Added datasources and wrapped it all in `closing()` so we can be sure the connection gets closed.

I went ahead and tested on large and small items on an internal server, and realized that the chunk size didn't actually seem to affect the download time that much -- so I went with 1KB, what most examples on the internet do.

LGraber:
This is really clean. I like it very much. :rocket:

PR #113, opened 2016-12-14
Issue 112 - Support for the new image api endpoint

jz-huang:
added support for the new image api endpoint. Added unit tests and a sample. Added the new sample to docs. 

Unit tests and pep8 run clean. Manually tested sample.


jz-huang:
Added a new ImageRequestOptions class because the "resolution" parameter doesn't really fit in with the current parameters in RequestOptions. Filter, sort, paging are all for endpoints that return a list of resources, doesn't really make sense to to add "resolution" to the group. And resolution is only parameter that's available for the image endpoint. With the new ImageRequestOptions, it would be easier to add more parameters for this endpoint in the future (i.e. filters within the viz).

t8y8:
Do we really need ABC and metaclass stuff? I think just a simple inherit is probably enough, it's a single method and not a complex class. It can also all probably rest in the same file -- splitting everything out into multiple files and enforcing the subclassing via metaclasses is very "java"

The rest of the code looks pretty solid, I'll do a detailed review tomorrow!

samples/get_view_image.py
```diff
@@ -0,0 +1,67 @@
...
+import getpass
+import logging
+
+import tableauserverclient as TSC
+
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Query View Image From Server')
+    parser.add_argument('--server', '-s', required=True, help='server address')
```
t8y8: Please add site as well (I know not all samples have site yet, but let's not build more debt)

samples/get_view_image.py
```diff
@@ -0,0 +1,67 @@
...
+    # Step 1: Sign in to server.
+    tableau_auth = TSC.TableauAuth(args.username, password)
+    server = TSC.Server(args.server)
+    server.version = 2.5
+
+    with server.auth.sign_in(tableau_auth):
+
+        # Step 2: Get all the projects on server, then look for the default one.
+        all_projects, pagination_item = server.projects.get()
+        default_project = next((project for project in all_projects if project.is_default()), None)
```
t8y8: You can just use use TSC.Pager here.

samples/get_view_image.py
```diff
@@ -0,0 +1,67 @@
...
+            req_option = TSC.RequestOptions()
+            req_option.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
+                                             TSC.RequestOptions.Operator.Equals, args.view_name))
+            all_views, pagination_item = server.views.get(req_option)
+            view_item = all_views[0]
+            image_req_option = TSC.ImageRequestOptions(imageresolution=TSC.ImageRequestOptions.Resolution.High)
+            server.views.populate_image(view_item, image_req_option)
+
+            with open(args.filepath, "wb") as image_file:
+                image_file.write(view_item.image)
```
t8y8: How large can this file get?

tableauserverclient/server/request_options_base.py
```diff
@@ -0,0 +1,9 @@
+import abc
+
+
+class RequestOptionsBase(object):
```
t8y8: I think the abc and metaclass magic is unnecessary and doesn't really buy us anything but complexity to new contributors. It's kinda black arts level stuff :)

samples/get_view_image.py
benlower: @t8y8 i think you are saying "don't assume that everything is in the default site." yes?

samples/get_view_image.py
```diff
@@ -0,0 +1,67 @@
+####
+# This script demonstrates how to use the Tableau Server Client
+# to query a high resolution image of a view from Tableau Server.
```
benlower: We're not really 'querying' an image are we? More like 'fetch' or 'download'?

samples/get_view_image.py
```diff
@@ -0,0 +1,67 @@
...
+
+import tableauserverclient as TSC
+
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Query View Image From Server')
+    parser.add_argument('--server', '-s', required=True, help='server address')
+    parser.add_argument('--username', '-u', required=True, help='username to sign into server')
+    parser.add_argument('--view-name', '-v', required=True, help='name of view')
```
benlower: can we expand the help text just a bit please? something like "name of the view we want an image of"

samples/get_view_image.py
```diff
@@ -0,0 +1,67 @@
...
+    password = getpass.getpass("Password: ")
+
+    # Set logging level based on user input, or error by default
+    logging_level = getattr(logging, args.logging_level.upper())
+    logging.basicConfig(level=logging_level)
+
+    # Step 1: Sign in to server.
+    tableau_auth = TSC.TableauAuth(args.username, password)
+    server = TSC.Server(args.server)
+    server.version = 2.5
```
benlower: @t8y8 @LGraber is the inclusion of 'server.version = x' the way we are indicating a min version for the sample? will the new auto-detect version help here or not? if we are going to keep server.version then let's add a comment that says this is the version in which this was introduced.

samples/get_view_image.py
```diff
@@ -0,0 +1,67 @@
...
+
+    with server.auth.sign_in(tableau_auth):
+
+        # Step 2: Get all the projects on server, then look for the default one.
+        all_projects, pagination_item = server.projects.get()
+        default_project = next((project for project in all_projects if project.is_default()), None)
+
+        # Step 3: If default project is found, download the image for the specified view name
+        if default_project is not None:
```
benlower: should we add optional ability to specify the project in the args or is that over-complicating the sample?

samples/get_view_image.py
```diff
@@ -0,0 +1,67 @@
...
+            req_option = TSC.RequestOptions()
+            req_option.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
+                                             TSC.RequestOptions.Operator.Equals, args.view_name))
+            all_views, pagination_item = server.views.get(req_option)
```
benlower: @t8y8 use pager here too? is our guidance to always use pager or only use it when you expect to have a lot of items (in this case we'd expect filter to reduce item count)?

samples/get_view_image.py
```diff
@@ -0,0 +1,67 @@
...
+            view_item = all_views[0]
+            image_req_option = TSC.ImageRequestOptions(imageresolution=TSC.ImageRequestOptions.Resolution.High)
+            server.views.populate_image(view_item, image_req_option)
+
+            with open(args.filepath, "wb") as image_file:
```
benlower: what does "wb" do here? confusing to me b/c I assume it means workbook but this is an image.

samples/get_view_image.py
jz-huang: sounds good, i'll add content_url as a parameter.

tableauserverclient/server/image_request_options.py
```diff
@@ -0,0 +1,20 @@
+from .request_options_base import RequestOptionsBase
+
+
+class ImageRequestOptions(RequestOptionsBase):
+    class Resolution:
+        High = 'high'
```
benlower: is there only one option?

samples/get_view_image.py
jz-huang: The files are usually a few hundred KB, probably gets to a couple of MB's at most.

samples/get_view_image.py
t8y8: Thanks! 

(Don't worry about the other samples for now, I might do that in a quick PR against development; and we don't want merge conflicts :) )

samples/get_view_image.py
jz-huang: ok

samples/get_view_image.py
t8y8: Cool, we shouldn't have to worry about it then.

test/test_view.py
```diff
@@ -63,3 +63,29 @@ def test_populate_preview_image_missing_id(self):
         single_view._id = None
         single_view._workbook_id = '3cc6cd06-89ce-4fdc-b935-5294135d6d42'
         self.assertRaises(TSC.MissingRequiredFieldError, self.server.views.populate_preview_image, single_view)
+
+    def test_populate_image(self):
+        with open(POPULATE_PREVIEW_IMAGE, 'rb') as f:
+            response = f.read()
+        with requests_mock.mock() as m:
+            m.get(self.baseurl + '/views/d79634e1-6063-4ec9-95ff-50acbf609ff5/image', content=response)
```
benlower: what server will this execute against? is this a valid view ID?

t8y8: It's a mock, he's registering that URL so that when the API calls it it'll hit this fake endpoint and return the image

tableauserverclient/server/request_options_base.py
jz-huang: Ok, I wanted to take advantage of the builtin abstract class support. But it's kind of an overkill for this simple use case and like you said, probably makes the code more difficult to understand as well. will switch back to simple inheritance =)

samples/get_view_image.py
t8y8: This will just fail if run against a new Server (actually, it will fail on sign in).

My PR for detecting the version is outstanding, #100 -- waiting for approval to check it in.

But even then it wouldn't know which calls each version can and cannot make.

@RussTheAerialist had an idea to decorate the calls in the API with what version they use so we could throw a nice "NotSupported" exception or something, that work isn't yet started that I'm aware of (and is sorta dependent on my PR to get the version from Server)

samples/get_view_image.py
jz-huang: Good catch, download is more appropriate, especially since we are saving the image to a specified filepath

samples/get_view_image.py
t8y8: "write binary" it's the standard in python code for opening a file in binary mode for writing

samples/get_view_image.py
t8y8: Ha, good point.

But since he's not searching for one in particular, I don't think it matters much -- he just wants to get a view

samples/get_view_image.py
t8y8: Good question.

All sites will have a default project, but exposing the ability to choose another might be nice (I think all the samples would want that though)

tableauserverclient/server/request_options_base.py
t8y8: Thank you!

Yeah, in this case it's not warranted (It's usually only done for large classes like a protocol in asyncio or twisted or something)

SIDE NOTE: I'm not a fan of the idea of having subclasses for all of our different url options/parameters, I feel like it'll just grow into a long list of "Which class do I need again", but in absence of a better idea I'll be quiet for now :)

samples/get_view_image.py
jz-huang: actually the views endpoint returns all the views on the site, so we don't need to specify the project. Copy and pasting problems... I'll remove the project related logic

tableauserverclient/server/image_request_options.py
jz-huang: yes, it defaults to a standard resolution (~800x800), high doubles that.

tableauserverclient/server/request_options_base.py
jz-huang: I'm not a big fan of this either, but I didn't like adding "resolution" to RequestOptions. We don't have a way of dealing with this in the current design. I think the intention might be that once we add "fields" to RequestOptions that would cover all of the URL parameters. This endpoint is kind of an edge case. @LGraber probably have an opinion or two on this.

tableauserverclient/server/image_request_options.py
benlower: ok. can you please add a comment here that says if you don't specify 'high' that the underlying REST API will default to standard?

tableauserverclient/server/request_options.py
```diff
@@ -46,3 +51,23 @@ def apply_query_params(self, url):
...
+
+class ImageRequestOptions(RequestOptionsBase):
+    # if 'high' isn't specified, the REST API endpoint returns an image with standard resolution
+    class Resolution:
+        High = 'high'
+
+    def __init__(self, imageresolution=None):
+        self.imageresolution = imageresolution
+
+    def image_resolution(self, imageresolution):
```
t8y8: I called :)

This is unnecessary (and technically slower)

samples/download_view_image.py
```diff
@@ -0,0 +1,68 @@
...
+        req_option = TSC.RequestOptions()
+        req_option.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
+                                         TSC.RequestOptions.Operator.Equals, args.view_name))
+        all_views, pagination_item = server.views.get(req_option)
+        if not all_views:
+            raise LookupError("View with the specified name was not found.")
+        view_item = all_views[0]
+
+        # Step 3: Query the image endpoint and save the image to the specified location
+        image_req_option = TSC.ImageRequestOptions(imageresolution=TSC.ImageRequestOptions.Resolution.High)
```
LGraber: I agree with Tyler. This is starting to feel heavy to code. The base RequestOptions felt more necessary because fields was more than a simple hash. As I watch this code being written in our samples, though, it is feeling heavier and heavier. It feels like the simple hash syntax might be more lightweight and we can still validate it. I am okay with you checking this in since it is in vnext but I think we should discuss this as I want users to be able to write minimal code that is easy to read and it is starting to feel like perhaps our use of enums and strongly typed objects is making the python harder and less ... pythonic. I am open to discussion.

LGraber: I meant to say filter, not fields. Fields, also, was going to be interesting.

t8y8: Thanks @LGraber , I opened #114 to track this idea -- I've been googling how some other libraries do it for inspiration (nothing struck me yet), but that'll make it so we don't lose the convo in this PR

jz-huang:
I'm merging this in for now so Ben can WAM, let's keep the discussion going in #114.

PR #115, opened 2016-12-20
Add support for the oAuth flag

hugoboos:


t8y8:
@hugoboos thank you for the submission!

We'll need a few things before we can review:
1) Please rebase this against the development branch
2) Please need to fill out the CLA (linked here http://tableau.github.io/)
3) We don't currently have any tests that check that the serialization works... Can you include a screenshot of some output that includes the new element? (It's just nice to have a sanity check on record!)

hugoboos:
Hi,

1. Rebased and changed the PR
2. Requested e-form to sign
3. Is it sufficient to show some code that publishes a datasource using the oAuth flag?

t8y8:
Thank you @hugoboos !

I checked and the CLA stuff is all in motion. Once it's in I'll officially review the code.

For the 3rd item, I think just making a WorkbookItem and calling the method `WorkbookRequest._generate_xml(wb_item)` and printing the output should show that the serialization works well enough for me (That's how we verified the first incarnation of ConnectionCredentials)

tableauserverclient/models/connection_credentials.py
```diff
@@ -9,10 +9,11 @@ class ConnectionCredentials(object):
 
     """
 
-    def __init__(self, name, password, embed=True):
+    def __init__(self, name, password, embed=True, oauth=True):
```
t8y8: I think this should default to 'false' since most connections don't need an oAuth parameter (and I'm not sure what they'll do if both are present)

tableauserverclient/server/request_factory.py
```diff
@@ -42,6 +42,7 @@ def _generate_xml(self, datasource_item, connection_credentials=None):
             credentials_element.attrib['name'] = connection_credentials.name
             credentials_element.attrib['password'] = connection_credentials.password
             credentials_element.attrib['embed'] = 'true' if connection_credentials.embed else 'false'
+            credentials_element.attrib['oAuth'] = 'true' if connection_credentials.oauth else 'false'
```
t8y8: Even with my above comment, this would be including the oauth parameter in all requests, even if it's not an oauth enabled connection... Can you test and see if that causes a problem? If it does I suggest defaulting to `None` and then doing something like:
```python
if connection_credentials.oauth is not None:
    redentials_element.attrib['oAuth'] = 'true'
```

hugoboos: Will check for that.

hugoboos: I have looked up the [REST API documentation](http://onlinehelp.tableau.com/v10.0/api/rest_api/en-us/help.htm#REST/rest_api_ref.htm#Publish_Datasource%3FTocPath%3DAPI%2520Reference%7C_____43) and only when the connection is oAuth, you should set it to true.

So adding `oAuth="false"` doesn't make any sense. I will update the PR and make the default value set to `False`.

hugoboos:
I have looked up the [REST API documentation](http://onlinehelp.tableau.com/v10.0/api/rest_api/en-us/help.htm#REST/rest_api_ref.htm#Publish_Datasource%3FTocPath%3DAPI%2520Reference%7C_____43) and only when the connection is oAuth, you should set it to true.

So adding `oAuth="false"` doesn't make any sense. I will update the PR and make the default value set to `False`.

Screen shot of the generated XML:
screen shot 2016-12-23 at 13 42 16

LGraber:
🚀

Issue #55, opened 2016-10-07
Add decorators to specify when a REST endpoint was added


RussTheAerialist:
Needed to be able to show users meaningful errors if they are using a version of the REST api that does not support a certain feature.  This is needed to fully address #54

t8y8:
Currently there is only one serializer for each request type -- since the schemas aren't always compatible we'll need to deal with that somehow.

RussTheAerialist:
I don't follow what you mean here?  As long as the major version number hasn't changed, the schemas should be compatible because we guarantee no breaking changes unless the major version is revved.  I would imagine we'll branch and no longer add features to the 2.x version once 3.x comes in.

t8y8:
We've added attributes over time, so 2.3 may have an attribute that 2.2 does not have. If the serializer doesn't know about those differences it'll get stripped out either on read or write.

The endpoint itself is backwards compatible because we ignore attributes we don't understand and treat them all as optional.

RussTheAerialist:
i still don't see where the problem is? If we ignore the extra attributes,
then it doesn't matter as long as the serializer includes all of them, the
rest endpoint should handle it properly.

If we changed the meaning of attributes, then that's a breaking change and
we should have revved the major version.

t8y8:
Yeah, I guess I'm saying I don't know -- you're probably right.

I'm not aware of any places we changed the meaning of an attribute, just their existence.

RussTheAerialist:
yeah, I think there will be some exploration that happens when we implement this enhancement.  If there are attributes we need to not serialize, we could have another decorator that says "remove this attribute if less than version x" or we could have separate serializers for each version where we need them.  It's something to figure out, but I _think_ it should all work.

t8y8:
I hacked up a quick decorator and tested that this does in fact work:

```python
def min_api(min_ver):
    def outter_api_decorator(func):
        @wraps(func)
        def wrapper(self, *args):
            server_version = self.parent_srv.version
            if float(server_version) < min_ver:
                error = "This endpoint is not support in version {}. Requires {}".format(server_version, min_ver)
                raise ValueError(error)
            return func(self, *args)
        return wrapper
    return outter_api_decorator
```

It should probably throw a specific exception `EndpointNotSupported` or something but this was copy/paste from an example to start.

1. Where to put it? What exception to use? Will this work? Etc.
2. I think it really needs the 'auto-detect' part of the version detection function to be the most useful

/cc @RussTheAerialist

[...]

PR #124, opened 2017-01-06
Add annotation for endpoints to indicate minimum supported API version

t8y8:
Addresses #55 

This is just the decorator, I manually tested with the `users.get` endpoint and a script.

Once we add the annotation to the actual endpoints this will get automagically covered by all existing tests. I can make a unittest that mocks everything to test this method, but that doesn't feel super useful (It'd literally be testing does `<` work).

I think I'd like to break it into two PRs, add this method first, and in the second PR add the annotations and and ensure existing tests still pass

Open to feedback on the exception name and where I put this method -- but endpoint felt like the most natural place for it

/cc @RussTheAerialist @LGraber

### DIFF
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -91,7 +91,7 @@ def api(version):
         None
 
     Example:
-    >>> @api(version=2.3)
+    >>> @api(version="2.3")
     >>> def get(self, req_options=None):
     >>>     ...
     '''
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -1,6 +1,12 @@
-from .exceptions import ServerResponseError
+from .exceptions import ServerResponseError, EndpointUnavailableError
+from functools import wraps
+
 import logging
 
+try:
+    from distutils2.version import NormalizedVersion as Version
+except ImportError:
+    from distutils.version import LooseVersion as Version
 
 logger = logging.getLogger('tableau.endpoint')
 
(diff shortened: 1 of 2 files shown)


tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -69,3 +70,33 @@ def post_request(self, url, xml_request, content_type='text/xml'):
                                   content=xml_request,
                                   auth_token=self.parent_srv.auth_token,
                                   content_type=content_type)
+
+
+def api(*, version):
+    '''Annotate the minimum supported version for an endpoint.
+
+    Checks the version on the server object and does a float comparison.
+    It will raise an exception if the server vesion is > the version specified.
```
t8y8: 'vesion' -> 'version' typo

tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -69,3 +70,33 @@ def post_request(self, url, xml_request, content_type='text/xml'):
...
+    Example:
+    >>> @api(version=2.3)
+    >>> def get(self, req_options=None):
+    >>>     ...
+    '''
+    def _decorator(func):
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            server_version = self.parent_srv.version
+            if float(server_version) < version:
```
RussTheAerialist: Please use https://docs.python.org/3.6/distutils/apiref.html#module-distutils.version to compare version numbers.  Casting to float will break if/when we decide to be more nuanced with your version numbers.

tableauserverclient/server/endpoint/users_endpoint.py
```diff
@@ -1,4 +1,4 @@
-from .endpoint import Endpoint
+from .endpoint import Endpoint, api
```
RussTheAerialist: Do you actually use api anywhere?

t8y8: oops, this was left over from my manual test script

PR #129, opened 2017-01-12
118 suppress insecurerequestwarning doc

Talvalin:
Okay, so I might have changed more than required here, so bear with me and feel free to reject with prejudice.

- Using `with` statements is idiomatic Python, so really it should be the first example that we provide rather than the second
- I added a section header to break up the page when introducing the topic of disabling certificate verification and warnings
- The last section title "A better way to avoid certificate warnings" is not amazing, so alternative suggestions are welcome

### DIFF
--- a/docs/docs/sign-in-out.md
+++ b/docs/docs/sign-in-out.md
@@ -53,12 +53,17 @@ C:\Program Files (x86)\Python35-32\lib\site-packages\requests\packages\urllib3\c
 InsecureRequestWarning)
 ```
 
-These warnings can be disabled with the following line:
+These warnings can be disabled by adding the following lines to your script:
 ```py
-requests.packages.urllib3.disable_warnings(InsecureRequestWa‌​rning)
+import requests
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
+requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 ```
 
 ### A better way to avoid certificate warnings
-Instead of disabling warnings or certificate verification when using self-signed certificates, we recommend using a Certificate Authority (CA) signed certificate.
+Instead of disabling warnings or certificate verification to avoid issues with self-signed certificates, the best practice is to use a certificate signed by a Certificate Authority.
 
-In addition to standard Certificate Authorities, [Let's Encrypt](https://letsencrypt.org/) is an automated and open Certificate Authority that can be used for absolutely free. Please note that no official Windows client currently exists, and so certificates must be signed using a Linux machine or using a third-party Windows client.
+If you have the ability to do so, we recommend the following Certificate Authorities:
+* [GlobalSign](https://www.globalsign.com/en/)
+* [Let's Encrypt](https://letsencrypt.org/) - a free, automated, and open Certificate Authority
+* [SSL.com](https://www.ssl.com/)
--- a/docs/docs/sign-in-out.md
+++ b/docs/docs/sign-in-out.md
@@ -2,8 +2,21 @@
 title: Sign In and Out
 layout: docs
 ---
+## Signing in and out
+To sign in and out of Tableau Server, simply call `Auth.sign_in` as follows:
+```py
+import tableauserverclient as TSC
 
-To sign in and out of Tableau Server, call the `Auth.sign_in` and `Auth.sign_out` functions like so:
+tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD')
+server = TSC.Server('http://SERVER_URL')
+
+with server.auth.sign_in(tableau_auth):
+    # Do awesome things here!   
+
+# The Auth context manager will automatically call 'Auth.sign_out' on exiting the with block
+```
+
+Alternatively, call both `Auth.sign_in` and `Auth.sign_out` functions explicitly like so:
 
 ```py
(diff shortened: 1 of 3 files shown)


docs/docs/sign-in-out.md
```diff
@@ -2,8 +2,21 @@
...
+```py
+import tableauserverclient as TSC
+
+tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD')
+server = TSC.Server('http://SERVER_URL')
+
+with server.auth.sign_in(tableau_auth):
+    # Do awesome things here!   
+
+# The Auth context manager will automatically call 'Auth.sign_out' on exiting the with block
```
t8y8: @RussTheAerialist may disagree with me here, but I don't think we should lead with the context manager.
I would lead by teaching `sign_in` and `sign_out` and then mentioning the context manger for simple scripts.

`with` blocks are intended for short lived transactions, RAII-style 'get-and-then-close', or similarly simple operations. Not long chains of REST API calls.

As evidence on this front, I've written a dozen TSC scripts for testing various other features my teams are interacting with, or configuring a server, etc, and I've never used the context manager because it baiscally automatically forces me out of pep8 for all but trivial code ... but that's my soapbox :)

docs/docs/sign-in-out.md
```diff
@@ -23,17 +36,29 @@ server.auth.sign_out()
     limited by the maximum session length (of four hours) on Tableau Server.
 </div>
 
-
-Alternatively, for short programs, consider using a `with` block:
+### Disabling certificate verification and warnings
+If a self-signed certificate is used, certificate verification can be disabled with ```add_http_options```:
```
t8y8: Self-signed isn't what's bad, necessarily, it's that they aren't trusted by the client.

The best way to phrase this might be something like:
```text
Certain environments may throw errors related to ssl configuration, such as self-signed certificates, expired certificates, etc...```

File: docs/docs/sign-in-out.md [hunk 3]
```diff
@@ -23,17 +36,29 @@ server.auth.sign_out()
...
 tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD')
 server = TSC.Server('http://SERVER_URL')
+server.add_http_options({'verify': False})
+```
 
-with server.auth.sign_in(tableau_auth):
-    # Do awesome things here!
+However, each subsequent REST API call will print an ```InsecureRequestWarning```:
+```
+C:\Program Files (x86)\Python35-32\lib\site-packages\requests\packages\urllib3\connectionpool.py:838: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/security.html
```
t8y8: Don't include the file path, it's just noise for the docs and our readers

docs/docs/sign-in-out.md
```diff
@@ -23,17 +36,29 @@ server.auth.sign_out()
...
+InsecureRequestWarning)
 ```
 
-The TSC library signs you out of Tableau Server when you exit out of the `with` block.
+These warnings can be disabled with the following line:
+```py
+requests.packages.urllib3.disable_warnings(InsecureRequestWa‌​rning)
```
t8y8: I think you have to do
```python
import requests
from requests.packages.urllib3.exceptions import InsecureRequestWarning
requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
```

Can you test?

Talvalin:
Updated as per discussion in issue 118, and also updated the workaround to add all of the lines required to disable `InsecureRequestWarning` messages

docs/docs/sign-in-out.md
```diff
@@ -23,17 +36,29 @@ server.auth.sign_out()
...
+```
+
+### A better way to avoid certificate warnings
+Instead of disabling warnings or certificate verification when using self-signed certificates, we recommend using a Certificate Authority (CA) signed certificate.
```
t8y8: See my note about the causes -- it could be signed by an internal CA but your client is misconfigured, etc (I think) but otherwise this section looks pretty solid.

Talvalin:
Updated following @t8y8's comments. :)

docs/docs/sign-in-out.md
Talvalin: This is indeed required, and I verified it earlier on.

jdomingu:
🚀 Nice work!

docs/docs/sign-in-out.md
Talvalin: Removed

docs/docs/sign-in-out.md
Talvalin: Rewritten in line with the above

docs/docs/sign-in-out.md
Talvalin: I reverted the change, but did shift the note below the `with` block example, since it looked weird sitting there in between

t8y8:
:rocket:

docs/docs/sign-in-out.md
Talvalin: This was also reworded in line with your previous comment about causes

RussTheAerialist:
Great work! Thanks for the contribution

Issue #101, opened 2016-11-08
Test Coverage for unit tests


RussTheAerialist:
Not too bad all things considered, but there are some notable holes we should probably add extra coverage to.

```
Name                                                          Stmts   Miss  Cover
---------------------------------------------------------------------------------
tableauserverclient/__init__.py                                   5      0   100%
tableauserverclient/models/__init__.py                           15      0   100%
tableauserverclient/models/connection_credentials.py             11      5    55%
tableauserverclient/models/connection_item.py                    38      0   100%
tableauserverclient/models/datasource_item.py                   104      5    95%
tableauserverclient/models/exceptions.py                          2      0   100%
tableauserverclient/models/fileupload_item.py                    17     10    41%
tableauserverclient/models/group_item.py                         39      0   100%
tableauserverclient/models/interval_item.py                     112      9    92%
tableauserverclient/models/pagination_item.py                    22      0   100%
tableauserverclient/models/project_item.py                       59      1    98%
tableauserverclient/models/property_decorators.py                63      6    90%
tableauserverclient/models/schedule_item.py                     145      5    97%
tableauserverclient/models/server_info_item.py                   21      0   100%
tableauserverclient/models/site_item.py                         147     14    90%
tableauserverclient/models/tableau_auth.py                       14      6    57%
tableauserverclient/models/tag_item.py                           13      0   100%
tableauserverclient/models/user_item.py                         111      4    96%
tableauserverclient/models/view_item.py                          56      6    89%
tableauserverclient/models/workbook_item.py                     146      8    95%
tableauserverclient/namespace.py                                  1      0   100%
tableauserverclient/server/__init__.py                            9      0   100%
tableauserverclient/server/endpoint/__init__.py                  12      0   100%
tableauserverclient/server/endpoint/auth_endpoint.py             34      3    91%
tableauserverclient/server/endpoint/datasources_endpoint.py     100     21    79%
tableauserverclient/server/endpoint/endpoint.py                  40      0   100%
tableauserverclient/server/endpoint/exceptions.py                16      0   100%
tableauserverclient/server/endpoint/fileuploads_endpoint.py      46     31    33%
tableauserverclient/server/endpoint/groups_endpoint.py           78      4    95%
tableauserverclient/server/endpoint/projects_endpoint.py         40      0   100%
tableauserverclient/server/endpoint/schedules_endpoint.py        46      8    83%
tableauserverclient/server/endpoint/server_info_endpoint.py      11      0   100%
[...]

Issue #114, opened 2016-12-15
Lighter weight Request Options


t8y8:
Today in the code we have request options for various requests, some applicable to many endpoints, some only one...

- pageSize
- pageNumber
- filters
- sort order
- flags (overwrite=T, resolution=high)

They're implemented, in most cases, as nested enums in `request_options.py` requiring sequences like this in user-code:
```
req_option.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
                                   TSC.RequestOptions.Operator.Equals, args.view_name))
```
```
request_options = TSC.RequestOptions(pagesize=1000)
all_workbooks = list(TSC.Pager(server.workbooks, request_options))
```

I'm opening this issue to track discussion of other options.

t8y8:
I went ahead and sketched out several ideas and examples for comparison and discussion. 
I'm not sure that I've got a strong preference on any yet.

```python
##################
###  Option 0 -- Today

# Name equals 'tyler' with a pageSize of 1000
request_options = TSC.RequestOptions(pagesize=1000)
request_options.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name, TSC.RequestOptions.Operator.Equals, 'tyler')
all_workbooks = list(TSC.Pager(server.workbooks, request_options))

# lastLogin is greater than or equal to 12/13/1988 with a page size of 10 starting on page 2
request_options = TSC.RequestOptions(pagesize=10, pageNumber=2)
request_options.filter.add(TSC.Filter(TSC.RequestOptions.Field.LastLogin, TSC.RequestOptions.Operator.GreaterThanOrEqual, "1988-12-13 00:00:00")
all_workbooks = list(TSC.Pager(server.workbooks, request_options))


####################
### Option 1 -- Format Strings

# Name equals 'tyler' with a pageSize of 1000
request_options = TSC.RequestOptions("pageSize=1000", "name:eq:tyler")
all_workbooks = list(TSC.Pager(server.workbooks, request_options))

# lastLogin is greater than or equal to 12/13/1988 with a page size of 10 starting on page 2
request_options = TSC.RequestOptions("pageSize=10", "pageNumber=2", "lastLogin:gte:1988-12-13 00:00:00")
all_workbooks = list(TSC.Pager(server.workbooks, request_options))

########################
### Option 2 -- Validating Dictionary

# Name equals 'tyler' with a pageSize of 1000
request_options = TSC.RequestOptions({"pageSize": 1000, "name": "tyler"}
[...]

PR #131, opened 2017-01-16
Deterministic order

RussTheAerialist:
Address the comment in #101 about sort/filter not being in a deterministic order
(not sure why the doc changes are pulling in here, but I can assure you I didn't make those changes :)

### DIFF
--- a/docs/_includes/analytics.html
+++ b/docs/_includes/analytics.html
@@ -1,0 +1,7 @@
+<!-- Google Tag Manager -->
+<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+})(window,document,'script','dataLayer','GTM-BVCN');</script>
+<!-- End Google Tag Manager -->
--- a/test/test_request_option.py
+++ b/test/test_request_option.py
@@ -97,8 +97,8 @@ class RequestOptionTests(unittest.TestCase):
         # To ensure that this is deterministic, run this a few times
         with requests_mock.mock() as m:
             # Sometimes pep8 requires you to do things you might not otherwise do
-            url = ''.join(self.baseurl, '/workbooks?pageNumber=1&pageSize=100&',
-                          'filter=name:eq:foo,tags:in:[sample,safari,weather]')
+            url = ''.join((self.baseurl, '/workbooks?pageNumber=1&pageSize=100&',
+                          'filter=name:eq:foo,tags:in:[sample,safari,weather]'))
             m.get(url, text=response_xml)
             req_option = TSC.RequestOptions()
             req_option.filter.add(TSC.Filter(TSC.RequestOptions.Field.Tags, TSC.RequestOptions.Operator.In,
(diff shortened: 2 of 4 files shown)


t8y8:
If order is important to us -- why not just store this as a list instead of a set, seems like a lot of hoops to keep it as a set.

(OrderedDict is another option).

We'd lose `.add` as the API, but we can just alias .add to mean `.append`?

RussTheAerialist:
Sets ensure uniqueness.  OrderedDict would be fine too.  A list would be more work since you'll have to check for uniqueness. I don't remember the exact reasons we picked set, though

t8y8:
OrderedDict is a nice data structure to extend in the future if need be, and has a nice key->value relationship.

Actually, that's what we're exploring in #114 anyway, so maybe I can defer my psuedo-objection to that discussion.

docs/_includes/analytics.html
```diff
@@ -0,0 +1,7 @@
+<!-- Google Tag Manager -->
+<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+})(window,document,'script','dataLayer','GTM-BVCN');</script>
+<!-- End Google Tag Manager -->
```
t8y8: Gotta rebase on top of development, this was forked from master.

@jdomingu owes me a 🍺 at this point :)

test/test_request_option.py
```diff
@@ -89,3 +90,20 @@ def test_filter_tags_in(self):
         self.assertEqual(set(['weather']), matching_workbooks[0].tags)
         self.assertEqual(set(['safari']), matching_workbooks[1].tags)
         self.assertEqual(set(['sample']), matching_workbooks[2].tags)
+
+    def test_multiple_filter_options(self):
+        with open(FILTER_MULTIPLE, 'rb') as f:
+            response_xml = f.read().decode('utf-8')
+        # To ensure that this is deterministic, run this a few times
+        with requests_mock.mock() as m:
+            # Sometimes pep8 requires you to do things you might not otherwise do
```
t8y8: lol

docs/_includes/analytics.html
jdomingu: @t8y8 More like  🍻 🍻 🍻 at this point :) Sorry!

RussTheAerialist: I thought I did rebase from development (unless I'm misunderstanding your comment)

Talvalin:
I agree that checking the uniqueness of filters is sensible, so it's weird that the REST API accepts requests with non-unique filters (even though we then state that only the last filter of that type specified is actually used).

The other point is that I'm having a little trouble with the `test_multiple_filter_options` code. When I run tests for `test_request_option.py`, `test_multiple_filter_options` failed consistently because the ordering of the filters in the test url didn't match what the code was returning.

The test url expects the name filter to be first, then the tags filter. My understanding was that the filters are in the same order that they are specified (tags, then name), and so if I change the test url such that the tags filter is first, the test passes consistently. This makes sense.

However, the test as it is here is passing the integration tests. What gives? 😖

t8y8:
@Talvalin yeah thatt's what this is intended to fix -- have you tried rebasing on top of @RussTheAerialist's changes and seeing if the ordering is consistent now?

docs/_includes/analytics.html
t8y8: These changes have popped up on 3 different PRs now... I thought they came form `master` but maybe they're from something else?

PR #139, opened 2017-02-07
initial implement of get all and get specific for Extract Refresh Tasks

RussTheAerialist:
Please note that this is against vnext which means this is provisional support.
Initial support for getting information about Extract Refresh Tasks.  This isn't complete, but wanted to get some eyes on it early for feedback.


samples/refresh_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+    server = TSC.Server(args.server)
+    server.version = '2.5'
+    with server.auth.sign_in(tableau_auth):
+        import pprint
+
+        if args.id:
+            task = server.tasks.get_by_id(args.id)
+            pprint.pprint(task)
+        else:
+            tasks, pagination = server.tasks.get()
```
t8y8: Pager or you won't get them all ;)

Or, a comment that says "First 100"

tableauserverclient/models/schedule_item.py
```diff
@@ -33,6 +33,10 @@ def __init__(self, name, priority, schedule_type, execution_order, interval_item
         self.priority = priority
         self.schedule_type = schedule_type
 
+    def __repr__(self):
+        print(self.__dict__)
+        return "<Schedule#{_id} \"{_name}\" {interval_item}>".format(**self.__dict__)
```
t8y8: Ew. But it's a `repr` so it's ok.

Can you file an item to backfill reprs for all our classes? It's something I intended to do and kept forgetting.

samples/refresh_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+
+import tableauserverclient as TSC
+
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Get all of the refresh tasks available on a server')
+    parser.add_argument('--server', '-s', required=True, help='server address')
+    parser.add_argument('--username', '-u', required=True, help='username to sign into server')
+    parser.add_argument('--site', '-S', default=None)
```
t8y8: I think the other samples use `-t` for the site argument

samples/refresh_tasks.py
RussTheAerialist: hah, um, yeah, my first checkin won't even support paging.  need to implement that.

tableauserverclient/models/schedule_item.py
RussTheAerialist: #140 filed.  And yeah, I guessed you wouldn't like it 👅
(and oh, I need to remove that print statement!)

samples/refresh_tasks.py
RussTheAerialist: I need to talk with @LGraber I'm not sure if makes sense to have paging on these, now that I looked into it.

RussTheAerialist:
@LGraber the commits should now be mostly fixed.  There is one still lingering that I can't get rid of.

tableauserverclient/models/task_item.py
```diff
@@ -0,0 +1,38 @@
...
+    def from_response(cls, xml):
+        parsed_response = ET.fromstring(xml)
+        all_tasks_xml = parsed_response.findall('.//t:task', namespaces=NAMESPACE)
+
+        all_tasks = (TaskItem._parse_task(x) for x in all_tasks_xml)
+
+        return list(all_tasks)
+
+    @classmethod
+    def _parse_task(cls, element):
```
t8y8: On the other Item classes we call this method `_parse_element` and I'll argue for foolish consistency :)

tableauserverclient/server/endpoint/tasks_endpoint.py
```diff
@@ -0,0 +1,30 @@
...
+from .. import TaskItem, PaginationItem
+import logging
+
+logger = logging.getLogger('tableau.endpoint.tasks')
+
+
+class Tasks(Endpoint):
+    @property
+    def baseurl(self):
+        return "{0}/sites/{1}/tasks/extract-refreshes".format(self.parent_srv.baseurl,
```
t8y8: I'm pretty sure we've talked about it, so I might just be dense -- but how do we handle this when we add subscriptions? Is `Tasks` the right endpoint or should it be more specific and wrapped by a higher level endpoint?

t8y8: Would we just call that endpoint `Subscriptions`?

RussTheAerialist: No, I'll expand this to handle both before it is released.  This PR will stay around until the server code is done.  Or, maybe I'll make a long-lived branch until it is done.

tableauserverclient/models/task_item.py
RussTheAerialist: 👍

tableauserverclient/server/endpoint/tasks_endpoint.py
t8y8: Got it.  I'm curious how we'll handle baseurl then, I'll swing by sometime

RussTheAerialist:
I have no idea what's going on here.  Ignore this while I work to clean it up

RussTheAerialist:
okay this is being stupid, I'm closing it and opening a new one.

Issue #141, opened 2017-02-12
Groups.populate_users() doesn't return more than 100 records


cmtoomey:
I'm trying to get a list of usernames in a large group (117 people), but populate_users only returns 100 items. The function doesn't take an upper bound page size, so it sticks with 100

```
#Get list of Tableau_Publishers
group_list = server.groups.get()
group_num = range(len(group_list[0]))
#Find Tableau_Publisher Group ID
for i in group_num:
    if group_list[0][i].name == 'Tableau_Publishers':
        publisher_id = group_list[0][i]
        print server.groups.populate_users(publisher_id)
        print len(group_list[0][i].users)
```

Populate_users appears to obey the default REST API request pagination, but doesn't accept a setting to bump the page size, and there is no documentation on the Groups endpoint at all.

Default should either be higher than 100, or accept input on max size.

t8y8:
Hey Chris!

So, I floated a crazy idea by @benlower and it worked -- can we follow up offline? I think I have your email or can find it easily :)

(PS, there is a woraround for you today -- populate_groups does take a RequestOptions object, so you can pass it a page size of 1000 -- would that work for your use case?)

```
opts = TSC.RequestOptions(pageSize=1000)
server.groups.populate_users(group_item, opts)

cmtoomey:
Always interested in crazy ideas - drop me an email whenever. If you don't have my email handy, it's on the homepage of my blog.

t8y8:
Wooooo.

This should be fixed in development... check it out @cmtoomey and @ugamarkj!

PR #142, opened 2017-02-12
Updates to Group Documentation and user pagination options

cmtoomey:
Per #141, adding a pagination object to groups.populate_users solved the problem. There is still no way to know how many users are in a group, without actually running populate_users and pulling total_available.

I also added documentation so others will be able to walk through group interaction.

### DIFF
--- a/docs/docs/api-ref.md
+++ b/docs/docs/api-ref.md
@@ -256,44 +256,34 @@ Source files: server/endpoint/datasources_endpoint.py, models/datasource_item.py
 Source files: server/endpoint/users_endpoint.py, models/user_item.py
 
 ## Groups
+Groups class contains following attributes: id, name, domain_name, and users. All but users can be accessed directly from Group class.
 
 Get Groups
 
 ```
-groups = server.groups.get()
-```
-Returns two objects - the groups and the pagination item. To access info about groups
-
-```
-group_list = groups[0]
+groups, _ = server.groups.get()
 ```
 
-The group object contains the same information as the standard XML from REST API. For Sites with more than one group, you can access each individual group like this.
-
-```
-#To see the attributes
-dir(group_list[i])
-
-#use these attributes (domain_name, id, name, users)
-group_list[i].attribute
-
-```
-
-Users is a special case, and cannot be populated without an additional step
+Viewing group attributes 
 
 ```
 
 #Get the group ID
-id = group_list[i].id
+id = groups[i].id
+
+```
+Accessing all users in a group requires an additional step
+
+```
 
 #Then you can populate the user list, with an optional page size
-server.groups.populate_users(publisher_id,200)
+server.groups.populate_users(id,200)
 
 #To see how many uses are in the group, to right-size your request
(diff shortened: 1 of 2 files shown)


docs/docs/api-ref.md
```diff
@@ -257,5 +257,45 @@ Source files: server/endpoint/users_endpoint.py, models/user_item.py
...
+Returns two objects - the groups and the pagination item. To access info about groups
+
+```
+group_list = groups[0]
+```
+
+The group object contains the same information as the standard XML from REST API. For Sites with more than one group, you can access each individual group like this.
+
+```
+#To see the attributes
```
t8y8: Technically true, but not really `scl` related, I would this out

docs/docs/api-ref.md
```diff
@@ -257,5 +257,45 @@ Source files: server/endpoint/users_endpoint.py, models/user_item.py
...
+dir(group_list[i])
+
+#use these attributes (domain_name, id, name, users)
```
t8y8: If this isn't doc'd yet (which I'm sure it isn't?) can you list out the group class and it's attributes with a brief one-liner.

OR, if you'd like, submit a docstring for those classes first, and then update the docs?

tableauserverclient/server/endpoint/groups_endpoint.py
```diff
@@ -22,11 +22,11 @@ def get(self, req_options=None):
...
 
     # Gets all users in a given group
-    def populate_users(self, group_item, req_options=None):
+    def populate_users(self, group_item,page_size,req_options=None):
         if not group_item.id:
             error = "Group item missing ID. Group must be retrieved from server first."
             raise MissingRequiredFieldError(error)
-        url = "{0}/{1}/users".format(self.baseurl, group_item.id)
+        url = "{0}/{1}/users?pageSize={2}".format(self.baseurl, group_item.id,page_size)
         server_response = self.get_request(url, req_options)
```
cmtoomey: It will fail - but so will the standard API call. Moving to `Pager` would be very helpful.

cmtoomey:
CLA Request submitted - @t8y8 updates pushed per your comments in the meantime.

t8y8:
@cmtoomey I think I came up with a way to make this use `Pager`.

```python
    def _get_group_users(self, group_id, req_options=None):
        url = "{0}/{1}/users".format(self.baseurl, group_item.id)
        server_response = self.get_request(url, req_options)
        pagination_item = PaginationItem.from_response(server_response.content)
        url = "{0}/{1}/users".format(self.baseurl, group_item.id)
        server_response = self.get_request(url, req_options)
        pagination_item = PaginationItem.from_response(server_response.content)


    # Gets all users in a given group
    def populate_users(self, group_item, req_options=None):
        if not group_item.id:
            error = "Group item missing ID. Group must be retrieved from server first."
            raise MissingRequiredFieldError(error)
        users = Pager(getter=_get_group_users)
        group_item._set_users(UserItem.from_response(server_response.content))
        logger.info('Populated users for group (ID: {0})'.format(group_item.id))
```

PR #144, opened 2017-02-14
Reword no_extract and update signature of added_in

t8y8:
I realized that the feature is to *not* download the extract, not to *only* download the extract.
Oops.

I also changed the way the `parameter_added_in` decorator works after @RussTheAerialist recommended a little cleaner api.

Again manually tested that it works like it should, plus it gets exercised by unittests already in place.


t8y8:
It makes use of (and added to) the new 'annotation' functionality, so this will just throw when called against a Server that doesn't support it, and unblocked testing against the 10.2 beta.

I discussed w/ @RussTheAerialist and we agreed that larger changes should go into vnext, but this was pretty minor and is protected against being used against a current version (see `parameter_added_in`)

PR #145, opened 2017-02-14
Updating high res sample

benlower:
Adding workbook as input so that we minimize risk of returning the
wrong view in the event there are multiple with the same name

### DIFF
--- a/samples/download_view_image.py
+++ b/samples/download_view_image.py
@@ -56,13 +56,13 @@ def main():
         all_workbooks, pagination_item = server.workbooks.get(req_option)
 
         if not all_workbooks:
-            raise LookupError("Workbook with the specificed name was not found.")
+            raise RuntimeError("Workbook with the specificed name was not found.")
         workbook_item = all_workbooks[0]
 
         server.workbooks.populate_views(workbook_item)
         view_item = None
 
-        # Step 3: Find the view we want
+        # Step 3: Find the view we want (note: view names are case sensitive)
         for v in workbook_item.views:
             if v.name == args.view_name:
                 view_item = v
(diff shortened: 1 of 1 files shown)


samples/download_view_image.py
```diff
@@ -46,16 +48,29 @@ def main():
...
+        if not all_workbooks:
+            raise LookupError("Workbook with the specificed name was not found.")
+        workbook_item = all_workbooks[0]
+
+        server.workbooks.populate_views(workbook_item)
+        view_item = None
+
+        # Step 3: Find the view we want
+        for v in workbook_item.views:
+            if v.name == args.view_name:
```
t8y8: Just a note, this will be case sensitive (and probably should be, Server is as well) -- but maybe add a comment to note that?

samples/download_view_image.py
```diff
@@ -46,16 +48,29 @@ def main():
...
+        # Step 2: Query for the workbook that contains the view
         req_option.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
-                                         TSC.RequestOptions.Operator.Equals, args.view_name))
-        all_views, pagination_item = server.views.get(req_option)
-        if not all_views:
+                                         TSC.RequestOptions.Operator.Equals, args.workbook_name))
+        all_workbooks, pagination_item = server.workbooks.get(req_option)
+
+        if not all_workbooks:
+            raise LookupError("Workbook with the specificed name was not found.")
```
t8y8: You can ignore this if you want -- but a LookupError isn't really the error to use here, it's just a `RuntimeError` with a clear string.

This implies a dict lookup or something, failed.

But that's like, hella pedantic, so up to you.

benlower:
Thanks, @t8y8. Modified per your suggestions.

t8y8:
:person_fencing: 🚀

PR #146, opened 2017-02-17
Feature datasource and view tagging

lbrendanl:
This adds the ability to add and delete tags from both datasources and views (currently only workbook tagging is supported through the REST API, even though you can tag datasources and views through Tableau).

I pulled the tagging functionality into a base class that the views/workbooks/datasource items and endpoints classes now inherit from.  Manually tested and added some unit tests.


samples/explore_datasource.py
```diff
@@ -67,6 +67,17 @@ def main():
             print(["{0}({1})".format(connection.id, connection.datasource_name)
                    for connection in sample_datasource.connections])
 
+            # Add some tags to the datasource
+            server.version = 2.6 # Datasource tagging requires server version 2.6
```
t8y8: It's not safe to just set this to 2.6, it'll cause all non 2.6 Servers to fail even on sign in.

That actually reminds me that maybe all samples should call `server.use_highest_version()`

samples/explore_workbook.py
```diff
@@ -82,14 +82,27 @@ def main():
...
+            print("\nWorkbook's old tag set: {}".format(original_tag_set))
+            print("Workbook's new tag set: {}".format(sample_workbook.tags))
             print("Workbook tabbed: {}".format(sample_workbook.show_tabs))
 
             # Delete all tags that were added by setting tags to original
             sample_workbook.tags = original_tag_set
             server.workbooks.update(sample_workbook)
 
+            # Add tag to just one view
+            server.version = 2.6 # View tagging requires server version 2.6
```
t8y8: Same comment about versions here

tableauserverclient/models/view_item.py
```diff
@@ -1,10 +1,12 @@
 import xml.etree.ElementTree as ET
 from .exceptions import UnpopulatedPropertyError
+from .tagged_resource_item import TaggedResourceItem
 from .. import NAMESPACE
 
 
-class ViewItem(object):
+class ViewItem(TaggedResourceItem):
     def __init__(self):
+        super(ViewItem, self).__init__()  # Python2 compatible super
```
t8y8: This comment isn't necessary, we know :)

tableauserverclient/server/endpoint/datasources_endpoint.py
```diff
@@ -16,7 +17,7 @@
 logger = logging.getLogger('tableau.endpoint.datasources')
 
 
-class Datasources(Endpoint):
+class Datasources(TaggedResourcesEndpoint):
```
t8y8: This design concerns me -- I don't like the idea of adding several types of Endpoints in the inheritance chain. What if we need to change endpoints for some other reason?

We can discuss offline, but maybe there's a compositional approach?

test/test_datasource.py
```diff
@@ -105,13 +106,31 @@ def test_update_copy_fields(self):
...
         self.assertEqual(single_datasource.tags, updated_datasource.tags)
         self.assertEqual(single_datasource._project_name, updated_datasource._project_name)
 
+    def test_update_tags(self):
+        with open(ADD_TAGS_XML, 'rb') as f:
+            add_tags_xml = f.read().decode('utf-8')
+        with open(UPDATE_XML, 'rb') as f:
+            update_xml = f.read().decode('utf-8')
+        with requests_mock.mock() as m:
+            m.put(self.baseurl + '/9dbd2263-16b5-46e1-9c43-a76bb8ab65fb/tags', text=add_tags_xml)
```
t8y8: Just cause there are so many endpoints on this test, can you comment what the endpoints are for?

eg
```
# Updating tags
m.put(self.baseurl + '/9dbd2263-16b5-46e1-9c43-a76bb8ab65fb', text=update_xml)
```

test/test_view.py
```diff
@@ -18,12 +20,13 @@ def setUp(self):
...
 
         self.baseurl = self.server.views.baseurl
+        self.siteurl = self.server.views.siteurl
 
     def test_get(self):
         with open(GET_XML, 'rb') as f:
             response_xml = f.read().decode('utf-8')
         with requests_mock.mock() as m:
-            m.get(self.baseurl + '/views', text=response_xml)
+            m.get(self.baseurl + '', text=response_xml)
```
t8y8: Why add an empty string?

test/test_view.py
```diff
@@ -47,7 +50,7 @@ def test_populate_preview_image(self):
         with open(POPULATE_PREVIEW_IMAGE, 'rb') as f:
             response = f.read()
         with requests_mock.mock() as m:
-            m.get(self.baseurl + '/workbooks/3cc6cd06-89ce-4fdc-b935-5294135d6d42/'
+            m.get(self.siteurl + '/workbooks/3cc6cd06-89ce-4fdc-b935-5294135d6d42/'
```
t8y8: Why this change?

tableauserverclient/models/view_item.py
lbrendanl: Well I didn't! I had done it the way I was used to and only thought to check python2 compat as an afterthought... will remove though since it sounds apparent to others.

test/test_view.py
lbrendanl: Just a find/replace error.  Will fix.

test/test_view.py
lbrendanl: It was motivated by the taggedresource change.  For most of our other endpoints, the baseurl is "....site/siteid/resource", for example "...site/siteid/datasources".  Views didn't match this, and as best I could tell this was because of the off logic needed for the populate preview image call, which gets info from the workbooks endpoint instead of the views endpoint.  I thought it would be cleaner to have the Views endpoint behave like the others endpoints.

test/test_datasource.py
lbrendanl: Chatted offline about this, calls seems self-documenting enough.

tableauserverclient/server/endpoint/resource_tagger.py
```diff
@@ -0,0 +1,33 @@
...
+from ...models.tag_item import TagItem
+import logging
+import copy
+
+logger = logging.getLogger('tableau.endpoint')
+
+
+class ResourceTagger(Endpoint):
+    # Add new tags to resource
+    def _add_tags(self, baseurl, resource_id, tag_set):
```
RussTheAerialist: better to not have these prefixed with _ and to preface the entire class with _ to signify that this isn't for end-user use.

RussTheAerialist: Will the base url change between invocations to the same instance of a class?  If not (which I don't think it does based on reading the code), pass in baseurl into the constructor of the class.

t8y8:
:rocket: as long as you run it against a server as a manual test as well :)

tableauserverclient/server/endpoint/resource_tagger.py
LGraber: I am not sure I agree with this. The users of this (the Workbook, Datasource, Views endpoints all construct the baseurl for each call to allow for users to change the version number between any calls. It is not a common case but it is one we allow. We should not break that model.

tableauserverclient/server/endpoint/views_endpoint.py
```diff
@@ -1,20 +1,31 @@
...
+    # Used because populate_preview_image functionaliy requires workbook endpoint
     @property
-    def baseurl(self):
+    def siteurl(self):
         return "{0}/sites/{1}".format(self.parent_srv.baseurl, self.parent_srv.site_id)
 
+    @property
+    def baseurl(self):
+        return "{0}/views".format(self.siteurl)
+
```
LGraber: Thank you for cleaning this up

test/test_view.py
LGraber: I agree and think you did the right thing.

tableauserverclient/models/datasource_item.py
```diff
@@ -53,17 +55,19 @@ def project_name(self):
...
     @property
     def updated_at(self):
         return self._updated_at
 
+    def _get_initial_tags(self):
+        return self._initial_tags
+
+    def _set_initial_tags(self, initial_tags):
+        self._initial_tags = initial_tags
+
```
LGraber: where are these functions used?

t8y8:
:rocket: :rocket:

tableauserverclient/server/endpoint/views_endpoint.py
```diff
@@ -1,20 +1,31 @@
 from .endpoint import Endpoint
 from .exceptions import MissingRequiredFieldError
-from .. import ViewItem, PaginationItem
+from .resource_tagger import _ResourceTagger
+from .. import RequestFactory, ViewItem, PaginationItem
+from ...models.tag_item import TagItem
 import logging
+import copy
```
RussTheAerialist: is this used anywhere?

tableauserverclient/server/endpoint/views_endpoint.py
```diff
@@ -1,20 +1,31 @@
...
 
 class Views(Endpoint):
+    def __init__(self, parent_srv):
+        super(Views, self).__init__(parent_srv)
+        self._resource_tagger = _ResourceTagger(parent_srv)
+
+    # Used because populate_preview_image functionaliy requires workbook endpoint
     @property
-    def baseurl(self):
+    def siteurl(self):
```
RussTheAerialist: Oh, now I feel like this should move to the Endpoint object.  Is siteurl the same everywhere?

lbrendanl: Will refactor this in a separate PR.

PR #147, opened 2017-02-19
[WIP] First run at internal pager

t8y8:
This is an attempt to address #141 from @cmtoomey

It totally works, but I don't like it very much and am very very open to alternative suggestions.

I tried partials and other things but ultimately I couldn't quite get all the arg passing to work.

Only one test fails, and that's because I changed the return signature of the `populate_users` call. 
I manually tested and this works, including setting request options to something custom.

@RussTheAerialist please save me from myself.


t8y8:
Take two!

Thanks to @RussTheAerialist my inability to get the lambda right even though I totally tried it is totally solved -- it works using the internal Pager and is backwards compatible.

BUT, there are some issues:
1) If we redo this API to be lazy-loading, we have a lot of work to do to keep the model up to date with the source, especially when calling add/remove user.
2) I need to do a few more minor things for code cleanup

tableauserverclient/server/endpoint/groups_endpoint.py
```diff
@@ -25,15 +25,22 @@ def get(self, req_options=None):
     # Gets all users in a given group
     @api(version="2.0")
     def populate_users(self, group_item, req_options=None):
+        from .. import Pager
         if not group_item.id:
             error = "Group item missing ID. Group must be retrieved from server first."
             raise MissingRequiredFieldError(error)
+
+        # TODO should this be a list or a Pager directly?
+        all_users = list(Pager(lambda options: self._get_users_for_group(group_item, options), req_options))
```
RussTheAerialist: I really, really, really, really don't like turning this into a list.  Imagine a group that has 10,000 users.  This will do 100 calls to the server before returning. (or 10, depending on what the default page size is, I think it's 100, but I don't remember)

What if all you wanted to do was users[0]?  You'd be fetching way more than you need.

I'd rather break compatibility (since this does actually break backcompat because it doesn't return a pagination object) and called it iter_users() or something.

t8y8: Yeah, this was before we spoke the second time.

I'll need to do more rework to update all the old tests to actually test add/remove user properly. Stay tuned.

RussTheAerialist: ah okay, didn't realize the change was before we talked.

LGraber:
lets talk in the office. I don't totally understand. :)

t8y8:
Recording the high level API:

```
group = server.groups.get().pop()
users = server.groups.iter_users(group)  # iter_users returns a Pager[UserItem]
```

Then you can call `list(users)` if you need a list, or just run it as a normal generator.

This is simple, clear, and works with existing APIs (after some adapting that @RussTheAerialist  and I worked out).

BUT, there's a problem... what do we do with `GroupItem.users` -- today it throws an exception until you call populate, then it gets a list back (the bug is that it's only the first page). With this design there's no way to safely populate the `users` property. If it gets handed the Pager, then it's not repeatable (it'll run until exhausted, then you'd need to call `iter_users` again... and you can't tell where in the iteration you are) 

One option is to make GroupItem.users get a new Pager everytime, which could be appealing -- but then our model would have to carry knowledge of the server session somehow, and I'm not sure how that API would go.

And that's where I'm stuck.

/cc @LGraber @RussTheAerialist 

I'll try to catch you guys during hack week, but you're so busy :P

t8y8:
Other than the work to get tests working, I think, if we like this (elegant and simple ;) ) approach, I can go and update all sub-lists (views, connections, etc), or do those in a new PR.

Thoughts?

t8y8:
Killing this in favor for #204

Issue #148, opened 2017-02-23
Workbook.update does not update server_address


cmtoomey:
This is the code flow
`item = server.workbooks.get_by_id(book_id[1])
server.workbooks.populate_connections(item)
print item.connections[0].id
item.connections[0].server_address = 'server_address'
server.workbooks.update(item)`

When I re-request the same workbook, and query the connection server addresses, they have not been updated.

This appears to be caused by workbook.update() calling the Update Workbook endpoint, but not able to call Update Workbook Connection endpoint, and there isn't a workbook.update_connections (or equivalent).

t8y8:
@cmtoomey You're spot on, we don't have update to the right thing here (yet) and would need to add the appropriate connections endpoints.

Wanna give it a PR ;)

PR #149, opened 2017-02-26
Suggestion for Workbook Update Connection

cmtoomey:
Per #148, an option for updating workbook connections. Having to iterate over each connection individually is sub-optimal, but core REST API only allows one connection update at a time. 

This only accounts for connection string updates and not ports, username, or password changes. 

Use case is DB migration. 

Also - I'm not sure why the two HTML files were included. No edits were made after the fork.

### DIFF
--- a/docs/docs/api-ref.md
+++ b/docs/docs/api-ref.md
@@ -180,6 +180,14 @@ Modifies a workbook. The workbook item object must include the workbook ID and o
 Workbooks.update(wb_item_object)
 ```
 
+### Update workbook connection
+
+Updates a workbook connection string. The workbook connections must be populated before they strings can be updated. 
+
+```py
+Workbooks.update_conn(workbook, workbook.connections[i]
+```
+
 ### Delete Workbook
 
 Deletes a workbook with the given ID.
--- a/docs/docs/api-ref.md
+++ b/docs/docs/api-ref.md
@@ -185,7 +185,7 @@ Workbooks.update(wb_item_object)
 Updates a workbook connection string. The workbook connections must be populated before they strings can be updated. 
 
 ```py
-Workbooks.update_conn(workbook, workbook.connections[i]
+Workbooks.update_conn(workbook, workbook.connections[0])
 ```
 
 ### Delete Workbook
--- a/docs/docs/populate-connections-views.md
+++ b/docs/docs/populate-connections-views.md
@@ -34,6 +34,14 @@ server.workbooks.populate_connections(workbook)
 print([connection.datasource_name for connection in workbook.connections])
 ```
 
+## Update connections for workbooks
+
+```py
+server.workbooks.populate_connections(workbook)
+workbook.connections[i].server_address = 'new_endpoint'
+server.workbooks.update_conn(workbook, workbook.connections[i])
+```
+
 ## Populate connections for data sources
 
 ```py
--- a/docs/docs/populate-connections-views.md
+++ b/docs/docs/populate-connections-views.md
@@ -43,7 +43,7 @@ conn_to_update.server_address = 'new_address'
 conn_to_update.server_port = 1234
 conn_to_update.username = 'username'
 conn_to_update.password = 'password'
-conn_to_update.embed_password = TRUE/FALSE
+conn_to_update.embed_password = True/False
 server.workbooks.update_conn(workbook, conn_to_update)
 ```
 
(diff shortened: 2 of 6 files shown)


tableauserverclient/server/request_factory.py
```diff
@@ -313,6 +312,13 @@ def publish_req_chunked(self, workbook_item, connection_credentials=None):
         parts = {'request_payload': ('', xml_request, 'text/xml')}
         return _add_multipart(parts)
 
+class WorkbookConnection(object):
+    def update_req(self, connection_item):
+        xml_request = ET.Element('tsRequest')
+        connection_element = ET.SubElement(xml_request, 'connection')
+        connection_element.attrib['serverAddress'] = connection_item.server_address.lower()
```
t8y8: May as well add the other attributes that are safe:
port
password

docs/docs/api-ref.md
```diff
@@ -180,6 +180,14 @@ Modifies a workbook. The workbook item object must include the workbook ID and o
 Workbooks.update(wb_item_object)
 ```
 
+### Update workbook connection
+
+Updates a workbook connection string. The workbook connections must be populated before they strings can be updated. 
+
+```py
+Workbooks.update_conn(workbook, workbook.connections[i]
```
t8y8: missing a closing `)`

docs/docs/populate-connections-views.md
```diff
@@ -34,6 +34,14 @@ server.workbooks.populate_connections(workbook)
 print([connection.datasource_name for connection in workbook.connections])
 ```
 
+## Update connections for workbooks
+
+```py
+server.workbooks.populate_connections(workbook)
+workbook.connections[i].server_address = 'new_endpoint'
```
t8y8: This is technically correct, but I don't like the pattern (Not your fault, the way we try to keep the models and the upstream in sync are a little wonky)

Can you document it like this:
```py
server.workbooks.populate_connections(workbook)
conn_to_update = workbook.connections[0]  # Or whichever you want, but i is context-less and can't be copy/pasted
conn_to_update.server_address = 'abcdefg'
conn_to_update.port = 1337
server.workbooks.update_conn(workbook, conn_to_update)
```

/cc @RussTheAerialist

tableauserverclient/server/request_factory.py
cmtoomey: No particular reason - I was just editing in parts that I needed to change. Probably a good idea to add the other ones too.

docs/docs/populate-connections-views.md
cmtoomey: That works - I'll make those changes.

cmtoomey:
Changes made to reflect @t8y8 comments

docs/docs/populate-connections-views.md
```diff
@@ -38,8 +38,13 @@ print([connection.datasource_name for connection in workbook.connections])
...
 ```py
 server.workbooks.populate_connections(workbook)
-workbook.connections[i].server_address = 'new_endpoint'
-server.workbooks.update_conn(workbook, workbook.connections[i])
+conn_to_update = workbook.connections[0]
+conn_to_update.server_address = 'new_address'
+conn_to_update.server_port = 1234
+conn_to_update.username = 'username'
+conn_to_update.password = 'password'
+conn_to_update.embed_password = TRUE/FALSE
```
t8y8: Oh, this should just be `True` or `False` in Python, the all caps could be a different variable

tableauserverclient/server/request_factory.py
```diff
@@ -312,7 +313,8 @@ def publish_req_chunked(self, workbook_item, connection_credentials=None):
         parts = {'request_payload': ('', xml_request, 'text/xml')}
         return _add_multipart(parts)
 
-class WorkbookConnection(object):
+
+    class WorkbookConnection(object):
```
t8y8: Oops, I don't think this should be indented

cmtoomey:
Current fixes complete

LGraber:
Looks good to me. At some point we might rethink how our "update" works. It is supposed to figure out what needs to be updated (based on anything that has changed) but it has gotten lax. Thus you need updateConn. If not, you might just call "update".

LGraber:
Thanks!!

t8y8:
I edited your commit messages and merged. Congrats on your first accepted PR, @cmtoomey !!

cmtoomey:
Yay - thanks @t8y8...now just to wait until it ships

PR #151, opened 2017-02-28
Site can now be queried by name

VathsalaAchar:
This is a Nice To Have for our end. We prefer to work with names rather than ids.

Issue #49, opened 2016-10-05
Add support for getting site by name and contenturl


benlower:
Query Site (http://onlinehelp.tableau.com/current/api/rest_api/en-us/help.htm#REST/rest_api_ref.htm#Query_Site%3FTocPath%3DAPI%2520Reference%7C_____56) has support for getting a site by name and content url:

GET /api/api-version/sites/site-name?key=name
GET /api/api-version/sites/content-url?key=contentUrl

Implement support in TSC for these (e.g. get_by_name(self, site_name))

t8y8:
This is legacy from v1.0 of the API, should we just expose the new Filter/Sort parameters on the get sites call and deprecate this method?

LGraber:
Yes. We already discussed. I meant to update this with exactly what you said. Those methods will be deprecated and we will expose / add Filter / Sort.

RussTheAerialist:
implemented by #153

PR #153, opened 2017-03-02
Site can now be queried by name

VathsalaAchar:
Related to #151 

This is a method to add site by name as the Filter/Sort approach mentioned in #49 is not available to use yet (or is it?).


t8y8:
This looks good and I discussed it with some other members of the core team, let's merge it in!

PR #159, opened 2017-03-23
Extract refresh support

RussTheAerialist:
Okay, that's better.


samples/refresh_tasks.py
```diff
@@ -0,0 +1,71 @@
+####
+# TODO
```
RussTheAerialist: dang it, I need to write the documentation for this.  Coming soon

tableauserverclient/models/schedule_item.py
```diff
@@ -33,6 +33,10 @@ def __init__(self, name, priority, schedule_type, execution_order, interval_item
         self.priority = priority
         self.schedule_type = schedule_type
 
+    def __repr__(self):
+        print(self.__dict__)
```
t8y8: Print?

tableauserverclient/models/task_item.py
```diff
@@ -0,0 +1,38 @@
...
+        parsed_response = ET.fromstring(xml)
+        all_tasks_xml = parsed_response.findall(
+            './/t:task/t:extractRefresh', namespaces=NAMESPACE)
+
+        all_tasks = (TaskItem._parse_task(x) for x in all_tasks_xml)
+
+        return list(all_tasks)
+
+    @classmethod
+    def _parse_task(cls, element):
```
t8y8: _parse_element

ferkaz:
Hey @RussTheAerialist 

This looks great, thanks for working on the extract refresh feature!

Is this safe to be used in production already? If not, do you have any ETA?

Thanks again!

RussTheAerialist:
Hi @ferkaz,  this feature is slated to be a future release of the Tableau Server.  It's not guaranteed in the next release, but should be there.  Oh, and now that I look at it, I forgot to tag them with the rest api version, so I need to go a new pull request for that!

ferkaz:
Sounds good, thank you!

dungkr89:
Hi all, I tried to run the file and got the error: "refresh_tasks.py: error: too few arguments"  when I called:
python samples\refresh_tasks.py -s "http://server.it" -u "my name" -S "Default"
Any idea? Thanks

RussTheAerialist:
Can you give us the full call stack that happens?  Also, which version of
TSC are you using?

-Russell

dungkr89:
@RussTheAerialist  Thank you for your reply, I'm using TSC0.5.1 and the error looks like that:

capture 1

RussTheAerialist:
Oh! okay, I see what you mean.  So the refresh tasks sample has three "commands" built into it.  You can list all of the refresh tasks, you can run one of the refresh tasks, or you get information about the refresh task.  In your case, if you run it like the following, it should work:

```
python samples\refresh_tasks.py -s "http://server.it" -u "my name" -S "Default" list
```

list will output a list of tasks, and you can then call info with the id of the task to get more details about it.  Once you've identified which one you want to run, you can then use the run command to kick it off.

dungkr89:
@RussTheAerialist Thank you, it works well, but I guest error401001: Signin Error, Error signing in to Tableau Server. 
However I'm trying to write code to refresh a datasource on Tableau Server from a location file. It looks like use "Tableau Data Extract Command-Line Utility to refresh extract" to extract data. You have any suggest? Thank you

RussTheAerialist:
If you are getting a sign in error either you typed the password wrong, or you don't have access to the site you are trying to login to.  If you are logging into the default site, you don't need to specify it on the command line.  Try removing the `-S "Default"`.

tarkah:
asdf

andylam1989:
Thank you @RussTheAerialist, I have moved to other project so I havent seen for some days. I have just tested it and it work as your suggestion. It's very nice experience, thank you again ;)

Issue #162, opened 2017-03-28
Can't add users to group


tagyoureit:
When trying to add users to a group, the following error is thrown:
```
Traceback (most recent call last):
  File "create_group.py", line 50, in <module>
    main()
  File "create_group.py", line 45, in main
    server.groups.add_user(group, new_user._id)
  File "/Library/Python/2.7/site-packages/tableauserverclient/server/endpoint/groups_endpoint.py", line 70, in add_user
    user_set.add(new_user)
AttributeError: 'list' object has no attribute 'add'
```

The following script demonstrates this error (this was modified from the `create_group.py` script):
```
####
# This script demonstrates how to create groups using the Tableau
# Server Client.
#
# To run the script, you must have installed Python 2.7.9 or later.
####


import argparse
import getpass
import logging

from datetime import time

import tableauserverclient as TSC


def main():

    parser = argparse.ArgumentParser(description='Creates groups and add a user to it.')
    parser.add_argument('--server', '-s', required=True, help='server address')
    parser.add_argument('--username', '-u', required=True, help='username to sign into server')
    parser.add_argument('--group', '-g', required=True, help='group to create')
    parser.add_argument('--user-to-add', '-ua', required=True, help='user to add')
    parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
                        help='desired logging level (set to error by default)')
    args = parser.parse_args()

    password = getpass.getpass("Password: ")

    # Set logging level based on user input, or error by default
    logging_level = getattr(logging, args.logging_level.upper())
    logging.basicConfig(level=logging_level)

    tableau_auth = TSC.TableauAuth(args.username, password)
    server = TSC.Server(args.server)
    with server.auth.sign_in(tableau_auth):
        group = TSC.GroupItem(args.group)
        group = server.groups.create(group)
        print("Created group {0} with ID: {1}".format(group.name, group._id))
        server.groups.populate_users(group)
        new_user = TSC.UserItem(name=args.user_to_add, site_role='Interactor')
        new_user = server.users.add(new_user)
        print("added {0} with ID: {1}".format(new_user.name, new_user._id))
        server.groups.add_user(group, new_user._id)
        print("added {0} to group {2}".format(new_user.name, group.name))

if __name__ == '__main__':
    main()
```

t8y8:
I swear I've fixed this before, but I see the bug in the code.

We stopped using sets before we even shipped v 0.1

t8y8:
[...]

PR #163, opened 2017-03-28
Fix for 162

t8y8:
Addresses #162 

I also did a search and the only remaining `set()` calls are tags or filters, places where we actually do use tags.

Note that this could go away once we decide on the final pattern for internal paging (like maybe we stop keeping models in sync with the Server), but this is a minimal fix that doesn't change the world in the mean time.

/cc @tagyoureit (can you try with this patch and see if it resolves your issue)


tagyoureit:
Works for me!  Thanks for the quick turnaround.  👍

PR #166, opened 2017-04-10
auto-sanitize filenames on download

### LINKED ISSUE #165: Downloads with invalid characters in the workbook name fail
When downloading workbooks with a name like `Option A or B?` it fails with the error `OSError(22, 'Invalid argument')` on windows because question marks aren't allowed in file names. (Note: when downloading through the browser it converts characters like this to safe ones.. i.e. `>` becomes `_`)

Looking at the [code](https://github.com/tableau/server-client-python/blob/8e6c9078c0f5894e034393f7fcbc9f14f10c5838/tableauserverclient/server/endpoint/workbooks_endpoint.py#L91) I don't see an easy way to overwrite the downloaded workbook name.

Two Requests: 
  1) adding an optional string parameter to set the downloaded workbook name.
  2) adding an optional boolean parameter to download the workbook name the way the browser / server handles it.

(One of these options should take precedence over the other as they are mutually exclusive but both options would be useful).

RussTheAerialist:
fixes #165

### DIFF
--- a/samples/initialize_server.py
+++ b/samples/initialize_server.py
@@ -11,7 +11,6 @@ import logging
 import tableauserverclient as TSC
 
 
-
 def main():
     parser = argparse.ArgumentParser(description='Initialize a server with content.')
     parser.add_argument('--server', '-s', required=True, help='server address')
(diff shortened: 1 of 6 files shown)


sakethoppula:
How to disable auto sanitize. I have to download datasources with '(' and '&' in its name

Issue #152, opened 2017-02-28
Update Contributing Guide


t8y8:
To include steps on how to branch off of `development` -- it'll avoid the merge conflicts from master we're seeing in PRs.

(Note: I'm not a git wizard, but my sequence of steps works every time :) )

RussTheAerialist:
@grbritz addressed this.

PR #167, opened 2017-04-10
Adding doc about the workflow of working an issue

RussTheAerialist:
Addresses #152


RussTheAerialist:
just noticed someone else already did this.

PR #168, opened 2017-04-10
Feature 99 add flag to use server version

RussTheAerialist:
To make @t8y8 happy.

### DIFF
--- a/tableauserverclient/server/server.py
+++ b/tableauserverclient/server/server.py
@@ -42,7 +42,7 @@ class Server(object):
         self.server_info = ServerInfo(self)
 
         if use_server_version:
-            self.use_highest_version()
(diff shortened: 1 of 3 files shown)


tableauserverclient/server/server.py
```diff
@@ -41,6 +41,9 @@ def __init__(self, server_address):
         self.schedules = Schedules(self)
         self.server_info = ServerInfo(self)
 
+        if use_server_version:
+            self.use_highest_version()
```
t8y8: This will throw the warning every time someone uses the flag since this is the legacy endpoint.

I think you want this to be `use_server_version` too :)

t8y8:
🚀

PR #172, opened 2017-04-12
Update changelog and contributors for 0.4

RussTheAerialist:
preping v0.4


RussTheAerialist:
Should I alphabetize the contributors lists?

t8y8:
I've seen both alphabetized and in order contributed. I sorta like the latter because it's kind of cool to see, but I'm fine either way.

PR #176, opened 2017-04-14
initial infrastructure for smoke tests

RussTheAerialist:
This isn't complete, but an idea I'm playing with.  @t8y8 thoughts?

```
$ python setup.py test       # Used to run unit tests for the library
$ python setup.py smoke  # Used to run all the live tests against a server (with environment variables to configure)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
+test.junit.xml
 
 # C extensions
 *.so
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,3 +20,9 @@ versionfile_build = tableauserverclient/_version.py
 tag_prefix = v
 #parentdir_prefix =
 
+[aliases]
+smoke=pytest
+
+[tool:pytest]
+testpaths = test smoke
+addopts = --junitxml=./test.junit.xml
--- a/setup.py
+++ b/setup.py
@@ -16,10 +16,14 @@ setup(
     license='MIT',
     description='A Python module for working with the Tableau Server REST API.',
     test_suite='test',
+    setup_requires=[
+        'pytest-runner>=2.11,<3.0'
+    ],
     install_requires=[
         'requests>=2.11,<2.12.0a0'
     ],
     tests_require=[
-        'requests-mock>=1.0,<1.1a0'
+        'requests-mock>=1.0,<1.1a0',
+        'pytest>=3.0,<4'
     ]
 )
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
+test.junit.xml
```
t8y8: What's gonna end up in here?

setup.cfg
```diff
@@ -20,3 +20,9 @@ versionfile_build = tableauserverclient/_version.py
 tag_prefix = v
 #parentdir_prefix =
 
+[aliases]
+smoke=pytest
```
t8y8: This means smoke will run unittests AND smoke, right?

setup.py
```diff
@@ -16,10 +16,14 @@
...
+    setup_requires=[
+        'pytest-runner'
+    ],
     install_requires=[
         'requests>=2.11,<2.12.0a0'
     ],
     tests_require=[
-        'requests-mock>=1.0,<1.1a0'
+        'requests-mock>=1.0,<1.1a0',
+        'pytest'
```
t8y8: I was wondering when we'd eventually end up moving beyond unittest as our test runner :P

setup.cfg
RussTheAerialist: Yup. I have smoke running both the test directory and the smoke directory.  Test will only run the test directory

t8y8:
Approach seems solid in terms of a runner. 

Feature requests (may not be our first pass):

- Command Line Interface `python setup.py smoke --server=http://server.com`
- Ability to group some tests at a high level 
    - `python setup.py smoke --server=http://server.com --suite=GetOnly`
    - `python setup.py smoke --server=http://server.com --suite=GetAndWrite`
- We'll need to figure out how we handle state between runs (there would be a delay if we just delete a site, for example, and rerun tests too quickly)

.gitignore
RussTheAerialist: It's the junior version of the results that our internal build system needs for test runs

RussTheAerialist: Junit not junior

t8y8: You'll get to be a grown-up results file someday, little test.junit.xml

t8y8:
I checked this out locally and was able to run everything, though it was mad at me that I was using a version of requests-mock that was too new (1.2 -- latest is 1.3). 

Why pin to 1.1?

RussTheAerialist:
@t8y8 oh, hrm, we should pin to 1.x not 1.1.x.  I'll fix that.

RussTheAerialist:
@t8y8 okay with me merging in this bit? I need it for integration work

PR #178, opened 2017-04-18
fix left in debug info

RussTheAerialist:


RussTheAerialist:
I've already left for my class, but I'll push the new version to pypi tonight after I get home

PR #184, opened 2017-04-26
merging the conflict fix back in, I think.

RussTheAerialist:
seriously, every time we cherry pick stuff, something gets messy.


jdomingu:
🚀

Issue #186, opened 2017-05-10
Support for revision history site settings


t8y8:
They're already parsed and added to the model but not serialized

t8y8:
Done in #187

PR #187, opened 2017-05-10
Add revision settings to Update Site

t8y8:
Addresses #186 

(And helps me with some internal stuff)

Updated tests and manually tested internally

### DIFF
--- a/tableauserverclient/models/property_decorators.py
+++ b/tableauserverclient/models/property_decorators.py
@@ -69,7 +69,17 @@ def property_is_valid_time(func):
     return wrapper
 
 
-def property_is_int(range):
+def property_is_int(range, allowed=None):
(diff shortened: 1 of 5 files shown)


LGraber:
How come you didn't added a getter / setter for revisionLimit? How come these values are not read from get requests? I haven't checked the code to see if we are returning them but it seems like we would be.

t8y8:
I didn't change the model, it was already there so I didn't even check, I just added it to the serializer

LGraber:
Huh. Not good that we have a partial implementation like that. Oh well. Thanks for fixing! For your CL, though, it looks like we are simply naming the member "revision_limit" and not using a getter / setter. This doesn't match any of our other properties. Can you fix that? We might also debate on requiring revision_history_enabled to be true to set it but for now, I don't think that matters. Besides that it looks great.

PR #192, opened 2017-05-12
Enable always sorting when using pager 

RussTheAerialist:
because queries are not currently deterministic

### DIFF
--- a/tableauserverclient/server/pager.py
+++ b/tableauserverclient/server/pager.py
@@ -23,7 +23,6 @@ class Pager(object):
         if len(self._options.sort) == 0:
             self._options.sort.add(Sort(RequestOptions.Field.Name, RequestOptions.Direction.Asc))
 
-
     def __iter__(self):
         # Fetch the first page
         current_item_list, last_pagination_item = self._endpoint(self._options)
--- a/tableauserverclient/server/pager.py
+++ b/tableauserverclient/server/pager.py
@@ -1,4 +1,5 @@
 from . import RequestOptions
+from . import Sort
 
 
 class Pager(object):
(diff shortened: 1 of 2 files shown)


tableauserverclient/server/pager.py
```diff
@@ -16,6 +17,11 @@ def __init__(self, endpoint, request_opts=None):
             self._count = ((self._options.pagenumber - 1) * self._options.pagesize)
         else:
             self._count = 0
+            self._options = RequestOptions()
+
+        # Pager assumes deterministic order but solr doesn't guarantee sort order unless specified
+        if len(self._options.sort) == 0:
```
t8y8: Can just be `if not self._options.sort` -- any non empty set will be `True` and any empty set will be `False`

RussTheAerialist:
@t8y8 Not sure what you mean?  It shouldn't miss the case where someone passes in the RequestsOptions because the check for sort = empty set is checked regardless of whether they pass in the requests options or not

LGraber:
We talked about this offline. The client library does not feel like the correct place to fix this. Let's talk more but I think we might want to revert this.

RussTheAerialist:
This is a temporary fix until the server is fixed so that the Pager is usable by default.  Once we release the fix, we can remove this.

LGraber:
The change itself looks fine but if we put in a default sort order for all queries it might not always be on name for every noun. Then Pager would have a weird behavior

RussTheAerialist:
I don't follow what you mean here.  Once the default sort is in, we'll delete this change.  Would you rather not have this temporary workaround in the meantime?

LGraber:
Nope. That sounds good although it might be a slight change in behavior for people

LGraber:
Sorry ... I didn't see your comment after my first comment and before my second comment. :)

PR #199, opened 2017-06-12
Sample group filtering

FFMMM:
Added a script for group filtering.


samples/group_filtering.py
```diff
@@ -0,0 +1,82 @@
...
+
+    with server.auth.sign_in(tableau_auth):
+        server.version = '2.7'
+        group_name = 'SALES NORTHWEST'
+        # Try to create a group named "SALES NORTHWEST"
+        try:
+            group1 = TSC.GroupItem(group_name)
+            group1 = server.groups.create(group1)
+            print(group1)
+        except:
```
RussTheAerialist: You should catch a specific set of exceptions rather than all of them

samples/group_filtering.py
```diff
@@ -0,0 +1,82 @@
...
+        try:
+            group2 = TSC.GroupItem(group_name)
+            group2 = server.groups.create(group2)
+            print(group2)
+        except:
+            print('Group \'%s\' already existed' % group_name)
+
+        group_name = 'SALES+ROMANIA'
+        options = TSC.RequestOptions()
+        options.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
```
RussTheAerialist: Maybe add a comment to explain what you are going here

FFMMM:
Just pushed a commit addressing past feedback.

samples/filter_sort_groups.py
```diff
@@ -0,0 +1,86 @@
...
+
+    parser = argparse.ArgumentParser(description='Filter on groups')
+    parser.add_argument('--server', '-s', required=True, help='server address')
+    parser.add_argument('--username', '-u', required=True, help='username to sign into server')
+    parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
+                        help='desired logging level (set to error by default)')
+    parser.add_argument('-p', default=None)
+    args = parser.parse_args()
+
+    password = ''
```
t8y8: This isn't necessary, the if/then should catch all cases

samples/filter_sort_groups.py
```diff
@@ -0,0 +1,86 @@
...
+    server = TSC.Server(args.server)
+
+    with server.auth.sign_in(tableau_auth):
+        server.version = '2.7'
+        group_name = 'SALES NORTHWEST'
+        # Try to create a group named "SALES NORTHWEST"
+        try:
+            group1 = TSC.GroupItem(group_name)
+            group1 = server.groups.create(group1)
+            print(group1)
```
t8y8: Do we want to print something more meaningful than the repr of a group?

eg "Group created: <repr>"

samples/filter_sort_groups.py
```diff
@@ -0,0 +1,86 @@
...
+        except TSC.ServerResponseError:
+             print('Group \'%s\' already existed' % group_name)
+
+        group_name = 'SALES ROMANIA'
+        # Try to create a group named "SALES ROMANIA"
+        try:
+            group2 = TSC.GroupItem(group_name)
+            group2 = server.groups.create(group2)
+            print(group2)
```
t8y8: Ditto

samples/filter_sort_groups.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # URL Encode the name of the group that we want to filter on
+        # i.e. turn spaces into plus signs
+        filter_group_name = 'SALES+ROMANIA'
+        options = TSC.RequestOptions()
+        options.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
+                              TSC.RequestOptions.Operator.Equals,
+                              filter_group_name))
+
+        filtered_group_paged = server.groups.get(req_options=options)
```
t8y8: Pager again (Samples aren't always just run against new servers, often it's existing servers)

samples/filter_sort_groups.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Return type is a tuple with the first entry as a list of matching groups
+        print(filtered_group_paged[0][0].name)
```
t8y8: tuple unpacking or comment please

samples/filter_sort_groups.py
t8y8: Ignore -- this is filtered

samples/filter_sort_groups.py
FFMMM: Just to clarify, you mean it is not necessary to set `passsword = ''` ?

t8y8: correct, password gets set in all cases because args.p is either `None`, or it isn't.

samples/filter_sort_groups.py
```diff
@@ -0,0 +1,87 @@
...
+        # i.e. turn spaces into plus signs
+        filter_group_name = 'SALES+ROMANIA'
+        options = TSC.RequestOptions()
+        options.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
+                                      TSC.RequestOptions.Operator.Equals,
+                                      filter_group_name))
+
+        filtered_groups, _ = server.groups.get(req_options=options)
+        # Result can either be a matching group or none
+        group_name = filtered_groups.pop().name if len(filtered_groups) != 0 else 'No project named \'' + str(filter_group_name) + '\' found '
```
t8y8:
```
if filtered_groups:
    group_name = filtered_groups.pop()
else:
    error = "No project named '{}' found".format(filter_group_name)
    print(error)
````
Reads cleaner to me, then we aren't overloading the group_name as either a name OR an error string.

Empty lists in python are false-y, so you can just truth test it (`if filtered_groups`). Or there's a bug, because `len(None)` will through an error.

I may run over to clarify some things

File: samples/filter_sort_groups.py [hunk 9]
```diff
@@ -0,0 +1,91 @@
...
+
+
+import argparse
+import getpass
+import logging
+
+import tableauserverclient as TSC
+
+
+def create_example_group(group_name='Example Group', server=None):
```
t8y8: `None` isn't actually a safe default, this should just be a standard required parameter.

PR #201, opened 2017-06-20
Sample project filtering

FFMMM:


samples/filter_sort_projects.py
```diff
@@ -0,0 +1,86 @@
...
+# To run the script, you must have installed Python 2.7.X or 3.3 and later.
+####
+
+import argparse
+import getpass
+import logging
+
+import tableauserverclient as TSC
+
+def create_example_project(_name='Example Project', _content_permissions='LockedToProject',
```
t8y8: Why are these all prefixed with `_`?

samples/filter_sort_projects.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # Set logging level based on user input, or error by default
+    logging_level = getattr(logging, args.logging_level.upper())
+    logging.basicConfig(level=logging_level)
+
+    tableau_auth = TSC.TableauAuth(args.username, password)
+    server = TSC.Server(args.server)
+
+    with server.auth.sign_in(tableau_auth):
+        server.version = '2.7'
```
t8y8: We can call server.use_server_version() instead and this will fail more gracefully

samples/filter_sort_projects.py
```diff
@@ -0,0 +1,86 @@
...
+
+        filter_project_name = 'default'
+        options = TSC.RequestOptions()
+
+        options.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
+                                      TSC.RequestOptions.Operator.Equals,
+                                      filter_project_name))
+
+        filter_projects_paged = server.projects.get(req_options=options)
```
t8y8: This will only get the first 100 projects, we should use the `next() and TSC.Pager` pattern or this will fail on a server with more than 100 projects

samples/filter_sort_projects.py
```diff
@@ -0,0 +1,86 @@
...
+        print(filter_projects_paged[0][0].name)
```
t8y8: What's this trying to do?

Clearer, if possible, is to use tuple unpacking.(Or a comment what this is double indexing into, I can't tell)

samples/filter_sort_projects.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # don't forget to URL encode the query names
+        options.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
+                                      TSC.RequestOptions.Operator.In,
+                                      ['Example+1', 'Example+2', 'Example+3']))
+
+        options.sort.add(TSC.Sort(TSC.RequestOptions.Field.Name,
+                                  TSC.RequestOptions.Direction.Desc))
+
+        matching_projects, pagination_item = server.projects.get(req_options=options)
```
t8y8: Same comment about Pager

samples/filter_sort_projects.py
RussTheAerialist: This is searching for a specific project name.  It should only return one or zero.

t8y8: Ooooooh. Yes, that makes sense for project name, we should be careful for things where names aren't unique (User display name comes to mind) but none of the samples touch that yet.

samples/filter_sort_projects.py
```diff
@@ -0,0 +1,92 @@
...
+# To run the script, you must have installed Python 2.7.X or 3.3 and later.
+####
+
+import argparse
+import getpass
+import logging
+
+import tableauserverclient as TSC
+
+def create_example_project(_name='Example Project', _content_permissions='LockedToProject',
```
t8y8: These shouldn't be prefixed with `_`

samples/filter_sort_projects.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Set logging level based on user input, or error by default
+    logging_level = getattr(logging, args.logging_level.upper())
+    logging.basicConfig(level=logging_level)
+
+    tableau_auth = TSC.TableauAuth(args.username, password)
+    server = TSC.Server(args.server)
+
+    with server.auth.sign_in(tableau_auth):
+        server.version = '2.7'
```
t8y8: `use_server_version()` like the other sample

samples/filter_sort_projects.py
```diff
@@ -0,0 +1,92 @@
...
+                                  TSC.RequestOptions.Direction.Desc))
+
+        matching_projects, pagination_item = server.projects.get(req_options=options)
+        print('Filtered projects are:')
+        for project in matching_projects:
+            print(project.name, project.id)
+
+
+if __name__ == '__main__':
+    main()
```
t8y8: I think there's a whitespace error here, it's hard to tell on github

samples/filter_sort_projects.py
RussTheAerialist: agree with @t8y8 here, let's remain these to not have `_` in them.

samples/filter_sort_projects.py
RussTheAerialist: You need a trailing newline at the EOF.
If you run pycodestyle on the file, it'll tell you all the style things that need to be fixed.

t8y8:
:rocket:

PR #203, opened 2017-06-26
Include extract

tagyoureit:
This is a PR to change the parameter to include the extract from no_extract to include_extract.  The logic changes from the default `no_extract=False` to `include_extract=True`.  There is logic to convert the no_extract into the include_extract parameter and warn the user if it is used.


tableauserverclient/server/endpoint/datasources_endpoint.py
```diff
@@ -73,13 +73,19 @@ def delete(self, datasource_id):
...
             raise ValueError(error)
         url = "{0}/{1}/content".format(self.baseurl, datasource_id)
 
-        if no_extract:
+        if no_extract is False or no_extract is True:
+            import warnings
+            warnings.warn('no_extract is deprecated, use include_extract instead.', DeprecationWarning)
+            include_extract = not no_extract
+
+        if no_extract or not include_extract:
```
t8y8: This can also be `if not include_extract`

tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -87,13 +87,19 @@ def update_conn(self, workbook_item, connection_item):
     # Download workbook contents with option of passing in filepath
     @api(version="2.0")
     @parameter_added_in(no_extract='2.5')
-    def download(self, workbook_id, filepath=None, no_extract=False):
+    @parameter_added_in(include_extract='2.5')
+    def download(self, workbook_id, filepath=None, no_extract=None, include_extract=True):
```
t8y8: For some reason I want to move `include_extract` to before `no_extract` so we just trim the end param, but this isn't mandatory, and others may disagree.

t8y8:
Ugh except there are conflicts with the base branch. Can you give me write access to your fork and I can try and take a look?

tagyoureit:
I added you, @t8y8.  Also made the formatting changes suggested and you can pull those in as well.

t8y8:
Good news, I confirmed that it works manually and with tests.

Bad news, the git-fu is beyond me, I think it's because d45's checking originated from master, then was merged into development, and you're branched properly from development.

@RussTheAerialist is the better git-fuer.

But the code looks good to me.

We should: 1) review tests and make sure no_extract isn't used anywhere and 2) update the docs, but given the merge conflicts, I say that can be a 2nd checkin

PR #204, opened 2017-06-29
Auto Paging for Users in Groups

t8y8:
Subsumes #147 because the git history got all weird.

I still need to make the tests work properly, but a manual test indicates this all works well.

/cc @cmtoomey who may want to try this out

### DIFF
--- a/tableauserverclient/server/endpoint/groups_endpoint.py
+++ b/tableauserverclient/server/endpoint/groups_endpoint.py
@@ -71,22 +71,12 @@ class Groups(Endpoint):
     @api(version="2.0")
     def remove_user(self, group_item, user_id):
         self._remove_user(group_item, user_id)
-        try:
-            users = group_item.users
-        except UnpopulatedPropertyError:
-            # If we aren't populated, do nothing to the user list
-            pass
         logger.info('Removed user (id: {0}) from group (ID: {1})'.format(user_id, group_item.id))
 
     # Adds 1 user to 1 group
     @api(version="2.0")
     def add_user(self, group_item, user_id):
-        new_user = self._add_user(group_item, user_id)
-        try:
-            users = group_item.users
-        except UnpopulatedPropertyError:
-            # If we aren't populated, do nothing to the user list
-            pass
+        self._add_user(group_item, user_id)
         logger.info('Added user (id: {0}) to group (ID: {1})'.format(user_id, group_item.id))
 
     def _remove_user(self, group_item, user_id):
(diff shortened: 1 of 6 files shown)


tableauserverclient/server/endpoint/groups_endpoint.py
```diff
@@ -59,10 +73,6 @@ def remove_user(self, group_item, user_id):
         self._remove_user(group_item, user_id)
         try:
             users = group_item.users
```
t8y8: This is actually unnecessary because every call is a new Pager now, but we could return the list from `add_user` or we can just remove this from the call enitrely

t8y8:
Usage:
```py

mapping = {}

groups = TSC.Pager(server.groups)
for group in groups:
    server.groups.populate_users(group)
    mapping[group.name] = group.users
```

tableauserverclient/server/endpoint/groups_endpoint.py
```diff
@@ -57,28 +71,12 @@ def create(self, group_item):
     @api(version="2.0")
     def remove_user(self, group_item, user_id):
         self._remove_user(group_item, user_id)
```
t8y8: At this point it may not even be necessary to have the `_remove_user` function since we don't do anything special now

tableauserverclient/server/endpoint/groups_endpoint.py
```diff
@@ -57,28 +71,12 @@ def create(self, group_item):
...
-                    break
-        except UnpopulatedPropertyError:
-            # If we aren't populated, do nothing to the user list
-            pass
         logger.info('Removed user (id: {0}) from group (ID: {1})'.format(user_id, group_item.id))
 
     # Adds 1 user to 1 group
     @api(version="2.0")
     def add_user(self, group_item, user_id):
-        new_user = self._add_user(group_item, user_id)
```
t8y8: ditto

t8y8:
@RussTheAerialist 

Here's a cleaned up rev -- I've been testing manually for a bit and seems to work the way I think it should.

I'd like to check this in and get @cmtoomey and @ugamarkj to check it out. If they like it, I can generalize it (I've got an idea to refactor it with passing callables all the way down) to other sub-models, whichever feels easier to maintain.

tableauserverclient/server/pager.py
```diff
@@ -4,12 +4,23 @@
...
+    Will loop over anything that returns (List[ModelItem], PaginationItem).
     """
 
     def __init__(self, endpoint, request_opts=None):
-        self._endpoint = endpoint.get
+
+        if hasattr(endpoint, 'get'):
+            # The simpliest case is to take an Endpoint and call its get
+            self._endpoint = endpoint.get
+        else:
```
RussTheAerialist: Elif callable(...):
   ...
Else:
   raise ...

(Sorry I'm on my phone, didn't bring a computer with me to California)

t8y8: Oh good call, a little cleaner!

PR #212, opened 2017-08-07
First pass at options builder

LGraber:
Not ready to actually submit this but I find our current syntax for building a filter / sort set ... yucky. I implemented a simple builder pattern. "_" prefixed methods have special meaning. For all other methods, the name is the property you are filtering on and the 2 params are an operator and value. I just accept simple strings that actually look like the operator. I imagine later I can add "_include" when we expose fields. You can see in the sample that it makes it much easier to read. All of our samples only have one filter so it isn't so bad but it would get bad quickly if you did any more

### DIFF
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -1,3 +1,6 @@
+from .filter import Filter
+from .sort import Sort
+
 class RequestOptionsBase(object):
     def apply_query_params(self, url):
         raise NotImplementedError()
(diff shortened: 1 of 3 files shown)


tableauserverclient/server/request_options.py
```diff
@@ -25,6 +28,35 @@ class Direction:
         Desc = 'desc'
         Asc = 'asc'
 
+    class Builder:
+        def __init__(self):
+            self.__pagenumber = 1
```
t8y8: I'm not sure the `__` is necessary (it invokes name mangling to avoid inheritance problems but here I think you can just make it the `_`... it's more 'polite' form :P

RussTheAerialist:
How about the contract for the Filters/Request Options/etc is that we call `str(...)` and the objects are required to implement `__str__(self)`, and that's our contract.  It's simple, and easy to understand.  No need for a build method, and when we are creating the request, we just have to do `format('...{}...', options)` and python should automatically call `__str__`.

t8y8:
@RussTheAerialist I like that contract

tableauserverclient/server/request_options.py
LGraber: The double underscore was needed to disambiguate with the call to _pagenumber that I was specifically handling. If I don't do this, it actually fails

LGraber:
I think I am a bit confused. If I don't call _build() and so still have a Builder object, what do I call to actually get a RequestOptions object? I need to call something, right? Also, if we don't have a "build" method, then why did I call it a Builder?

t8y8:
I think @RussTheAerialist is proposing you just create a `RequestOptions` object and it  has all the methods you currently have on `Builder`.

Then, the contract for endpoints is to call `str(request_object)` -- which is really easy to test and manage

RussTheAerialist:
Closing for inactivity

PR #213, opened 2017-08-07
Adding get_default method to the projects endpoint

LGraber:
Want a quick view on this. I sent Russell a mail cause our endpoints are broken because of the default sorting that was added. Ignoring that, I added support for get_default and put an implementation for older versions so that our samples could be updated, too. I accidentally parented off of builders so I probably need to reparent. Tests are still to be done

### DIFF
--- a/samples/options_builder.py
+++ b/samples/options_builder.py
@@ -1,4 +1,5 @@
 import sys
+sys.path.append("~/git/server-client-python")
 
 import tableauserverclient as TSC
 
--- a/samples/publish_workbook.py
+++ b/samples/publish_workbook.py
@@ -41,14 +41,13 @@ def main():
     # Step 1: Sign in to server.
     tableau_auth = TSC.TableauAuth(args.username, password)
     server = TSC.Server(args.server)
+    server.use_server_version()
 
     overwrite_true = TSC.Server.PublishMode.Overwrite
 
     with server.auth.sign_in(tableau_auth):
 
-        # Step 2: Get all the projects on server, then look for the default one.
-        all_projects, pagination_item = server.projects.get()
-        default_project = next((project for project in all_projects if project.is_default()), None)
+        default_project = server.projects.get_default()
 
         # Step 3: If default project is found, form a new workbook item and publish.
         if default_project is not None:
--- a/tableauserverclient/server/endpoint/projects_endpoint.py
+++ b/tableauserverclient/server/endpoint/projects_endpoint.py
@@ -1,7 +1,12 @@
 from .endpoint import Endpoint, api
 from .exceptions import MissingRequiredFieldError
-from .. import RequestFactory, ProjectItem, PaginationItem
+from .. import RequestFactory, RequestOptions, Filter, Pager, ProjectItem, PaginationItem
 import logging
+
+try:
+    from distutils2.version import NormalizedVersion as Version
+except ImportError:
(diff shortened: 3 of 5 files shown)


samples/options_builder.py
```diff
@@ -1,4 +1,5 @@
 import sys
```
t8y8: This file shouldn't be in the commit for this, right? I think it's part of Options Builder

samples/publish_workbook.py
```diff
@@ -41,14 +41,13 @@ def main():
     # Step 1: Sign in to server.
     tableau_auth = TSC.TableauAuth(args.username, password)
     server = TSC.Server(args.server)
+    server.use_server_version()
```
t8y8: Thanks :)

tableauserverclient/server/endpoint/projects_endpoint.py
```diff
@@ -20,6 +25,20 @@ def get(self, req_options=None):
         all_project_items = ProjectItem.from_response(server_response.content)
         return all_project_items, pagination_item
 
+    def get_default(self):
+        if (Version(self.parent_srv.version) < Version("2.7")):
+            for proj in Pager(self):
```
t8y8: Nit `Pager(self.parent_srv.projects)` is a little clearer than remembering what `self` is pointing to

tableauserverclient/server/endpoint/projects_endpoint.py
```diff
@@ -20,6 +25,20 @@ def get(self, req_options=None):
...
+                if proj.is_default():
+                    return proj
+
+            return None
```
t8y8: You can nix this, `None` is always returned implicitly.

tableauserverclient/server/endpoint/projects_endpoint.py
```diff
@@ -20,6 +25,20 @@ def get(self, req_options=None):
...
+        else:
+            ro = RequestOptions()
+            ro.filter.add(Filter("name", "eq", "Default"))
```
t8y8: Sometimes it's uppercase and sometimes it's lower case (I think pre 9.0 installs vs post) so you'll need to normalize the casing

tableauserverclient/server/endpoint/projects_endpoint.py
```diff
@@ -20,6 +25,20 @@ def get(self, req_options=None):
...
+            items, _ = self.get(ro)
+            return items[0] if (len(items) > 0) else None
```
t8y8: `return items.pop() if items else None`

tableauserverclient/server/endpoint/projects_endpoint.py
LGraber: I don't think we support pre-9.0 anymore so I am okay with that. :)

tableauserverclient/server/endpoint/projects_endpoint.py
LGraber: ok

LGraber:
Thanks for all the feedback

tableauserverclient/server/endpoint/projects_endpoint.py
t8y8: No, like, if their Server was upgraded from pre 9.0.

You must do this!

RussTheAerialist:
Closing for inactivity

PR #219, opened 2017-09-13
194 fix namespace issue

RussTheAerialist:
Introduce detection of namespace url for backwards compatibility.

### DIFF
--- a/tableauserverclient/namespace.py
+++ b/tableauserverclient/namespace.py
@@ -1,5 +1,9 @@
+from xml.etree import ElementTree as ET
+import re
+
 OLD_NAMESPACE = 'http://tableausoftware.com/api'
 NEW_NAMESPACE = 'http://tableau.com/api'
+NAMESPACE_RE = re.compile(r'\{(.*?)\}')
 
 class Namespace(object):
     def __init__(self):
@@ -13,16 +17,13 @@ class Namespace(object):
         if self._detected:
             return
 
-        lines = (n for n in xml.split('\n') if n.startswith('<tsResponse'))
+        if not xml.startswith(b'<?xml'):
+            return  # Not an xml file, don't detect anything
 
-        for n in lines:
-            pieces = (p for p in n.split('xmlns') if p.startswith('='))
-            for piece in pieces:
-                ns = piece.replace('=', '').replace('"', '').strip()
-                if ns == OLD_NAMESPACE:
-                    self._namespace = {'t': OLD_NAMESPACE}
-                    self._detected = True
-                elif ns == NEW_NAMESPACE:
-                    self._namespace = {'t': NEW_NAMESPACE}
-                    self._detected = True
-
+        root = ET.fromstring(xml)
+        matches = NAMESPACE_RE.match(root.tag)
+        if matches:
+            detected_ns = matches.group(1)
+            if detected_ns in (OLD_NAMESPACE, NEW_NAMESPACE):
+                self._namespaces = {'t': detected_ns }
+                self._detected = True
--- a/tableauserverclient/namespace.py
+++ b/tableauserverclient/namespace.py
@@ -1,1 +1,25 @@
-NAMESPACE = {'t': 'http://tableau.com/api'}
+OLD_NAMESPACE = 'http://tableausoftware.com/api'
+NEW_NAMESPACE = 'http://tableau.com/api'
+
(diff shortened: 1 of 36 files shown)


tableauserverclient/namespace.py
```diff
@@ -1 +1,28 @@
...
+        self._detected = False
+
+    def __call__(self):
+        return self._namespace
+
+    def detect(self, xml):
+        if self._detected:
+            return
+
+        lines = (n for n in xml.split('\n') if n.startswith('<tsResponse'))
```
t8y8: There's breaks all ova' them tests because of str/bytes stuff.

I think we can reduce the parsing overhead a bit as well, by parsing the namespace outta the root element.

Something like this:
```
>>> root = tree.getroot()
>>> root.tag
'{http://maven.apache.org/POM/4.0.0}project'
```

And then run the find over just that smaller string

tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -39,6 +39,7 @@ def _make_request(self, method, url, content=None, request_object=None,
             parameters['data'] = content
 
         server_response = method(url, **parameters)
+        self.parent_srv._namespace.detect(server_response.content)
```
t8y8: I see, the first request, of any kind, checks and caches itself, then we shortcut on subsequent requests.

I like the deisgn.

tableauserverclient/server/server.py
```diff
@@ -95,6 +97,10 @@ def baseurl(self):
         return "{0}/api/{1}".format(self._server_address, str(self.version))
 
     @property
+    def namespace(self):
```
t8y8: Is 'xml_namespace' a better name? (No strong opinion here)

tableauserverclient/namespace.py
RussTheAerialist: hrm.  Weird, I even tested wth python 3.6 on my mac.  yeah I think I like this solution better.

tableauserverclient/server/endpoint/endpoint.py
RussTheAerialist: yup.  I think the only place where we might get in trouble is if the first request isn't the use server version workflow.  At this point, I want to make that the default behavior. @LGraber any objections to always determining the server version to help solve this issue?

tableauserverclient/server/server.py
RussTheAerialist: I doubt anyone outside of TSC contributors will use that property.  I don't have a strong opinion either way.

tableauserverclient/namespace.py
```diff
@@ -1 +1,30 @@
...
+            return
+
+        if not xml.startswith(b'<?xml'):
+            return  # Not an xml file, don't detect anything
+
+        root = ET.fromstring(xml)
+        matches = NAMESPACE_RE.match(root.tag)
+        if matches:
+            detected_ns = matches.group(1)
+            if detected_ns in (OLD_NAMESPACE, NEW_NAMESPACE):
```
LGraber: Is there a reason we don't throw here if we don't know what it is? Won't we fail in an odd way later?

tableauserverclient/server/__init__.py
```diff
@@ -4,7 +4,7 @@
 from .sort import Sort
 from .. import ConnectionItem, DatasourceItem,\
     GroupItem, PaginationItem, ProjectItem, ScheduleItem, SiteItem, TableauAuth,\
-    UserItem, ViewItem, WorkbookItem, TaskItem, NAMESPACE
+    UserItem, ViewItem, WorkbookItem, TaskItem, namespace
```
LGraber: Why is this the only one lowercase?

tableauserverclient/server/endpoint/endpoint.py
LGraber: Why would it be a problem if that wasn't the first call? I am not sure I see how this only happens on the first call?

tableauserverclient/namespace.py
t8y8: Not every request necessarily returns xml (deletes for example). 

This just skips it an checks on the next request. Once it succeeds we cache the result and don't check again.

We default to the new namespace so nothing should fail

RussTheAerialist: no real reason.  I'll add a new exception

LGraber:
:rocket:

PR #231, opened 2017-09-23
205 - Pager/Fetcher all the things!

t8y8:
Here's a ~~checkpoint~~ PR for where I'm at in concerting all the models to 'fetch-on-demand' after you call `populate_XXX` on something.

Mostly it's a straight forward conversation but a few places have some interesting caveats so posting here for feedback.

EDIT: I've converted all the endpoints, I believe, so they all have on-demand fetching. WOO.

Also added test coverage where there wasn't any. This covers @grbritz's bug as well


tableauserverclient/server/endpoint/users_endpoint.py
```diff
@@ -70,15 +70,24 @@ def add(self, user_item):
     # Get workbooks for user
     @api(version="2.0")
     def populate_workbooks(self, user_item, req_options=None):
+        from .. import Pager
```
RussTheAerialist: Why do this here?

t8y8: I did that on the 'users-in-groups' bugfix and just copied the pattern.

You could say "Only imported when needed" but there's not really any penalty to just importing it at the top.

I can move it to a top-level import

PR #236, opened 2017-09-27
Add support for PDF and CSV Export

t8y8:
TODO:
- [x] Add all the options for PDF export (landscape, size)
- [x] Add tests for the new endpoints
- [x] Should PDF also return an iterable-stream?
- [x] How do we feel about csv returning an iterable-stream?
- [x] How do we feel about calling 'data' > 'csv' on the model?
- [x] Clean up error strings (They're find/replace right now)
- [x] Do we want to rename `image` to `png`?
- [x] I called 'type' for the pdf size 'output_size' but open to better names. I don't like type

/cc @RussTheAerialist


t8y8:
Thanks!

Yeah, I agree we need to revisit the model-call-sync problems (tags still feel weird to me).
This implementation was the 'least bad' Russell and I could think of to solve paging lists of sub-models (like users in groups).

This PR has more work, so stay tuned for the updates

RussTheAerialist:
RE: data vs. csv - At some point we want to support json output.  Would you see that as being a "json" endpoint also?

RE: type/outputSize - PageSize would fit the domain of pdf better than outputSize.

RE: Iterable Stream - for csv it makes sense(ish), but does pdf really make sense to be iterable?  What's the use case?  pdfs will generally be small enough since we are only doing one sheet/dashboard at a time.

t8y8:
re: json -- yeah, I think I slightly prefer that to something like `populate_data(type=json)`. But I don't feel very strongly.

re: pageSize -> But we use that for the regular request options. I don't like it meaning 1-1000 on one and 'Letter' on another :P Any other ideas?

re: iterables -> It could technically be a 10000 by 100000 pixel dashboard with 1 billion marks. If we're not worried about that I can keep it simple and leave it as is. I agree it's very much an edge case and just paranoia.

test/assets/file_upload.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
t8y8: er, this was to cover your other item but I mixed my CLs. Ignore for now.

tableauserverclient/models/view_item.py
```diff
@@ -49,6 +57,20 @@ def preview_image(self):
...
 
     @property
+    def pdf(self):
+        if self._pdf is None:
+            error = "View item must be populated with its pdf first."
+            raise UnpopulatedPropertyError(error)
+        return self._pdf()
+
+    @property
+    def csv(self):
```
RussTheAerialist: Man, this pattern is screaming for a decorator or something.

PR #237, opened 2017-09-28
fix ns issue with tasks get_by_id

ajbosco:
This method is missing the namespace parameter.

### DIFF
--- a/tableauserverclient/server/endpoint/tasks_endpoint.py
+++ b/tableauserverclient/server/endpoint/tasks_endpoint.py
@@ -30,7 +30,7 @@ class Tasks(Endpoint):
         logger.info("Querying a single task by id ({})".format(task_id))
         url = "{}/{}".format(self.baseurl, task_id)
         server_response = self.get_request(url)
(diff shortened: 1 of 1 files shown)


t8y8:
@ajbosco thank you for the report!

The fix looks good to me, but I can't recall if you've submitted the CLA with us before; have you had a chance to do so?

ajbosco:
I haven't submitted the CLA. What's the process for that?

t8y8:
https://tableau.github.io/contributing.html outlines the process. 
Basically sign the PDF and email in!

Once we have that we can accept the PR and get this merged

t8y8:
I reviewed our tests and realized that this also could use a test case that would have caught the miss.

Would you be able to add a test case? You can model it off of (https://github.com/tableau/server-client-python/blob/ada7dbcc74838da2f9620f5a61135cea0e677014/test/test_workbook.py#L75)

kaogilvie:
This would be helpful to merge -- currently running off this commit

RussTheAerialist:
I don't think we need the CLA in this case, so I'm going to merge this in.

PR #240, opened 2017-10-06
Add subscriptions

williamlang:
Adding subscriptions!

### DIFF
--- a/tableauserverclient/models/subscription_item.py
+++ b/tableauserverclient/models/subscription_item.py
@@ -1,38 +1,45 @@
 import xml.etree.ElementTree as ET
 from .exceptions import UnpopulatedPropertyError
-from .. import NAMESPACE
 from .target import Target
 
 class SubscriptionItem(object):
 
-    def __init__(self, id_, schedule_id, user_id, target):
-        self.id = id_
+    def __init__(self, subject, schedule_id, user_id, target):
+        self.id = None
+        self.subject = subject
         self.schedule_id = schedule_id
         self.user_id = user_id
         self.target = target
 
     def __repr__(self):
-        return "<Subscription#{id} schedule_id({schedule_id}) user_id({user_id}) \
-            target({target})".format(**self.__dict__)
+        if self.id is not None:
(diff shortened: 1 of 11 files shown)


tableauserverclient/models/subscription_item.py
```diff
@@ -0,0 +1,58 @@
...
+        if content_element is not None:
+            target = Target(content_element.get('id', None), content_element.get('type'))
+
+        if user_element is not None:
+            user_id = user_element.get('id')
+
+        id_ = element.get('id', None)
+        subject = element.get('subject', None)
+        sub = cls(subject, schedule_id, user_id, target)
+        sub._set_id(id_)
```
t8y8: Why do you have the id only settable via a private method? (If other Items do this to it's fine, just asking).
The other attributes are all available via `__init__`.

test/test_subscription.py
```diff
@@ -0,0 +1,50 @@
...
+
+        subscription = all_subscriptions[0]
+        self.assertEqual('382e9a6e-0c08-4a95-b6c1-c14df7bac3e4', subscription.id)
+        self.assertEqual('View', subscription.target.type)
+        self.assertEqual('cdd716ca-5818-470e-8bec-086885dbadee', subscription.target.id)
+        self.assertEqual('c0d5fc44-ad8c-4957-bec0-b70ed0f8df1e', subscription.user_id)
+        self.assertEqual('Not Found Alert', subscription.subject)
+        self.assertEqual('7617c389-cdca-4940-a66e-69956fcebf3e', subscription.schedule_id)
+
+    def test_get_single_subscription(self):
```
t8y8: `test_get_subscription_by_id`

tableauserverclient/models/subscription_item.py
williamlang: modeled it after the UserItem -- because when creating a new User you don't have an id.

Same goes for Subscription.

t8y8:
:rocket:

t8y8:
@RussTheAerialist says :rocket:

Issue #179, opened 2017-04-19
Get usage statistics for Views


d45:
Currently, there is no way to get usage stats for Views using the TSC library.  Getting usage stats would be helpful if you wanted to monitor how the site was being used.  The stats could identify popular content and content that wasn't viewed that could be removed, etc.  

You can retrieve the usage stats for views in the REST API by including a query in the URL in a GET request for views:  '?includeUsageStatistics=true'  

The TSC library is set up for storing this data, but it isn't implemented.  The ViewItem class has the attribute `total_views` , but there is no way to add the usage data.  To collect the data, you could add a new endpoint, similar to the `populate_views()` endpoint for workbooks, or you could modify the `views.get()` endpoint to return the data by default, or with a parameter.  Whatever works best with your design model. 

Here's one idea using a parameter: 

From `class Views`: 

```
   def get(self, req_options=None, usage=False):
          logger.info('Querying all views on site')
           url = "{0}/views?includeUsageStatistics={1}".format(self.baseurl,usage)
           server_response = self.get_request(url, req_options)
           pagination_item = PaginationItem.from_response(server_response.content)
           all_view_items = ViewItem.from_response(server_response.content)
           return all_view_items, pagination_item
```

This implementation works, but it changes the signature of the get method. It is false by default to mirror the REST API behavior. 

Cheers,
Dave

RussTheAerialist:
implemented by #242

PR #242, opened 2017-10-24
added ability to get usage statistics for views

shinchris:
Issue #179.

### DIFF
--- a/tableauserverclient/server/endpoint/views_endpoint.py
+++ b/tableauserverclient/server/endpoint/views_endpoint.py
@@ -26,7 +26,9 @@ class Views(Endpoint):
     @api(version="2.2")
     def get(self, req_options=None, usage=False):
         logger.info('Querying all views on site')
-        url = "{0}?includeUsageStatistics={1}".format(self.baseurl, usage)
+        url = self.baseurl
+        if usage:
+            url += "?includeUsageStatistics=true"
         server_response = self.get_request(url, req_options)
         pagination_item = PaginationItem.from_response(server_response.content, self.parent_srv.namespace)
         all_view_items = ViewItem.from_response(server_response.content, self.parent_srv.namespace)
--- a/tableauserverclient/server/endpoint/views_endpoint.py
+++ b/tableauserverclient/server/endpoint/views_endpoint.py
@@ -24,9 +24,10 @@ class Views(Endpoint):
         return "{0}/views".format(self.siteurl)
 
     @api(version="2.2")
-    def get(self, req_options=None):
+    def get(self, req_options=None, usage=False):
         logger.info('Querying all views on site')
-        server_response = self.get_request(self.baseurl, req_options)
+        url = "{0}?includeUsageStatistics={1}".format(self.baseurl, usage)
+        server_response = self.get_request(url, req_options)
         pagination_item = PaginationItem.from_response(server_response.content, self.parent_srv.namespace)
         all_view_items = ViewItem.from_response(server_response.content, self.parent_srv.namespace)
         return all_view_items, pagination_item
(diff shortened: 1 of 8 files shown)


t8y8:
Should we add a test to cover this case?

tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -118,19 +118,21 @@ def download(self, workbook_id, filepath=None, include_extract=True, no_extract=
...
-            return self._get_views_for_workbook(workbook_item)
+            return self._get_views_for_workbook(workbook_item, usage)
 
         workbook_item._set_views(view_fetcher)
         logger.info('Populated views for workbook (ID: {0}'.format(workbook_item.id))
 
-    def _get_views_for_workbook(self, workbook_item):
-        url = "{0}/{1}/views".format(self.baseurl, workbook_item.id)
+    def _get_views_for_workbook(self, workbook_item, usage):
+        url = "{0}/{1}/views".format(self.baseurl, workbook_item.id, usage)
```
t8y8: `usage` doesn't need to be in the `.format`

t8y8:
:rocket:

tableauserverclient/server/endpoint/views_endpoint.py
```diff
@@ -24,9 +24,12 @@ def baseurl(self):
         return "{0}/views".format(self.siteurl)
 
     @api(version="2.2")
-    def get(self, req_options=None):
+    def get(self, req_options=None, usage=False):
         logger.info('Querying all views on site')
-        server_response = self.get_request(self.baseurl, req_options)
+        url = self.baseurl
+        if usage:
+            url += "?includeUsageStatistics=true"
```
RussTheAerialist: This could lead to the url being `?includeUsageStatistics=true?page=2` if you pass in request options.  Can you add a test to check that the url being generated in this case would be `?includeUsageStatistics=true&page=2` ?

You'll probably want to amend request options so that it inspects the url for the existence of `?` and then uses an `&` to append options to the query string.

t8y8:
```py
def apply_query_params(self, url):
    params = []

    url, existing_params = url.split('?')

    params.append(existing_params)

    if self.page_number:
        params.append('pageNumber={0}'.format(self.page_number))
    if self.page_size:
        params.append('pageSize={0}'.format(self.page_size))
    if len(self.sort) > 0:
        sort_options = (str(sort_item) for sort_item in self.sort)
        ordered_sort_options = sorted(sort_options)
        params.append('sort={}'.format(','.join(ordered_sort_options)))
    if len(self.filter) > 0:
        filter_options = (str(filter_item) for filter_item in self.filter)
        ordered_filter_options = sorted(filter_options)
        params.append('filter={}'.format(','.join(ordered_filter_options)))

```

That works in cases where the url contains an existing parameter

test/assets/view_get_usage.xml
```diff
@@ -2,14 +2,16 @@
 <tsResponse xmlns="http://tableau.com/api" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://tableau.com/api http://tableau.com/api/ts-api-2.3.xsd">
     <pagination pageNumber="1" pageSize="100" totalAvailable="2" />
     <views>
-        <view id="d79634e1-6063-4ec9-95ff-50acbf609ff5" name="ENDANGERED SAFARI" contentUrl="SafariSample/sheets/ENDANGEREDSAFARI">
+        <view id="d79634e1-6063-4ec9-95ff-50acbf609ff5" name="foo" contentUrl="SafariSample/sheets/ENDANGEREDSAFARI">
```
RussTheAerialist: Why the name change?

PR #243, opened 2017-11-09
Swaggerize - This will be an on-going to track hackathon progress on swaggerizing

RussTheAerialist:
@benlower @t8y8


RussTheAerialist:
This will continue to fail because I haven't checked in the ability to pull the other repro until it's public.

RussTheAerialist:
This was just for informational purposes.  Closing

PR #244, opened 2017-11-16
Enable add to schedule

RussTheAerialist:
Still a couple of todos for error handling which I'll update this pull request for, and of course, I need to write tests, but wanted to get this out or review sooner rather than later since I'm AFK all next week.


tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -110,6 +110,34 @@ def wrapper(self, *args, **kwargs):
     return _decorator
 
 
+def item_must_be_of_type(**params):
```
t8y8: I think this whole thing can be replaced with just an `isinstance` or `issubclass` check in the functions that need to check for type-ness -- that also lets us expand the types more easily in the future.

But I'm also not totally sure why it's necessary. If you just need to know the type to build the xml properly, we had a wrapper type for `Target` which you could borrow here (ScheduableItem or something) or just do type(THING).__name__

tableauserverclient/server/endpoint/schedules_endpoint.py
```diff
@@ -58,3 +63,22 @@ def create(self, schedule_item):
...
         return new_schedule
+
+    @api(version="2.8")
+    @item_must_be_of_type(workbook_or_datasource=[WorkbookItem, DatasourceItem])
+    def add_to_schedule(self, schedule_id, workbook_or_datasource):
+        type_ = item_type(workbook_or_datasource)
+
+        # id will exist because item_must_be_of_type ensures this
+        id_ = workbook_or_datasource.id
+        req_factory = getattr(RequestFactory.Schedule, 'add_{}_req'.format(type_), None)
```
t8y8: Similar comments to above -- I think this can just be a simple if statement on the type (or a mapping if there are lots of types) but that feels more robust than the meta-programming with getattr

tableauserverclient/server/endpoint/endpoint.py
RussTheAerialist: I'm not sure I follow what you mean.  Could you drop by if you have a moment?

tableauserverclient/type_helpers.py
```diff
@@ -0,0 +1,2 @@
+def item_type(obj):
```
t8y8: Looks like this doesn't get used anymore, can just kill it

test/test_schedule.py
```diff
@@ -182,3 +185,33 @@ def test_update(self):
...
+            m.get(self.server.workbooks.baseurl + '/bar', text=workbook_response)
+            m.put(baseurl + '/foo/workbooks', text="OK")
+            workbook = self.server.workbooks.get_by_id("bar")
+            result = self.server.schedules.add_to_schedule('foo', workbook=workbook)
+        self.assert_(result, "Added properly")
+
+    def test_add_datasource(self):
+        self.server.version = "2.8"
+        baseurl = "{}/sites/{}/schedules".format(self.server.baseurl, self.server.site_id)
+        print(baseurl)
```
t8y8: debug print

test/test_schedule.py
```diff
@@ -182,3 +185,33 @@ def test_update(self):
         self.assertEqual(time(7), single_schedule.interval_item.start_time)
         self.assertEqual(("Monday", "Friday"),
                          single_schedule.interval_item.interval)
+
+    def test_add_workbook(self):
+        self.server.version = "2.8"
+        baseurl = "{}/sites/{}/schedules".format(self.server.baseurl, self.server.site_id)
+        print(baseurl)
```
t8y8: debug print

tableauserverclient/server/request_factory.py
```diff
@@ -209,6 +209,29 @@ def update_req(self, schedule_item):
                 single_interval_element.attrib[expression] = value
         return ET.tostring(xml_request)
 
+    def _add_to_req(self, id_, type_):
+        """
+        <task>
```
t8y8: Man we should have done this for all the request factories, it's really nice.

RussTheAerialist: yeah... hindsight...

RussTheAerialist:
I'm planning to take care of the TODOs before submitting, I just wanted to get your feedback on the approach before I went too deep

tableauserverclient/server/endpoint/schedules_endpoint.py
```diff
@@ -71,10 +74,12 @@ def add_to(resource, type_, req_factory):
...
             add_req = req_factory(id_)
-            _ = self.put_request(url, add_req)
-            # TODO: Assert that server_response is 2xx, otherwise, throw an error
+            response = self.put_request(url, add_req)
+            if response.status_code < 200 or response.status_code >= 300:
+                return AddResponse(result=False,
+                                   error="Status {}: {}".format(response.status_code, response.reason))
             logger.info("Added {} to {} to schedule {}".format(type_, id_, schedule_id))
-            return True
+            return OK
```
t8y8: The other `PUT` endoints just return `True` or `None` or something equally meaningless; looks like you want to touch that concept up?

I think the idea of an `OK` is solid -- it won't bother anyone who doesn't care, and it's a little cleaner for those who do. I'm mulling over the real usage of it.. would it be something like:

```
from blah import OK
resp = server.schedules.add_to_schedule(schedule_id, workbook)
if resp is not OK:
    error_stuff()
# OR

resp = server.schedules.add_to_schedule(schedule_id, workbook)
if not resp.is_ok():
    error_stuff()

# OR

resp = server.schedules.add_to_schedule(schedule_id, workbook)
if not resp:  # I don't know how truthy-ness of namedtuples works but we can probably overwrite __bool__ 
    error_stuff()
```

Or something else?

I'm not a big fan of requiring users to import `OK` if they want to check that a response went well, but I agree the OK concept is better than just raw 'True'

tableauserverclient/server/endpoint/schedules_endpoint.py
```diff
@@ -84,5 +89,4 @@ def add_to(resource, type_, req_factory):
             items.append((datasource, "datasource", RequestFactory.Schedule.add_datasource_req))
 
         results = (add_to(*x) for x in items)
-
-        return all(results)
+        return filter(lambda x: not x.result, results)
```
t8y8: Just curious -- why do we have to filter out empty results... or maybe the question is "why does Server return empty results?"

tableauserverclient/server/endpoint/schedules_endpoint.py
```diff
@@ -58,3 +66,28 @@ def create(self, schedule_item):
...
+        items = []
+
+        if workbook is not None:
+            items.append((workbook, "workbook", RequestFactory.Schedule.add_workbook_req))
+        if datasource is not None:
+            items.append((datasource, "datasource", RequestFactory.Schedule.add_datasource_req))
+
+        results = (add_to(*x) for x in items)
+        # list() is needed for python 3.x compatibility
+        return list(filter(lambda x: not x.result, results))
```
t8y8: I thought harder and rewrote me question: Why filter out the 'bad' responses? How would the user know?

RussTheAerialist: 👍

PR #248, opened 2017-12-29
Update api-ref.md

BingoDinkus:
In the workbooks.get() example, the variable was declared as `all_workbook_items` and then referenced as `all_workbooks`. A minor typo that prevents the code example from running.


RussTheAerialist:
@BingoDinkus Thanks for catching this.  We make all of our changes against development, rather than master, so I'm going to close this PR and resubmit it under development.

PR #250, opened 2018-01-15
Reviving Lee's Refresh now

RussTheAerialist:
There has been an outstanding change for refresh now for a while.  This is to finally get this into the code base.


tableauserverclient/models/job_item.py
```diff
@@ -0,0 +1,60 @@
...
+        return "<Job#{_id} {_type} created_at({_created_at}) started_at({_started_at}) completed_at({_completed_at})" \
+               " finish_code({_finish_code})>".format(**self.__dict__)
+
+    @classmethod
+    def from_response(cls, xml, ns):
+        parsed_response = ET.fromstring(xml)
+        all_tasks_xml = parsed_response.findall(
+            './/t:job', namespaces=ns)
+
+        all_tasks = (JobItem._parse_element(x, ns) for x in all_tasks_xml)
```
t8y8: Nit: You can just do a list comprehension since you just convert to a `list`a line later.

Issue #190, opened 2017-05-12
Update Datasource Connection


MouradHRM:
The REST API supports the feature "Update Datasource Connection" but there is not an implementation in server-client-python for that.
And I try to set new connection settings to datasources freshly published.
There is an analog feature for workbooks and I'd like to propose an implementation similar to it.

MouradHRM:
I could not find the test of update workbook connection. I'm not sure how I can implement the test of update datasource connection

RussTheAerialist:
#253  adds this ability and unifies the api around this.

bheem123:
I tested the update_connection.py on my server version 2018.2, and it does not work. No error,  but nothing is updated on server. Did this work for anyone?

tdelosh:
I also can run the code, but nothing is updated. Does anyone have information on this?

tdelosh:
Figured this out. Finally. This does nothing (paraphrasing):

wb.connections[i].username = 'new_username'
wb.connections[i].password = 'new_password'

Because the connection_items fetched by server.workbooks.populate_connections() are not writable.

I know next to nothing about Python, but this works:

from copy import deepcopy
# as you're iterating though the connections using index var i:
conn_copy = deepcopy(wb.connections[i])
conn_copy.username = 'dracula'
conn_copy.password = 'mypwd'
conn_copy.embed_password = True
server.workbooks.update_connection(wb, conn_copy)

Not sure a deep copy is necessary, but I'm not going to mess with it.

PR #253, opened 2018-01-17
#190 add update datasource connection

RussTheAerialist:
Clean up the API a bit because I didn't really click with it.

### DIFF
--- a/samples/update_connection.py
+++ b/samples/update_connection.py
@@ -50,7 +50,7 @@ def main():
         update_function = endpoint.update_connection
         resource = endpoint.get_by_id(args.resource_id)
         endpoint.populate_connections(resource)
-        connections = filter(lambda x: x.id == args.connection_id, resource.connections)
+        connections = list(filter(lambda x: x.id == args.connection_id, resource.connections))
         assert(len(connections) == 1)
         connection = connections[0]
         connection.username = args.datasource_username
--- a/samples/update_connection.py
+++ b/samples/update_connection.py
@@ -12,7 +12,7 @@ import tableauserverclient as TSC
 
 
 def main():
-    parser = argparse.ArgumentParser(description='Get all of the refresh tasks available on a server')
+    parser = argparse.ArgumentParser(description='Update a connection on a datasource or workbook to embed credentials')
     parser.add_argument('--server', '-s', required=True, help='server address')
     parser.add_argument('--username', '-u', required=True, help='username to sign into server')
     parser.add_argument('--site', '-S', default=None)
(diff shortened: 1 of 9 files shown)


samples/update_connection.py
```diff
@@ -0,0 +1,63 @@
...
+
+import argparse
+import getpass
+import logging
+
+import tableauserverclient as TSC
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Get all of the refresh tasks available on a server')
```
t8y8: The description is for the wrong sample (I bet a lot of our samples are, actually)

samples/update_connection.py
```diff
@@ -0,0 +1,63 @@
...
+        endpoint = {
+            'workbook': server.workbooks,
+            'datasource': server.datasources
+        }.get(args.resource_type)
+
+        update_function = endpoint.update_connection
+        resource = endpoint.get_by_id(args.resource_id)
+        endpoint.populate_connections(resource)
+        connections = filter(lambda x: x.id == args.connection_id, resource.connections)
+        assert(len(connections) == 1)
```
t8y8: Do you have to `list(filter` Here for python3? I forget if filter_results objects have len

tableauserverclient/models/connection_item.py
```diff
@@ -29,6 +29,10 @@ def id(self):
     def connection_type(self):
         return self._connection_type
 
+    def __repr__(self):
+        return "<ConnectionItem#{_id} embed={embed_password} type={_connection_type} username={username}>"\
```
t8y8: One day we can use f-strings. One day...

tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -87,11 +87,16 @@ def update(self, workbook_item):
         return updated_workbook._parse_common_tags(server_response.content, self.parent_srv.namespace)
 
     # Update workbook_connection
-    def update_conn(self, workbook_item, connection_item):
+    @api(version="2.3")
+    def update_connection(self, workbook_item, connection_item):
```
t8y8: Should probably set `update_conn` to `update_connection` for back compact; and throw a warning?

t8y8:
I like it!

We may want to let the guy who was working on adding multiple connections know (did that ever get done? I should pester Ben...)

samples/update_connection.py
RussTheAerialist: Doh. Yes.

tableauserverclient/server/endpoint/workbooks_endpoint.py
RussTheAerialist: Probably... I'll make the change.

bheem123:
I tested the update_connection.py on my server version 2018.2, and it does not work. No error,  but nothing is updated on server. Did this work for anyone?

PR #263, opened 2018-02-14
Export sample

RussTheAerialist:
Add a sample/tool for using export from the rest api.  It also supports view filters.

### DIFF
--- a/samples/export.py
+++ b/samples/export.py
@@ -6,7 +6,7 @@ import tableauserverclient as TSC
 
 
 def main():
-    parser = argparse.ArgumentParser(description='Get all of the refresh tasks available on a server')
+    parser = argparse.ArgumentParser(description='Export a view as an image, pdf, or csv')
     parser.add_argument('--server', '-s', required=True, help='server address')
     parser.add_argument('--username', '-u', required=True, help='username to sign into server')
     parser.add_argument('--site', '-S', default=None)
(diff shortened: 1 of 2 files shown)


samples/export.py
```diff
@@ -0,0 +1,69 @@
+import argparse
+import getpass
+import logging
+
+import tableauserverclient as TSC
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Get all of the refresh tasks available on a server')
```
shinchris: Description here and for list.py is for refresh tasks.

samples/export.py
```diff
@@ -0,0 +1,69 @@
...
+    logging_level = getattr(logging, args.logging_level.upper())
+    logging.basicConfig(level=logging_level)
+
+    tableau_auth = TSC.TableauAuth(args.username, password, args.site)
+    server = TSC.Server(args.server, use_server_version=True)
+    with server.auth.sign_in(tableau_auth):
+        views = filter(lambda x: x.id == args.resource_id,
+                       TSC.Pager(server.views.get))
+        view = views.pop()
+        (populate_name, option_factory_name, member_name, extension) = args.type
```
t8y8: It took me a few reads to understand what's going on here. 
I think the code is fine, but some suggestions for clarity:

- Rename 'populate_name' to 'populate_func_name'
- Add a one line comment above that says something like "We've got lots of types and classes, this is code to figure out all the right functions to call" or something

samples/export.py
```diff
@@ -0,0 +1,69 @@
...
+        password = args.p
+
+    # Set logging level based on user input, or error by default
+    logging_level = getattr(logging, args.logging_level.upper())
+    logging.basicConfig(level=logging_level)
+
+    tableau_auth = TSC.TableauAuth(args.username, password, args.site)
+    server = TSC.Server(args.server, use_server_version=True)
+    with server.auth.sign_in(tableau_auth):
+        views = filter(lambda x: x.id == args.resource_id,
```
t8y8: Code is fine, but why use a pager with filter instead of the native filtering in request options?

Can we not filter on ID yet?

RussTheAerialist: I'm not sure views supports that right now.

PR #274, opened 2018-03-09
Add a simple static method to strip non-XML responses from debug log

### LINKED ISSUE #273: This log line is overly chatty
https://github.com/tableau/server-client-python/blob/608aa7694d0560ea3c8c37b10127b11207e56e8d/tableauserverclient/server/endpoint/endpoint.py#L53

When using server client python to download workbooks or data sources and you've got log_level=Debug, this log line ends up blowing up your logs. It outputs the hexadecimal representation of the entire file you're downloading, which is not very helpful and explodes your log size. Can we remove this line, or only log out the response contents when you're not using the endpoint to download a file?

t8y8:
We could add other types (like JSON) in the future, or if I missed any in the CR I can fix them up, it's a simple fix.

The more complicated approaches, that I abandoned, were to try and grab some chunk of the file and find a safe way to encode it to utf8 for either file or stdout streams -- but that didn't seem worth it :)

Fixes #273

### DIFF
--- a/samples/download_view_image.py
+++ b/samples/download_view_image.py
@@ -43,7 +43,7 @@ def main():
     tableau_auth = TSC.TableauAuth(args.username, password, site_id=site_id)
     server = TSC.Server(args.server)
     # The new endpoint was introduced in Version 2.5
-    server.version = 2.5
+    server.version = "2.5"
 
     with server.auth.sign_in(tableau_auth):
         # Step 2: Query for the view that we want an image of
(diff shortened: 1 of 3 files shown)


samples/download_view_image.py
```diff
@@ -43,7 +43,7 @@ def main():
     tableau_auth = TSC.TableauAuth(args.username, password, site_id=site_id)
     server = TSC.Server(args.server)
     # The new endpoint was introduced in Version 2.5
-    server.version = 2.5
+    server.version = "2.5"
```
t8y8: I ran this as a test; turns out there was a small bug in that this needs to be a string.

We can file a new issue on making that a `@property.setter` and more robust to that kind of error

shinchris:
:rocket:

PR #276, opened 2018-03-10
Multi-Credential Support in TSC

t8y8:
Following on from @marianotn's work, I rebased on top of the latest Development and reorganized a few things for readability and back-compat.

Still need to update tests, I'm probably going to test `RequestFactory.Workbook.publish_req` specifically, which would be one of the few serializer tests, but I think it's worth the coverage.

Thoughts on the approach? I tried to make it so connections would work older than 2.8, but it was getting really nasty.

I also need to add support to Data Sources too.

### DIFF
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -32,11 +32,12 @@ class Endpoint(object):
         '''Checks if the server_response content is not xml (eg binary image or zip)
         and and replaces it with a constant
         '''
-        ALLOWED_CONTENT_TYPES = ('application/xml',)
+        ALLOWED_CONTENT_TYPES = ('application/xml', 'application/xml;charset=utf-8')
         if server_response.headers.get('Content-Type', None) not in ALLOWED_CONTENT_TYPES:
+            print(server_response.headers)
             return '[Truncated File Contents]'
         else:
-            return server_response.content
+            return server_response.content[:300]
 
     def _make_request(self, method, url, content=None, request_object=None,
                       auth_token=None, content_type=None, parameters=None):
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -34,7 +34,6 @@ class Endpoint(object):
         '''
         ALLOWED_CONTENT_TYPES = ('application/xml', 'application/xml;charset=utf-8')
         if server_response.headers.get('Content-Type', None) not in ALLOWED_CONTENT_TYPES:
-            print(server_response.headers)
             return '[Truncated File Contents]'
         else:
             return server_response.content[:300]
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -36,7 +36,7 @@ class Endpoint(object):
         if server_response.headers.get('Content-Type', None) not in ALLOWED_CONTENT_TYPES:
             return '[Truncated File Contents]'
         else:
-            return server_response.content[:300]
+            return server_response.content
 
     def _make_request(self, method, url, content=None, request_object=None,
                       auth_token=None, content_type=None, parameters=None):
--- a/tableauserverclient/server/endpoint/workbooks_endpoint.py
+++ b/tableauserverclient/server/endpoint/workbooks_endpoint.py
@@ -204,8 +204,7 @@ class Workbooks(Endpoint):
 
         if connection_credentials is not None:
             import warnings
-            warnings.warn("connection_credentials is being deprecated. Use connections instead, " + \
-            "see https://onlinehelp.tableau.com/current/api/rest_api/en-us/help.htm#REST/rest_api_ref.htm#Publish_Workbook",
+            warnings.warn("connection_credentials is being deprecated. Use connections instead",
                           DeprecationWarning)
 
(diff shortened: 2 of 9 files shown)


tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -32,11 +32,11 @@ def _safe_to_log(server_response):
         '''Checks if the server_response content is not xml (eg binary image or zip)
         and and replaces it with a constant
         '''
-        ALLOWED_CONTENT_TYPES = ('application/xml',)
+        ALLOWED_CONTENT_TYPES = ('application/xml', 'application/xml;charset=utf-8')
```
t8y8: I should allow this type through as well, missed it in first PR

tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -32,11 +32,11 @@ def _safe_to_log(server_response):
...
         if server_response.headers.get('Content-Type', None) not in ALLOWED_CONTENT_TYPES:
             return '[Truncated File Contents]'
         else:
-            return server_response.content
+            return server_response.content[:300]
```
t8y8: Will remove this.

tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -231,15 +238,18 @@ def publish(self, workbook_item, file_path, mode, connection_credentials=None):
             upload_session_id = Fileuploads.upload_chunks(self.parent_srv, file_path)
             url = "{0}&uploadSessionId={1}".format(url, upload_session_id)
             xml_request, content_type = RequestFactory.Workbook.publish_req_chunked(workbook_item,
```
t8y8: I need to pipe this through to 'chunked' code paths as well

tableauserverclient/server/request_factory.py
```diff
@@ -313,24 +313,44 @@ def add_req(self, user_item):
...
-            credentials_element.attrib['name'] = connection_credentials.name
-            credentials_element.attrib['password'] = connection_credentials.password
-            credentials_element.attrib['embed'] = 'true' if connection_credentials.embed else 'false'
 
-            if connection_credentials.oauth:
-                credentials_element.attrib['oAuth'] = 'true'
+        if connection_credentials is not None and connections is not None:
+            raise RuntimeError('You cannot set both `connections` and `connection_credentials`')
+
+            if connection_credentials is not None:
```
t8y8: This is redundant, will remove

tableauserverclient/server/request_factory.py
```diff
@@ -313,24 +313,44 @@ def add_req(self, user_item):
...
+                self._add_credentials_element(workbook_element, connection_credentials)
+
+        if connections is not None:
+            connections_element = ET.SubElement(workbook_element, 'connections')
+            for connection in connections:
+                self._add_connections_element(connections_element, connection)
         return ET.tostring(xml_request)
 
+    def _add_connections_element(self, connections_element, connection):
```
t8y8: Can make these static, or move to a helper class for datasources to share

t8y8:
The rabbit hole goes deeper -- I'm adding parsers so we can extract returned values from requests... at least for testing.

t8y8:
Alright, I'm ready to declare this ready for review.

But I realized something: We don't support multi-credential Data Source publish, so you can't publish a federated data source that requires multiple credentials. Only workbooks.

tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -230,16 +237,21 @@ def publish(self, workbook_item, file_path, mode, connection_credentials=None):
             logger.info('Publishing {0} to server with chunking method (workbook over 64MB)'.format(filename))
             upload_session_id = Fileuploads.upload_chunks(self.parent_srv, file_path)
             url = "{0}&uploadSessionId={1}".format(url, upload_session_id)
+            conn_creds = connection_credentials
```
shinchris: Is conn_creds variable necessary?

tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -199,7 +199,14 @@ def _get_wb_preview_image(self, workbook_item):
 
     # Publishes workbook. Chunking method if file over 64MB
     @api(version="2.0")
-    def publish(self, workbook_item, file_path, mode, connection_credentials=None):
+    @parameter_added_in(connections='2.8')
+    def publish(self, workbook_item, file_path, mode, connection_credentials=None, connections=None):
+
+        if connection_credentials is not None:
+            import warnings
+            warnings.warn("connection_credendials is being deprecated. Use connections instead, see http://...",
```
shinchris: Typo in connection_credendials. Also I'm not too familiar with connection_credentials but should this warning also be in datasources_endpoint.py?

tableauserverclient/server/endpoint/workbooks_endpoint.py
t8y8: Only because of line-length limits 🤷‍♂️

tableauserverclient/server/endpoint/workbooks_endpoint.py
t8y8: If this were supported in Datasources then yes, but it turns out my code is ahead of the curve here, datasources don’t do multi-credential publishing yet.

PR #279, opened 2018-03-23
277 update group feature

sotnich:
Adding support for update method for groups.

### DIFF
--- a/tableauserverclient/server/endpoint/groups_endpoint.py
+++ b/tableauserverclient/server/endpoint/groups_endpoint.py
@@ -55,6 +55,13 @@ class Groups(Endpoint):
         self.delete_request(url)
         logger.info('Deleted single group (ID: {0})'.format(group_id))
 
+    @api(version="2.0")
+    def update(self, group_item, default_site_role=UserItem.Roles.Unlicensed):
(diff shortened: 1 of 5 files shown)


tableauserverclient/server/endpoint/groups_endpoint.py
```diff
@@ -55,6 +55,16 @@ def delete(self, group_id):
         self.delete_request(url)
         logger.info('Deleted single group (ID: {0})'.format(group_id))
 
+    @api(version="2.0")
+    def update(self, group_item, default_site_role=UserItem.Roles.Unlicensed):
```
t8y8: Pull `UserItem.Roles.Unlicensed` into a constant `UNLICENSED_USER` or something a little shorter, it helps this read a little easier for me

t8y8:
There's a linter error in the tests, but it looks good otherwise, once it's passing checks I'll approve. I'll ping the team to have another set of eyes take a look as well.

t8y8:
I will merge this PR, but we realized that docs changes need to go to the `gh-pages` branch directly now. Can you open a PR there? @d45 and @RussTheAerialist can help if you have questions there.

We'll be removing the docs from this branch to make it less confusing in the future :)

PR #291, opened 2018-05-08
Adding all of the new site roles

RussTheAerialist:
This is against master so I can release a 0.6.2 version


tableauserverclient/models/user_item.py
```diff
@@ -16,6 +16,15 @@ class Roles:
...
         Guest = 'Guest'
 
+        Creator = 'Creator'
+        Explorer = 'Explorer'
+        ExplorerCanPublish = 'ExplorerCanPublish'
+        ReadOnly = 'ReadOnly'
+        SiteAdministratorCreator = 'SiteAdministratorCreator'
+        SiteAdministratorExplorer = 'SiteAdministratorExplorer'
+        UnlicensedWithPublish = 'UnlicensedWithPublish'
+        ViewerWithPublish = 'ViewerWithPublish'
```
shinchris: ViewerWithPublish already exists on line 16.

PR #298, opened 2018-05-29
initial checkin for get background jobs

RussTheAerialist:
New Endpoint from the scheduling team.  First pass for the first endpoint

### DIFF
--- a/tableauserverclient/models/job_item.py
+++ b/tableauserverclient/models/job_item.py
@@ -1,4 +1,5 @@
 import xml.etree.ElementTree as ET
+from ..datetime_helpers import parse_datetime
 from .target import Target
 
 
@@ -86,7 +87,8 @@ class BackgroundJobItem(object):
 
     @property
     def name(self):
-        """For APi consistency"""
+        """For API consistency - all other resource endpoints have a name attribute which is used to display what
+        they are.  Alias title as name to allow consistent handling of resources in the list sample."""
         return self._title
 
     @property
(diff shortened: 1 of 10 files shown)


tableauserverclient/models/job_item.py
```diff
@@ -58,3 +58,86 @@ def _parse_element(cls, element, ns):
...
+        self._title = title
+        self._subtitle = subtitle
+
+    @property
+    def id(self):
+        return self._id
+
+    @property
+    def name(self):
+        """For APi consistency"""
```
t8y8: This docstring can probably go, or expand on what "API consistency" means

tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -29,9 +29,9 @@ def _make_common_headers(auth_token, content_type):
 
     @staticmethod
     def _safe_to_log(server_response):
-        '''Checks if the server_response content is not xml (eg binary image or zip)
+        """Checks if the server_response content is not xml (eg binary image or zip)
         and and replaces it with a constant
```
t8y8: Since you're tweaking my comment, can you fix my double 'and and' while you're there :)

tableauserverclient/server/server.py
```diff
@@ -94,6 +100,14 @@ def use_highest_version(self):
         import warnings
         warnings.warn("use use_server_version instead", DeprecationWarning)
 
+    def assert_at_least_version(self, version):
```
t8y8: Why move this to the server object itself?

tableauserverclient/server/endpoint/jobs_endpoint.py
```diff
@@ -11,7 +11,20 @@ def baseurl(self):
...
 
     @api(version='2.6')
-    def get(self, job_id):
+    def get(self, job_id=None, req_options=None):
+        # Backwards Compatibility fix until we rev the major version
+        if job_id is not None and isinstance(job_id, basestring):
+            import warnings
+            warnings.warn("Jobs.get(job_id) is deprecated, update code to use Jobs.get_by_id(job_id)")
+            return self.get_by_id(job_id)
+        self.parent_srv.assert_at_least_version('3.1')
```
t8y8: I see now, you need to call this assert to check that it's > 3.1 and also keep the > 2.6 for old back compat. I need to think about this, maybe it's better to have another decorator behavior_changed_in... or something like that. 

But this seems fine for now

t8y8:
🚀

PR #300, opened 2018-05-31
D45 docs v06.1

d45:
Hi, 
The biggest part of this update is to api-ref.md. I did a diff between the v4.1 release and the v6.1 and then tried to add every significant change.  I left out some things, for example, the `target` class isn't listed separately. I just lumped it under subscriptions (as the target member).   

I'd like to get this PR merged to gh-pages, so that we can make smaller, faster, incremental updates. Once a year is not often enough!


docs/dev-guide.md
```diff
@@ -36,45 +43,47 @@ This topic describes how to contribute to the Tableau Server Client (Python) pro
...
+### Code and commit
 
-   Here's a quick checklist for ensuring a good pull request:
+   Here's a quick checklist to follow when coding to ensure a good pull request (PR):
 
-   - Only touch the minimal amount of files possible while still accomplishing the goal.
+   - Only touch the fewest number of files possible while still accomplishing the goal.
    - Ensure all indentation is done as 4-spaces and your editor is set to unix line endings.
-   - The code matches PEP8 style guides. If you cloned the repo you can run `pycodestyle .`
+   - The code matches PEP8 style guides. If you cloned the repo you can run `pycodestyle server-client-python test samples`
```
RussTheAerialist: this should be `pycodestyle tableauserverclient test samples`

docs/versions.md
```diff
@@ -21,26 +21,34 @@ server = TSC.Server('http://SERVER_URL')
...
-## Use another version of the REST API
+## Use the REST API version supported by the server
 
-To use another version of the REST API, set the version like so:
+To always use the latest version of the REST API that is supported by the instance of Tableau Server you are connecting to, set the version using the `use_server_version()` method:
 
 ```py
 
```
RussTheAerialist: probably have it be `server = TSC.Server('http://SERVER_URL', use_server_version=True)` since that's more compact

docs/versions.md
```diff
@@ -21,26 +21,34 @@ server = TSC.Server('http://SERVER_URL')
...
+```py
+import tableauserverclient as TSC
+
+server = TSC.Server('http://SERVER_URL')
+
+ 
+server.version = '2.6'
 
 
->>>>>>> 87acd20... Update versions.md (#195)
```
RussTheAerialist: There is a conflict here.  You need to resolve that before we push this.

docs/dev-guide.md
d45: Updated the `pycodestyle` command as suggested.

docs/versions.md
d45: Yes. That is pretty slick.

d45:
I resolved the merge conflicts.

docs/versions.md
d45: Resolved.

PR #305, opened 2018-06-18
 Adding reference to new REST API version 3.0 

marcomix54:
On server.py

Adding to "_PRODUCT_TO_REST_VERSION" the new REST API version.
Changing in "def __init__(self, server_address, use_server_version=False):" 
Changing in "def _determine_highest_version(self):"

### DIFF
--- a/tableauserverclient/server/server.py
+++ b/tableauserverclient/server/server.py
@@ -31,7 +31,7 @@ class Server(object):
         self._session = requests.Session()
         self._http_options = dict()
 
-        self.version = "3.0"
+        self.version = "2.3"
         self.auth = Auth(self)
         self.views = Views(self)
         self.users = Users(self)
@@ -77,7 +77,7 @@ class Server(object):
     def _determine_highest_version(self):
         try:
             old_version = self.version
-            self.version = "3.0"
+            self.version = "2.4"
             version = self.server_info.get().rest_api_version
         except ServerInfoEndpointNotFoundError:
             version = self._get_legacy_version()


tableauserverclient/server/server.py
```diff
@@ -76,7 +77,7 @@ def _get_legacy_version(self):
     def _determine_highest_version(self):
         try:
             old_version = self.version
-            self.version = "2.4"
+            self.version = "3.0"
```
RussTheAerialist: Don't change this one.  The endpoint we are using to check the highest version was introduced in 2.4, so we need to force it to use that version to cover old versions.

tableauserverclient/server/server.py
```diff
@@ -30,7 +31,7 @@ def __init__(self, server_address, use_server_version=False):
         self._session = requests.Session()
         self._http_options = dict()
 
-        self.version = "2.3"
+        self.version = "3.0"
```
RussTheAerialist: Don't change this one either.  2.3 is the older version we support, so we want to default to the oldest version unless specifically detected to be earlier.

marcomix54: Hi Russ,

So if I remove this change and initialise the server with use_server_version = True. then it works. Is that how I'm supposed to do it?

Do we expect people to know that or is it documented?
Marc

t8y8:
I would support a comment in the code that says "this doesn't need to be updated" because of that

pmalafosse:
With Tableau Online 2018.2, it tries to use the API version '2.3' that is set by default in server.py what causes many functions to fail.

irwando:
As mentioned we changed how version numbers were accessed, now going through the getServerInfo call.  No need for these updates and no action in a year.  Closing this.

PR #313, opened 2018-07-06
Merge multicredential and as job

RussTheAerialist:
There was a divergence that happened when something was checked into master rather than development.  This is to try to merge it into development so that I can get development and master back in sync.


gaoang2148:
Thank you @RussTheAerialist ! LGTM!

PR #316, opened 2018-07-10
Merge in v0.7 from development.

RussTheAerialist:
This required a bit of work because of the as_job getting accidentally checked into master.


t8y8:
Whoo-wee, that's a big PR.

We've worked on a lot!

Issue #320, opened 2018-07-18
Inconsistent list comprehension in Tasks


happyshows:
```
#working
workbooks, pagination_item = server.workbooks.get()
[workbooks for workbooks  in TSC.Pager(server.workbooks)]
# not working
tasks, pagination_item = server.tasks.get()
[task for task  in TSC.Pager(server.tasks)]
```

See Error
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-66-391b125b936f> in <module>()
----> 1 [task for task  in TSC.Pager(server.tasks)]

<ipython-input-66-391b125b936f> in <listcomp>(.0)
----> 1 [task for task  in TSC.Pager(server.tasks)]

/usr/local/lib/python3.6/site-packages/tableauserverclient/server/pager.py in __iter__(self)
     38 
     39         # Get the rest on demand as a generator
---> 40         while self._count < last_pagination_item.total_available:
     41             if len(current_item_list) == 0:
     42                 current_item_list, last_pagination_item = self._load_next_page(last_pagination_item)

TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

t8y8:
This looks to be the same as #294

t8y8:
We could also have Pager guard against failure, and for something that doesn't return a total_available we just stop after the first 'page'

https://github.com/tableau/server-client-python/blob/master/tableauserverclient/server/pager.py#L38

t8y8:
Scratch that, I need to dig deeper to keep the semantics of the iterator and not blow up at the wrong time

t8y8:
```py
        if last_pagination_item.total_available is None:
            # This endpoint does not support pagination, drain the list and return
            while current_item_list:
                yield current_item_list.pop(0)
            
            return
```

Is the real code that fixes the bug

happyshows:
@t8y8 I'm new to python, how do I debug it?

t8y8:
@happyshows Ah, no worries; I actually pushed the fix to the development branch, you can try out out by pip installing from the github directly:
```
pip install https://github.com/tableau/server-client-python.git@development
```

Then rerun your script (remember to delete it after :) )

irwando:
Looks like @t8y8 's fix has been pushed to release.  I see it in pager.py.  CLosing this.

PR #322, opened 2018-07-19
Fix #294 & #320

### LINKED ISSUE #294: tasks_endpoint attempts to use pagination
First time filing an issue on GitHub, so I apologize if it's formatted incorrectly (or just wrong)...

I am attempting to create functions that will get the associated resources. In most cases, I would use something along the lines of:

```
def get_schedules():
    for schedule in TSC.Pager(server.schedules):
        print("\nSchedule Name: {}\nSchedule ID: {}".format(schedule.name, schedule.id)
```

However, the 'Get Extract Refresh Tasks API call does not support pagination. As a result, I've created my function like so:

```
def get_refresh_tasks():
    for task in server.tasks.get():
        task_list = "\nTask ID: {}\nSchedule ID: {}\nRefresh Priority: {}\nExtract type: {}\nExtract ID: {}"
        print(task_list.format(task.id, task.schedule_id, task.priority, task.target.type, task.target.id))
```
This results in a AttributeError: 'list' object has no attribute 'id'. But I found that by commenting out two parts of the modified the tasks_endpoint.py file, my code worked entirely as expected:
```
    def get(self, req_options=None):
        logger.info('Querying all tasks for the site')
        url = self.baseurl
        server_response = self.get_request(url, req_options)
        #pagination_item = PaginationItem.from_response(server_response.content, self.parent_srv.namespace)
        all_extract_tasks = TaskItem.from_response(server_response.content, self.parent_srv.namespace)
        return all_extract_tasks#, pagination_item

```

As an alternative workaround, the following function works just fine AS-IS:

```
def get_refresh_tasks():
    tasks = server.tasks.get()
    for task in tasks[0]:
        task_list = "\nTask ID: {}\nSchedule ID: {}\nRefresh Priority: {}\nExtract type: {}\nExtract ID: {}"
        print(task_list.format(task.id, task.schedule_id, task.priority, task.target.type, task.target.id))

```
Am I correct in assuming the tasks_endpoint file should not return pagination_item since the Get Extracts Tasks API call does not support pagination? Or should I be writing the the get_refresh_tasks function like the last code example

t8y8:
We should update tasks to be paginated, but `Pager` can be more defensive, and should now work against any endpoint

### DIFF
--- a/tableauserverclient/server/pager.py
+++ b/tableauserverclient/server/pager.py
@@ -36,6 +36,13 @@ class Pager(object):
         # Fetch the first page
         current_item_list, last_pagination_item = self._endpoint(self._options)
 
+        if last_pagination_item.total_available is None:
+            # This endpoint does not support pagination, drain the list and return
+            while current_item_list:
+                yield current_item_list.pop(0)
+
+            return
+
         # Get the rest on demand as a generator
         while self._count < last_pagination_item.total_available:
             if len(current_item_list) == 0:


t8y8:
Since I manually tested the repro and fix...

Before:
image

After:
image

PR #341, opened 2018-09-27
added possibility to export an image of a view that is up to date

daniel1608:
- tableau caches images of views for some time (720 min by default)
- image exports come from cache if image was cached
- image might be outdated because of cache
- not for all people possible to change cache time


irwando:
Ancient pull requests, and while updating the client is noce the server side doesn't support this. Please open an enhancement request issue if this is necessary.

PR #345, opened 2018-10-02
fix link pointing to old doc content on master branch

mbren:
The link points to the old version of the dev-guide on the master branch. Point link to the new version on the gh-pages branch by linking to the built docs instead of repo location.


d45:
@mbren Hi Mary, I know this is a minor fix, but I think you need to make the fix to the development branch first. The development branch should always have the latest/greatest.

mbren:
Closing this PR without merging and made the change in the development branch instead. Thanks for the feedback.

Issue #351, opened 2018-10-23
basestring is no longer supported in Python 3


julien10:
Hi guys,

Here is an extract of Python 3.0's changelog:
> The builtin basestring abstract type was removed. Use str instead. The str and bytes types don’t have functionality enough in common to warrant a shared base class. The 2to3 tool (see below) replaces every occurrence of basestring with str.

Could you please take care of this issue ?

https://github.com/tableau/server-client-python/blob/b7b544e5ef405ca76f9e0c5681b001373811c6b2/tableauserverclient/server/endpoint/jobs_endpoint.py#L17

shinchris:
Thanks for pointing this out.

Merged into development branch: #362

PR #356, opened 2018-11-09
fixing usage of basestring in job_endpoint, #351

shinchris:
Addressing #351


shinchris:
I will fix the commit history

PR #359, opened 2018-11-14
Add maxage param to the download view image request

preguraman:


preguraman:
Added the MaxAge query param to the Query View Image Rest API

RussTheAerialist:
Please rebase against development and not master since we do all development to the development branch and then merge into master when we release.

PR #362, opened 2018-11-14
adding basestring conversion to str for python3

shinchris:
Addressing #351 where the usage of `basestring` would cause errors in python 3

### DIFF
--- a/tableauserverclient/server/endpoint/jobs_endpoint.py
+++ b/tableauserverclient/server/endpoint/jobs_endpoint.py
@@ -3,6 +3,12 @@ from .. import JobItem, BackgroundJobItem, PaginationItem
 from ..request_options import RequestOptionsBase
 import logging
 
+try:
+    basestring
+except NameError:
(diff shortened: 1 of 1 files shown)


thagraybush:
I downloaded TSC via pip only a few weeks ago and did NOT come with this fix for python 3. Still errors out with basestring issue you fixed @shinchris. Had come up with same and came here to create pull request to add but glad to see fixed. Why is this fix not coming when download latest via pip?

shinchris:
@thagraybush This fix was merged into the 'development' branch, waiting to be released into 'master' with version 0.8. When you downloaded TSC via pip a few weeks ago, we were still on version 0.7, so it didn't include this fix. We released version 0.8 today (4/8/19), so you should be able to pip install now and have this fix applied.

thagraybush:
Thanks! @shinchris I literally added exact same code to my files before coming here, was just going to suggest same change but saw you already fixed it :) So I am good to go and glad getting released to everyone.

PR #371, opened 2018-12-21
Change default server API version to 2.4

aboisvert-playq:
On our Tableau server (version 2018.2.1 / API version 3.1) the original code thinks the server version is 2.3 because it can't run line 86. Setting the default version to 2.4 fixes the issue and the code can now correctly identify the server version.

This is an extremely minor change that should not negatively affect anyone but helps situations like ours.


aboisvert-playq:
This didn't pass the tests and I believe I have a workaround anyway.

PR #375, opened 2019-01-16
Query acceleration

baixin137:
1. Added materialized views settings to workbook/site item class
2. Added a sample python script to enable materialized views for workbooks/sites


samples/materialize_workbooks.py
```diff
@@ -0,0 +1,106 @@
...
+                print "Updated materialized views settings for site:", site_to_update.name
+            print
+
+        elif args.type == 'workbook':
+            tableau_auth = TSC.TableauAuth(args.username, password, site_id=site_content_url)
+            server = TSC.Server(args.server)
+
+            # Now it updates all the workbooks in the site
+            # To update selected ones pass filter into 'workbooks = server.workbooks.get()[0]'
+            # This only updates the workbooks in the site you are signing into
```
guodah: Maybe here we can add a link to https://github.com/tableau/server-client-python/blob/master/docs/docs/filter-sort.md so that whoever wants to use filters can go there easily.

PR #378, opened 2019-01-18
Add materialized views settings support for workbooks/sites

baixin137:
1. Added materialized views settings to workbook/site item class
2. Added a sample python script to enable materialized views for workbooks/sites

### DIFF
--- a/samples/materialize_workbooks.py
+++ b/samples/materialize_workbooks.py
@@ -21,7 +21,8 @@ def main():
     parser.add_argument('--type', '-t', required=False, choices=['site', 'workbook', 'project_name',
                                                                  'project_id', 'project_path'],
                         help='type of content you want to update materialized views settings on')
-    parser.add_argument('--file-path', '-fp', required=False, help='path to a list of workbooks')
+    parser.add_argument('--path-list', '-pl', required=False, help='path to a list of workbook paths')
+    parser.add_argument('--name-list', '-nl', required=False, help='path to a list of workbook names')
     parser.add_argument('--project-name', '-pn', required=False, help='name of the project')
     parser.add_argument('--project-path', '-pp', required=False, help="path of the project")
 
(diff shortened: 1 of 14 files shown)


samples/materialize_workbooks.py
```diff
@@ -60,45 +94,121 @@ def show_materialized_views_status(args, password, site_content_url):
...
+
+
+def update_workbooks_by_paths(all_projects, enable_materialized_views, server, workbook_path_mapping):
+    for workbook_name, workbook_paths in workbook_path_mapping.items():
+        req_option = TSC.RequestOptions()
+        req_option.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
+                                         TSC.RequestOptions.Operator.Equals,
+                                         workbook_name))
+        workbooks = list(TSC.Pager(server.workbooks, req_option))
+        if len(workbooks) == 1:
```
xinsui: even for only one workbook is there, if the project path is given, we may still want to verify the path matches the workbook name given

tableauserverclient/server/endpoint/sites_endpoint.py
```diff
@@ -44,6 +44,17 @@ def get_by_name(self, site_name):
         server_response = self.get_request(url)
         return SiteItem.from_response(server_response.content, self.parent_srv.namespace)[0]
 
+    # Gets 1 site by content url
+    @api(version="3.3")
```
shinchris: I think this endpoint was available since v2.0, just not included in this library.

test/test_site.py
```diff
@@ -91,7 +91,8 @@ def test_update(self):
             single_site = TSC.SiteItem(name='Tableau', content_url='tableau',
                                        admin_mode=TSC.SiteItem.AdminMode.ContentAndUsers,
                                        user_quota=15, storage_quota=1000,
-                                       disable_subscriptions=True, revision_history_enabled=False)
+                                       disable_subscriptions=True, revision_history_enabled=False,
+                                       materialized_views_enabled=False)
```
shinchris: Could you add 'materializedViewsEnabled' to the server response xml (test/assets/site_update.xml) and verify that we are parsing the response properly?
Also would be nice to have test_workbook.py updated, since the new field is added to workbook_item.

Issue #389, opened 2019-01-25
Fix update endpoint for views


shinchris:
The update endpoint for views manages its tags (add/delete). It currently doesn't support deleting tags properly.

Issue #388, opened 2019-01-25
Add views endpoint for 'get_view_by_id'


shinchris:
This endpoint was added in the REST API for version 3.1.

PR #390, opened 2019-01-25
View docs

shinchris:
Docs on views were a bit outdated. Updated fields and added more methods. Also added information about view filters for CSVRequestOptions, ImageRequestOptions, and PDFRequestOptions.

Found some issues with the code itself while updating docs:
- No support for 'get_view_by_id', added version 3.1 (#388)
- The 'update' endpoint (add/delete tags) does not delete properly (#389)

Once the above issues are resolved, I will update the docs for them.


d45:
Yes! Thanks, Chris!  Looks good.

Issue #402, opened 2019-02-20
Support for 2019.1


nishantnath:
As per https://tableau.github.io/server-client-python/docs/versions
Highest Supported Version is 2018.3 while 2019.1 was released Feb 15, 2019

Using a server with 2019.1 and existing TSC python-client, TSC.Pager(server.schedules) throws an error:  `ValueError: Invalid value: Flow. schedule_type must be of type Type`

TSC.Pager(server.workbooks) & TSC.Pager(server.datasources) seems to be working fine.

shinchris:
Thanks for opening an issue for this. This has been fixed in the 'development' branch with #403.

Info on how to use the 'development' branch: https://tableau.github.io/server-client-python/docs/#install-from-the-development-branch

mwilton:
Using that dev branch to solve this problem.  It works, but then the following no longer works:
workbook.show_tabs = True
self.server.workbooks.update(workbook)

error: 
400000: Bad Request     Payload is either malformed or incomplete

nachtgold:
> 
> 
> Using that dev branch to solve this problem. It works, but then the following no longer works:
> workbook.show_tabs = True
> self.server.workbooks.update(workbook)
> 
> error:
> 400000: Bad Request Payload is either malformed or incomplete

I have the same problem with Tableauserverclient 0.8 and Tableau 20183.19.0123.2221

AlexandreRodrigues1:
Also having the same problem with tableau version 2018.3.2
and TSC version 0.8

PR #403, opened 2019-02-20
Added support for flows

shinchris:
There are 3 types of schedules now with 2019.1: Extract, Flow, and Subscription. Flow has been added to address #402.


RussTheAerialist:
Can you add a unit test to cover the functionality? Otherwise 🚀

RussTheAerialist:
🚀

Issue #381, opened 2019-01-19
Cannot update revision history setting for site


baixin137:
https://github.com/tableau/server-client-python/blob/b7b544e5ef405ca76f9e0c5681b001373811c6b2/tableauserverclient/server/request_factory.py#L291
I'm pretty sure it should be **if site_item.revision_history_enabled is not None** here, this block tries to change revision history setting but it's checking **site_item.subscribe_others_enabled**
Also i think we don't want to use **if site_item.revision_history_enabled**, because it ignores the case when the value is False, and doesn't write False into the request.

shinchris:
Thanks for pointing this out, as well as the other 2 issues.

irwando:
@shinchris to close when the merge is done.

Issue #380, opened 2019-01-19
Cannot update site_item.subscribe_others_enabled to False


baixin137:
https://github.com/tableau/server-client-python/blob/b7b544e5ef405ca76f9e0c5681b001373811c6b2/tableauserverclient/server/request_factory.py#L287
Same as line 285, if the value is False the next line will be ignored and the value will not be written into request

irwando:
@shinchris to close when the request is merged and ready to go

Issue #379, opened 2019-01-19
Cannot enable subscriptions for site


baixin137:
https://github.com/tableau/server-client-python/blob/b7b544e5ef405ca76f9e0c5681b001373811c6b2/tableauserverclient/server/request_factory.py#L285
The condition here means **disable_subscriptions** will not be written into the request **if site_item.disable_subscriptions is not None or site_item.disable_subscriptions != False**
Looks like when the value is not None we need to update the site with given value (could be True or False), however, False would be ignored here along with None, which means you can never update site to site.disable_subscriptions = False
Should change to **if site_item.disable_subscriptions is not None**

irwando:
@shinchris to close this when the merge is done and this is ready to go.

PR #408, opened 2019-03-11
Add new fields for SiteItem and clean up

shinchris:
Went through all the site-related operations to fill in missing gaps and to clean up. Also addressing #379 #380 #381.
1. Added fields:
- CacheWarmupEnabled
- CommentingEnabled
- ExtractEncryptionMode
- FlowsEnabled
- GuestAccessEnabled
- TierCreatorCapacity
- TierExplorerCapacity
- TierViewerCapacity
- NumCreators (usage)
- NumExplorers (usage)
- NumViewers (usage)

2. Deleted StatusReason field, because this field never existed in any of our XSDs.
3. Changed SiteItem parsing logic to avoid having a laundry list parameter of fields
4. Added support for '?includeUsage=True' URL param option to get usage information
5. Added a new property decorator to represent positive numbers (userQuota for example)


tableauserverclient/models/property_decorators.py
```diff
@@ -69,7 +69,27 @@ def wrapper(self, value):
...
 
 
-def property_is_int(range, allowed=None):
+def property_is_positive_int(allowed=None):
+    ''' Ensures that a property must be a positive integer (e.g. site storage quota).
+    Also takes in a list of allowed values to override the rule. This is useful for
+    when a property can use sentinel values.
+    '''
+    if allowed is None:
+        allowed = ()    # Empty tuple for fast no-op testing.
```
t8y8: This isn't an empty tuple, should be `(,)`

tableauserverclient/models/site_item.py
```diff
@@ -125,118 +208,129 @@ def subscribe_others_enabled(self, value):
...
-        if storage_quota:
-            storage_quota = int(storage_quota)
-
-        revision_limit = site_xml.get('revisionLimit', None)
-        if revision_limit:
-            revision_limit = int(revision_limit)
-
-        num_users = None
-        storage = None
+        site_fields = site_xml.attrib
```
t8y8: I'm not sure this is a net improvement.

It's more concise, but it also means we're not parsing in this layer anymore, we're just trusting whatever the XMLParser grabbed from the attrib, we can't handle specific errors in this layer, or scan the code to see what's getting parsed out and how.

tableauserverclient/models/site_item.py
```diff
@@ -125,118 +208,129 @@ def subscribe_others_enabled(self, value):
...
         usage_elem = site_xml.find('.//t:usage', namespaces=ns)
         if usage_elem is not None:
-            num_users = usage_elem.get('numUsers', None)
-            storage = usage_elem.get('storage', None)
-
-        materialized_views_mode = site_xml.get('materializedViewsMode', '')
+            usage_fields = usage_elem.attrib
+            site_fields['usage'] = usage_fields
```
t8y8: This would conflict with an attribute named 'usage' if we were ever to add one to the site element. 

Would it be a little cleaner to return two dicts (or, in the case of more nested elements, one dict for each element) or maybe a wrapper dict that has a 'site_fields' and 'usage_fields' sub dict? It just feels like a cleaner model if we wanted to go apply this cleanup to other more deeply nested return types.

(And yeah, that means I'm coming around on just using 'attrib')

t8y8:
Closing because I've implemented this pattern in 'large' models. We can redo this for sites in the future by reopening this or making a new one

PR #410, opened 2019-03-12
Fix #projects anchor

philshon:
Anchor name was incorrect and breaking navigation to the Project API


shinchris:
Thanks for finding and fixing this. We now have all of our documentations under the "gh-pages" branch of our repository, instead of master. So I've made the same change in "gh-pages".

PR #429, opened 2019-05-07
Permissions Support 2.0

t8y8:
This is an old massive PR that slowly works on getting permissions into tsc.

@DepthDeluxe contributed a lot (most) of the initial code, but I've forked and simplified a lot of the interfaces. I think we can still build the convenience functions they want but this is a simpler starting point.

I have two main open questions:

1. What do we do with Grantees? The GranteeUser/Group object only needs an 'id' to be passed into a PermissionsRule. However, the Group and User Item classes have a bunch of required (and non-nullable) properties. It's nice that we can just pass in a UserItem or GroupItem from a query, into PermissionsRules but I ALSO want to enable the ability to just create the users/groups you need based on ID. Currently I have this subtyped and overriding the problematic properties, but we could also just change UserItem/GroupItem to be more flexible. A middle ground is some kind of classmethod that constructs a UserItem/GroupItem out of only ID (a classmethod of some sort). What do you think?

2. How should we handle Default Permissions? [Help Link](https://onlinehelp.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#add_default_permissions)
I have a few ideas, one is to just have a default permissions endpoint that's largely copy/paste from the permissions endpoint, and then have an update method on it that takes (ResourceType, List[PermissionsRule]). Alternatively I can reuse most of the existing code if I just make a `update_datasource_default_permissions`, `update_workbook_default_permissions`, etc -- on to the permissions endpoint, and then somehow only enable that for projects (via an optional parameter in the init for the permissions endpoint? Via a classmethod? something else?) I need to prototype here more.

/cc @shinchris @RussTheAerialist @gaoang2148 

If I can get answers to those two, I can implement the remaining stuff and start on the tests.

### DIFF
--- a/tableauserverclient/models/group_item.py
+++ b/tableauserverclient/models/group_item.py
@@ -61,8 +61,16 @@ class GroupItem(object):
             all_group_items.append(group_item)
         return all_group_items
 
+    @staticmethod
+    def for_permissions(id_):
+        return GranteeGroup(id_)
+
 
 class GranteeGroup(GroupItem):
+    """Reduced version of the GroupItem class that lets you
+    build Group objects with only an ID. Necessary for Groups returned
+    from Permissions requests, and a shortcut for creating permissions.
+    """
 
     def __init__(self, id_):
         self.id = id_
--- a/tableauserverclient/models/group_item.py
+++ b/tableauserverclient/models/group_item.py
@@ -67,18 +67,10 @@ class GranteeGroup(GroupItem):
     def __init__(self, id_):
         self.id = id_
 
-    @property
-    def id(self):
-        return self._id
-
-    @id.setter
+    @GroupItem.id.setter
     def id(self, value):
         self._id = value
 
-    @property
-    def name(self):
-        return self._name
-
-    @name.setter
+    @GroupItem.name.setter
     def name(self, value):
         self._name = value
--- a/tableauserverclient/models/group_item.py
+++ b/tableauserverclient/models/group_item.py
@@ -3,8 +3,6 @@ from .exceptions import UnpopulatedPropertyError
 from .property_decorators import property_not_empty
 
 
-    
-
 class GroupItem(object):
 
     permissions_grantee_type = 'group'
@@ -62,6 +60,7 @@ class GroupItem(object):
                 group_item._domain_name = domain_elem.get('name', None)
             all_group_items.append(group_item)
         return all_group_items
+
 
 class GranteeGroup(GroupItem):
 
--- a/tableauserverclient/models/group_item.py
+++ b/tableauserverclient/models/group_item.py
@@ -1,11 +1,12 @@
 import xml.etree.ElementTree as ET
 from .exceptions import UnpopulatedPropertyError
 from .property_decorators import property_not_empty
+from .reference_item import ResourceReference
 
 
 class GroupItem(object):
 
-    permissions_grantee_type = 'group'
+    tag_name = 'group'
 
     def __init__(self, name=None):
         self._domain_name = None
@@ -38,6 +39,9 @@ class GroupItem(object):
         #  Each call to `.users` should create a new pager, this just runs the callable
         return self._users()
 
+    def to_reference(self):
+        return ResourceReference(id_=self.id, tag_name=self.tag_name)
+
     def _set_users(self, users):
         self._users = users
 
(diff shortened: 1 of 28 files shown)


tableauserverclient/models/permissions_item.py
```diff
@@ -0,0 +1,103 @@
...
+        parsed_response = ET.fromstring(resp)
+
+        rules = []
+        permissions_rules_list_xml = parsed_response.findall('.//t:granteeCapabilities',
+                                          namespaces=ns)
+
+        for grantee_capability_xml in permissions_rules_list_xml:
+            capability_dict = {}
+
+            grantee_element = grantee_capability_xml.findall('.//*[@id]', namespaces=ns).pop()
```
t8y8: This could move to a grantee constructor that builds the right subtype

shinchris:
Thanks for tackling the permissions work!

Here are my thoughts for the two questions above:
1. My initial thought was that creating an item based on ID would go against our existing patterns, but I can't come up with a better way to avoid that due to the design of the permissions requests. I think the subtyped grantees are good since we can just pass in a UserItem or GroupItem from a query into PermissionsRules, rather than extracting out the IDs and creating new grantee objects. I think that would be better than making changes to make UserItem and GroupItem more flexible.

2. I think the first idea of adding a default permissions function to PermissionsEndpoint and then calling that from WorkbookEndpoint/DatasourceEndpoint with the ResourceType would be cleaner with the structure we have now. Maybe we'll also support flows some day.

/cc @RussTheAerialist @gaoang2148

t8y8:
@shinchris Great, I moved the `Grantee*` classes to behind a static method.

On the default front, I encountered a problem, the update/delete can take a parameter called 'template_type' for now, but maybe 'target_type' or 'default_permissions_type'... something like that.

However, the 'get' is a problem, it doesn't really make sense to merge them all, since it's all subsets:
Default permissions for workbooks, data sources, and flows.

I'm thinking that a project will have
project.item.workbook_default_permissions
project.item.datasource_default_permissions
project.item.flow_default_permissions

With their own 'populate' methods?

tableauserverclient/models/group_item.py
```diff
@@ -53,3 +56,25 @@ def from_response(cls, resp, ns):
                 group_item._domain_name = domain_elem.get('name', None)
             all_group_items.append(group_item)
         return all_group_items
+
+    @staticmethod
+    def for_permissions(id_):
+        return GranteeGroup(id_)
+
+
+class GranteeGroup(GroupItem):
```
gaoang2148: Instead of a subtype, make an AssetReference (ObjectReference?) class that has
```
class ResourceReference:
    def __init__(self, id_):
        self.id=id_
    
    @classmethod tag_name: str
```

And then you have UserReference(ResourceReference).
UserItem.@as_reference() -> UserReference

UserItem inherits from UserReference (Or not, but it does have the method to make an as_reference)

t8y8:
Aaaaand done.

I've made the calls all consistent, and implemented the Reference pattern.

I decided not to subtype things because they all need a tag_name that they take from the Item class, and they all need an ID passed in to the constructor... the subtypes added no value at all!

Maybe someday they will, we can safely refactor this then, but for now, this implements the `Reference` method with takes the tag_name defined from the real content-type class, and then an ID.

Ran a manual test and the few tests in the suite, I'll add more tests but wanted to get this out for review so I can start adding the other content types.

tableauserverclient/server/endpoint/projects_endpoint.py
```diff
@@ -50,3 +60,51 @@ def create(self, project_item):
...
+    @api(version='2.0')
+    def update_permission(self, item, rules):
+        return self._permissions.update(item, rules)
+
+    @api(version='2.0')
+    def delete_permission(self, item, rules):
+        return self._permissions.delete(item, rules)
+
+    @api(version='2.1')
+    def populate_workbook_default_permissions(self, item):
```
shinchris: I might be remembering wrong, but I thought we talked about having 1 method for each verb that takes in a map containing all default permissions for all resource types?

t8y8: We did at first, but when we reconsidered in terms of type-safety (or at least the premise of it with autocomplete and maybe py3 mypy stuff that I'd love ot add) it felt a little safer to go this route, than having to introspect dicts and such.

Basically, is it less work to say
``` server.update(THING_TYPE, THING)```
or
```server.update_type_of_thing(thing)```

And I think for consistency and simplicity, the latter won out.

shinchris:
🚀
@gaoang2148 @RussTheAerialist

PR #431, opened 2019-05-07
Initial support for Tableau Metadata API

t8y8:
In preparation for the Metadata API (soon to be in wide beta) I wanted to add support to TSC.

It's very basic right now, it just takes a GraphQL query (a string) and returns a dict from the json response.

We could build in some helpers over time for pagination (It's more complicated in GraphQL) and methods to convert from metadata node objects to REST API objects... but for now this should let people get going.

### DIFF
--- a/tableauserverclient/server/endpoint/metadata_endpoint.py
+++ b/tableauserverclient/server/endpoint/metadata_endpoint.py
@@ -24,7 +24,8 @@ class Metadata(Endpoint):
 
         # Setting content type because post_reuqest defaults to text/xml
         server_response = self.post_request(url, graphql_query, content_type='text/json')
-        results = json.loads(server_response.content)
+        breakpoint()
+        results = server_response.json()
 
         if abort_on_error and results.get('errors', None):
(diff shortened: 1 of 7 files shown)


tableauserverclient/server/endpoint/metadata_endpoint.py
```diff
@@ -0,0 +1,32 @@
...
+
+logger = logging.getLogger('tableau.endpoint.metadata')
+
+
+class Metadata(Endpoint):
+    @property
+    def baseurl(self):
+        return "{0}/api/exp/metadata/graphql".format(self.parent_srv._server_address)
+
+    @api("3.2")
```
t8y8: @gaoang2148 what's the latest, is it 3.4?

gaoang2148: 3.4 is 19.2, 3.5 is 19.3

gaoang2148: This also jumps out to me, minor issue so just merged!

PR #432, opened 2019-05-07
Update .travis.yml

t8y8:
3.3 and 3.4 are EOL, as well as not significantly different than the other 3's.

Dropping them, and pypy (which is also not significantly different at this point, and is slow to run the pipeline), and adding 3.7


shinchris:
🚀

PR #434, opened 2019-05-08
Add support for Cataloging and Prep Conductor to TSC.

t8y8:
These are toggleable only when your server is licensed for it; but these settings themselves are simple site settings.


shinchris:
🚀

PR #435, opened 2019-05-11
Https link

coliff:


coliff:
No downside. Http redirects to https anyway - so this saves a redirect and is slightly faster.

PR #439, opened 2019-05-24
Change server_info endpoint error message to be verbose

jacobj10:
Add handling for case in which API version is not supported by the Tableau server. While trying to explore this library, I ran into an error when trying to call `server.server_info_get()`
```
>>> server.version = "2.5"
>>> server.server_info.get()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jojacob/development/dasre-tableau-insights_trunk/build/dasre-tableau-insights/venv/lib/python3.7/site-packages/tableauserverclient/server/endpoint/endpoint.py", line 108, in wrapper
    return func(self, *args, **kwargs)
  File "/Users/jojacob/development/dasre-tableau-insights_trunk/build/dasre-tableau-insights/venv/lib/python3.7/site-packages/tableauserverclient/server/endpoint/server_info_endpoint.py", line 25, in get
    server_info = ServerInfoItem.from_response(server_response.content, self.parent_srv.namespace)
UnboundLocalError: local variable 'server_response' referenced before assignment
```

This stems from the code here:

https://github.com/jacobj10/server-client-python/blob/e4ec849b19f02f016f90e641756d9bd81d23d026/tableauserverclient/server/endpoint/server_info_endpoint.py#L17-L21

which can catch an error without setting the variable, leading to a confusing and non descriptive UnboundLocalError. Right now I just added an extra check for my specific error (version unsupported), but it my be worth just raising whatever error comes from the response.

This is my first time contributing so please let me know if you have any questions, comments, or concerns.

### DIFF
--- a/tableauserverclient/server/endpoint/server_info_endpoint.py
+++ b/tableauserverclient/server/endpoint/server_info_endpoint.py
@@ -22,6 +22,5 @@ class ServerInfo(Endpoint):
             if e.code == "404001":
                 raise EndpointUnavailableError
 
-
         server_info = ServerInfoItem.from_response(server_response.content, self.parent_srv.namespace)
         return server_info
--- a/tableauserverclient/server/endpoint/server_info_endpoint.py
+++ b/tableauserverclient/server/endpoint/server_info_endpoint.py
@@ -19,6 +19,8 @@ class ServerInfo(Endpoint):
         except ServerResponseError as e:
             if e.code == "404003":
                 raise ServerInfoEndpointNotFoundError
+            if e.code == "404001":
+                raise EndpointUnavailableError
 
         server_info = ServerInfoItem.from_response(server_response.content, self.parent_srv.namespace)
         return server_info
--- a/tableauserverclient/server/endpoint/server_info_endpoint.py
+++ b/tableauserverclient/server/endpoint/server_info_endpoint.py
@@ -22,5 +22,6 @@ class ServerInfo(Endpoint):
             if e.code == "404001":
                 raise EndpointUnavailableError
 
+
         server_info = ServerInfoItem.from_response(server_response.content, self.parent_srv.namespace)
         return server_info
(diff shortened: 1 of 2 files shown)


t8y8:
I haven't reviewed in detail yet, but I think the problem is that we handle this a level above at https://github.com/jacobj10/server-client-python/blob/e4ec849b19f02f016f90e641756d9bd81d23d026/tableauserverclient/server/server.py#L76

In this case, you're manually setting the API version to 2.5, avoiding the 'use legacy version check' fallback. I agree we should handle this, and I'll take a look at your PR!

jacobj10:
Ahh I see, I think I was manually updating the version due to confusion when reading the documentation. The correct way is to use the class methods?

t8y8:
Hey!

Yeah, the recommendation is to just use the methods like this:
```py
auth = TSC.TableauAuth(USERNAME, PASSWORD)
server = TSC.Server("http://SERVER", use_server_version=True)

# OR
server = TSC.Server("http://SERVER")
server.use_server_version()
```

t8y8:
... I'm not sure how that test is failing given you didn't touch sorting/filtering.

Can you make sure your fork is updated tot he latest development branch?

jacobj10:
Weird, seems like the fork is up to date with the origin dev branch? I tried kicking off another CI run but to no avail :(

jacobj10:
The test fails locally on my master branch as well

t8y8:
The only explanation I have is some update in a library we depend on changed the encoding behavior?

Feel free to take a look at that and include it in your PR, or ignore it and I won't hold it against you :)

jacobj10:
Ack I'll try to include in this PR, I'll let you know if I can't figure it out.

jacobj10:
Yep that did it. More of a single issue fix. It might be worth seeing if there are any `utils` that the `request` module uses to URI decode/encode the response and use that accordingly with all relevant assertions, but that seems like it should be addressed in its own PR if at all. LGTM!

t8y8:
Thanks, I'll take it!

Can you confirm you've sent in the CLA? (I don't have an easy way to check on my end) I'll wait a day and then merge this.

jacobj10:
Sent :)

james-baker:
CLA approved.

PR #445, opened 2019-06-13
WIP: External Content Support

t8y8:
This is a work in progress, and is based on the permissions work, so this PR will fail against dev.

Still, here's a preview of the content types coming. They're pretty standard, following existing patterns.

The one thing that's a bit new, is I took some pieces of @shinchris 's refactor for objects that have lots of attributes.

### DIFF
--- a/tableauserverclient/models/column_item.py
+++ b/tableauserverclient/models/column_item.py
@@ -1,0 +1,67 @@
+import xml.etree.ElementTree as ET
+
+from .property_decorators import property_is_enum, property_not_empty
+from .exceptions import UnpopulatedPropertyError
+
+
+class ColumnItem(object):
+    def __init__(self, name, description=None):
+        self._content_permissions = None
+        self._id = None
+        self.description = description
+        self.name = name
+
+    @property
+    def id(self):
+        return self._id
+
+    @property
+    def name(self):
+        return self._name
+
+    @name.setter
+    @property_not_empty
+    def name(self, value):
+        self._name = value
+
+    def _set_values(self, id, name, description, remote_type, embedded, schema):
+        if id is not None:
+            self._id = id
+        if name:
+            self._name = name
+        if description:
+            self.description = description
+        if embedded:
+            self.embedded = embedded
+        if schema:
+            self.schema = schema
+        if remote_type:
+            self.remote_type = remote_type
+
+    @classmethod
+    def from_response(cls, resp, ns):
+        all_column_items = list()
+        parsed_response = ET.fromstring(resp)
+        all_column_xml = parsed_response.findall('.//t:column', namespaces=ns)
+
+        for column_xml in all_column_xml:
+            (id, name, description, remote_type, embedded, schema) = cls._parse_element(column_xml, ns)
+            column_item = cls(name)
+            column_item._set_values(id, name, description, remote_type, embedded, schema)
+            all_column_items.append(column_item)
+        return all_column_items
+
+    @staticmethod
+    def _parse_element(column_xml, ns):
+        id = column_xml.get('id', None)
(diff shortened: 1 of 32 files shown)


t8y8:
Now that permissions is in, I've rebased this against HEAD.
The commits will all get squashed, don't worry about the gross history :P

This is much simpler than permissions, though it is more vulnerable to copy/paste errors, so some dedicated eyes are much appreciated. I removed the WIP, but I actually haven't spent as much time on this set, so I'll be testing against a Tableau Catalog enabled Server over the next few days to make sure this is ready. I will also add some tests :)

tableauserverclient/server/endpoint/tables_endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+        logger.info('Populated columns for table (ID: {0}'.format(table_item.id))
+
+    def _get_columns_for_table(self, table_item, req_options=None):
+        url = "{0}/{1}/columns".format(self.baseurl, table_item.id)
+        server_response = self.get_request(url, req_options)
+        columns = ColumnItem.from_response(server_response.content,
+                                           self.parent_srv.namespace)
+        return columns
+
+    # Update workbook_connection
```
shinchris: Copied comment from workbook

tableauserverclient/models/table_item.py
```diff
@@ -0,0 +1,140 @@
...
+    @description.setter
+    def description(self, value):
+        self._description = value
+
+    @property
+    def certified(self):
+        return self._certified
+
+    @certified.setter
+    def certified(self, value):
```
shinchris: This one is missing the boolean property decorator

tableauserverclient/models/database_item.py
```diff
@@ -0,0 +1,241 @@
...
+    @property
+    def certification_note(self):
+        return self._certification_note
+
+    @certification_note.setter
+    def certification_note(self, value):
+        self._certification_note = value
+
+    @property
+    def metadata_type(self):
```
shinchris: This one and a few below are not defined in the constructor. Also for port, should it be an integer property?

tableauserverclient/models/column_item.py
```diff
@@ -0,0 +1,69 @@
...
+    def description(self):
+        return self._description
+
+    @description.setter
+    def description(self, value):
+        self._description = value
+
+    @property
+    def remote_type(self):
+        return self._remote_type
```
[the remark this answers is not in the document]
t8y8: That's a good point -- what is the pattern here? 
That's not a user-editable attribute, and it's something we set purely from the XML response, sometimes we put these in constructors, sometimes we don't.

I can't add it here, just wondering.

tableauserverclient/models/database_item.py
t8y8: Not all of these properties will exist for all types (though `metadata_type` should).

Right now it lazily adds them based on whatever we pass to `_set_values`. 
We could enumerate them all, but I thought it was a lot of copy/paste work to add them there and to `_set_values`. 

Happy to do it if you think we should. It raises the question again about what makes it into constructor vs what doesn't.

t8y8:
Thanks! Question for you, fixed the small things, and I will create some basic tests for these endpoints, it'll take a bit... lots of xml files to copy/paste :)

tableauserverclient/models/database_item.py
shinchris: In my opinion, I think it's good practice to have all properties defined in the constructor. That way we can easily see all the properties that the class can have, and avoids the possibility of an error from accessing it before assigned from the response.
Looks like we've been keeping that pattern for the other models as well.

PR #449, opened 2019-06-19
Fix serialization error in WB request generation

jacobj10:
I was trying to publish a workbook via the TSC and received the following error:
`TypeError: cannot serialize UUID('bfd7b712-****-****-****-************') (type UUID)`

Seems like this should be fixed on the UUID level? But this is a workaround nevertheless.


t8y8:
Looks good to me!

PR #451, opened 2019-06-19
Pass **kwargs through to the paged function in Pager

### LINKED ISSUE #444: Pager does not allow pulling usage information for views.get()
The Pager generator takes in RequestOptions, but usage information for views.get() requires another parameter to be passed in. Currently, we cannot use Pager to get all views WITH usage information.

t8y8:
This should fix #444 and any other time we need to page over a 'get' that has some extra parameters


shinchris:
Thanks for doing this! 🚀

Issue #454, opened 2019-07-01
Workbook Update Fails in Tableau Server versions prior to Materialized Views


austinderrick:
A simple update workbook call fails on any Tableau Server version prior to the introduction of materialized views. This occurs in TSC version 0.8, which added support for materialized views. Downgrading to 0.7 fixes the issue.

Basically the XML for materialized views is appended to the update call and Tableau Server does not know how to handle it, resulting in failed API calls.

From the Tableau Server logs (2018.2):
````org.springframework.http.converter.HttpMessageConversionException: Deserialization problem: unexpected element (uri:"", local:"materializedViewsEnablementConfig"). Expected elements are <{}owner>,<{}site>,<{}project>,<{}connections>,<{}connectionCredentials>,<{}views>,<{}tags>; nested exception is com.sun.istack.internal.SAXParseException2; lineNumber: 1; columnNumber: 274; unexpected element (uri:"", local:"materializedViewsEnablementConfig"). Expected elements are <{}owner>,<{}site>,<{}project>,<{}connections>,<{}connectionCredentials>,<{}views>,<{}tags>````

Code for reproducing the problem:

```python
        published_workbook = server.workbooks.get_by_id(published_workbook_id)
        published_workbook.owner_id = new_owner_id
        published_workbook.show_tabs = True

        # Added to try to override; didn't help
        published_workbook.materialized_views_config = None

        workbook = server.workbooks.update(published_workbook)
        server.auth.sign_out()
```

shinchris:
This has been fixed with the v0.8.1 release.

PR #461, opened 2019-07-15
Fix update_workbook endpoint to address #454

shinchris:
Fixing error in update workbook that was caused by materialized view config being included in the request, even for old server versions.
Also fixed a bug where showTabs property would not get updated properly if changing it to 'false'.


RussTheAerialist:
🚀 also, we should cut a new release for this fix.  Let's talk about doing
0.8.1 tomorrow when we are both in the office.

PR #464, opened 2019-07-20
Adding support for Personal Access Token login

fpagliar:
The rest api sign in request will support providing a personal access token, along with its name, as a valid credential instead of username and password.

Setting up shared code to be able to migrate all samples to use the same helper and give support to standard arguments, like serverURL, logLevel and sign in credentials

### DIFF
--- a/samples/setup_helper.py
+++ b/samples/setup_helper.py
@@ -1,0 +1,62 @@
+####
+# This script provides the shared login logic and server creation for all other samples.
+#
+# To run the script, you must have installed Python 2.7.9 or later.
+####
+
+import argparse
+import getpass
+import logging
+
+from datetime import time
+
+import tableauserverclient as TSC
+
+class SetupHelper(object):
+    def __init__(self, description):
+        self._parser = argparse.ArgumentParser(description=description)
+        self._addServerArgs()
+        self._addLoginArgs()
+        self._addLoggerArgs()
+
+    def _addLoginArgs(self):
+        group = self._parser.add_mutually_exclusive_group(required=True)
+        group.add_argument('--username', '-u', help='username to sign into the server')
+        group.add_argument('--token-name', '-n', help='name of the personal access token used to sign into the server')
+
+    def _addServerArgs(self):
+        self._parser.add_argument('--server', '-s', required=True, help='server address')
+
+    def _addLoggerArgs(self):
+        self._parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
+                    help='desired logging level (set to error by default)')
+
+    def getParser(self):
+        return self._parser
+
+    def createServer(self):
+        args = self._parser.parse_args()
+
+        # Set logging level based on user input, or error by default
+        logging_level = getattr(logging, args.logging_level.upper())
+        logging.basicConfig(level=logging_level)
+
+        self._server = TSC.Server(args.server)
+        return self._server
+
+    def login(self):
+        if not self._server:
+            raise RuntimeError('Please call createServer first')
+
+        args = self._parser.parse_args()
+
+        if args.username:
+            # Trying to authenticate using username and password.
+            password = getpass.getpass("Password: ")
+            tableau_auth = TSC.TableauAuth(args.username, password)
+        else:
+            # Trying to authenticate using personal access tokens.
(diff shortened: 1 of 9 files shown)


samples/setup_helper.py
```diff
@@ -0,0 +1,63 @@
...
+        self._parser.add_argument('--server', '-s', required=True, help='server address')
+
+    def _addLoggerArgs(self):
+        self._parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
+                                  help='desired logging level (set to error by default)')
+
+    def getParser(self):
+        return self._parser
+
+    def createServer(self):
```
t8y8: snake_case

tableauserverclient/models/tableau_auth.py
```diff
@@ -1,5 +1,6 @@
 class TableauAuth(object):
-    def __init__(self, username, password, site=None, site_id='', user_id_to_impersonate=None):
+    def __init__(self, username='', password='', site=None, site_id='', user_id_to_impersonate=None, token_name='',
```
t8y8: It's been a few versions, maybe we can safely remove the deprecated site parameter while we're here. (\cc @RussTheAerialist?)

tableauserverclient/models/tableau_auth.py
```diff
@@ -11,6 +12,8 @@ def __init__(self, username, password, site=None, site_id='', user_id_to_imperso
         self.password = password
         self.site_id = site_id
         self.username = username
+        self.token_name = token_name
```
t8y8: I believe there's logic where some of these can't be set at the same time, we should check in the __init__ and throw an exception for those cases (is it either username/password OR PAT?)

tableauserverclient/models/tableau_auth.py
```diff
@@ -1,5 +1,6 @@
 class TableauAuth(object):
-    def __init__(self, username, password, site=None, site_id='', user_id_to_impersonate=None):
+    def __init__(self, username='', password='', site=None, site_id='', user_id_to_impersonate=None, token_name='',
+                 personal_access_token=''):
```
t8y8: Why default to `''` instead of `None` for these params? Admittedly we're a little inconsistent, but `None` won't silently get serialized if there's a bug

samples/setup_helper.py
```diff
@@ -0,0 +1,63 @@
+####
+# This script provides the shared login logic and server creation for all other samples.
```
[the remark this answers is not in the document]
fpagliar: It's fine either way for me. 
I just thought it would be easier to have centralized login to give support for PersonalAccessTokens vs copy-pasting it into each sample. I saw different samples take different arguments to log in, so I wanted a consistent experience as a cli user trying it out.

I'm not sure about the style followed by the project in general. If you think plain is better, I can change it back.

tableauserverclient/models/tableau_auth.py
fpagliar: The server will return an error if both are set with a clear message saying they are not allowed.
Do we want to replicate that logic in the client side? 
I haven't seen much validation going on in the models so I kept it out.

t8y8: The example I was thinking of was `connections` and `connection_credentials`, though I looked into it and that's validated in the endpoint, where it chooses which type of model to build, and not in the model itself.

If the error from Server is clear and caught correctly, that's sufficient thanks!

(The alternative I thought was to have 2 different auth objects, to keep this one from growing too large. Maybe that's not necessary now though)

samples/setup_helper.py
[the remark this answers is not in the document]
t8y8: In that case, I would keep this out of the change -- each script can be stand alone and therfore copy/pasted around to the users needs.

tableauserverclient/models/tableau_auth.py
RussTheAerialist: The pattern should be to create a different object for personal access tokens.  TableauAuth is specifically for Tableau username/password login.  You want to create a PersonalTokenAuth or similarly named object.

fpagliar: Implemented the new approach on PR 465

PR #465, opened 2019-07-23
Personal Access Token Authentication

fpagliar:
The rest api sign in request will support providing a personal access token, along with its name, as a valid credential instead of username and password.

Created a new credential type aside of TableauAuth. They both have the same request/response format, the only difference is the credentials section of the request, where they provide different arguments.

Adding tests and a sample that shows how to use the new capabilities

### DIFF
--- a/tableauserverclient/models/personal_access_token_auth.py
+++ b/tableauserverclient/models/personal_access_token_auth.py
@@ -6,5 +6,6 @@ class PersonalAccessTokenAuth(object):
         # Personal Access Tokens doesn't support impersonation.
         self.user_id_to_impersonate = None
 
+    @property
     def credentials(self):
         return {'clientId': self.token_name, 'personalAccessToken': self.personal_access_token}
--- a/tableauserverclient/models/personal_access_token_auth.py
+++ b/tableauserverclient/models/personal_access_token_auth.py
@@ -7,4 +7,4 @@ class PersonalAccessTokenAuth(object):
         self.user_id_to_impersonate = None
 
     def credentials(self):
-        return { 'clientId': self.token_name, 'personalAccessToken': self.personal_access_token }
+        return {'clientId': self.token_name, 'personalAccessToken': self.personal_access_token}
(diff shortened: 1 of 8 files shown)


tableauserverclient/models/personal_access_token_auth.py
```diff
@@ -0,0 +1,10 @@
+class PersonalAccessTokenAuth(object):
+    def __init__(self, token_name, personal_access_token, site_id=''):
+        self.token_name = token_name
+        self.personal_access_token = personal_access_token
+        self.site_id = site_id
+        # Personal Access Tokens doesn't support impersonation.
+        self.user_id_to_impersonate = None
```
t8y8: Why include it in the object if it's not supported?

tableauserverclient/models/tableau_auth.py
```diff
@@ -25,3 +25,6 @@ def site(self, value):
         warnings.warn('TableauAuth.site is deprecated, use TableauAuth.site_id instead.',
                       DeprecationWarning)
         self.site_id = value
+
+    def credentials(self):
```
t8y8: Please make this a `@property`

tableauserverclient/server/request_factory.py
```diff
@@ -47,11 +47,14 @@ def _add_credentials_element(parent_element, connection_credentials):
 class AuthRequest(object):
     def signin_req(self, auth_item):
         xml_request = ET.Element('tsRequest')
+
         credentials_element = ET.SubElement(xml_request, 'credentials')
-        credentials_element.attrib['name'] = auth_item.username
-        credentials_element.attrib['password'] = auth_item.password
+        for attribute_name, attribute_value in auth_item.credentials().items():
```
t8y8: When this is a property you can remove the `()` here after `credentials`

tableauserverclient/models/personal_access_token_auth.py
fpagliar: Sign in logic will check against it to see if we are trying to impersonate and if it should be added in the sign in request.

fpagliar:
I think I don't have permissions to merge.

fpagliar:
Moved the sign in call to a new method to validate the rest api version before using Personal Access Tokens since older versions of the rest api won't support it

PR #468, opened 2019-07-25
Update api-ref.md

irwando:
Fix the sample for datasource.update in response to Issue 382.  Data Sources cannot be renamed yet our sample showed it.  I changed the sample to show project_id being change.


shinchris:
Could you make the change into the gh-pages branch instead of master?

PR #469, opened 2019-07-25
Update api-ref.md

irwando:
Fixing the formatting of the table in the ConnectItem class show attributes.  It wasn't showing up as a table.


irwando:
Wrong docs.

PR #470, opened 2019-07-30
Fix support for tags in ViewItem

toomyem:
Now tags are properly being red to ViewItem.


james-baker:
CLA received.

irwando:
Looks good!

PR #472, opened 2019-07-30
Documentation: Update in Write your first program

user9747:
**TableauAuth** Class requires a third parameter **site_id** as specified in the _API reference_. But the Getting started sample code passes only **username** and **password**. This leads to an authentication  error.


irwando:
Updates should be made directly on the page:
https://tableau.github.io/server-client-python/docs/

Click "edit" at the top of the page to edit the correct spot.  This will get the changes into the correct branch.

Closing this merge request as the doc branch here is old.

PR #474, opened 2019-07-31
Documentation: Update Write your first program

user9747:
TableauAuth Class requires a third parameter site_id as given in the API reference. If only username and password is given it would lead to a Authentication Error.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -74,7 +74,7 @@ with server.auth.sign_in(tableau_auth):
     print("\nThere are {} datasources on site: ".format(pagination_item.total_available))
     print([datasource.name for datasource in all_datasources])
 ```
-For Tableau online TSC.TableauAuth should be used with site_id
+To authenticate to Tableau Online, or a site other than the default site within a local Tableau Server installation, the site_id(SITE_NAME) parameter must be used to specify the site.
 ```py
-tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD','SITEID')
+tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD','SITENAME')
 ```
--- a/docs/index.md
+++ b/docs/index.md
@@ -66,7 +66,7 @@ Run the following code to get a list of all the data sources on your installatio
 ```py
 import tableauserverclient as TSC
 
-tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD')
+tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD','SITEID')
 server = TSC.Server('http://SERVER_URL')
 
 with server.auth.sign_in(tableau_auth):
--- a/docs/index.md
+++ b/docs/index.md
@@ -74,7 +74,7 @@ with server.auth.sign_in(tableau_auth):
     print("\nThere are {} datasources on site: ".format(pagination_item.total_available))
     print([datasource.name for datasource in all_datasources])
 ```
-For Tableau online TSC.TableauAuth is used with site_id
+For Tableau online TSC.TableauAuth should be used with site_id
 ```py
 tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD','SITEID')
 ```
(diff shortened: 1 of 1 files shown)


user9747:
Is there something wrong with this PR? I am a complete noob . This issue made me loose so much time. That is why i put up this PR.

irwando:
The code as written works if you are installing to a local installation (not Tableau Online) and are logging into the default site.

In your case what were you trying to log into?

user9747:
@irwando in my case i was trying to log into Tableau Online my bad. But for people using Tableau online is there a documentation specifying this requirement. I struggled a bit because i couldnt find such a documentation.

irwando:
@user9747 No worries!

We do specify it in the API Reference under Authentication:

https://tableau.github.io/server-client-python/docs/api-ref#authentication

" For Tableau Online, you must provide a value for the site_id."  It could be useful to add an additional example for Online though.  Would you like to update your changes to be that?

user9747:
So should i add it as different example for online.

user9747:
@irwando i ve made some changes to the PR.

docs/index.md
```diff
@@ -74,7 +74,7 @@ with server.auth.sign_in(tableau_auth):
     print("\nThere are {} datasources on site: ".format(pagination_item.total_available))
     print([datasource.name for datasource in all_datasources])
 ```
-For Tableau online TSC.TableauAuth is used with site_id
+For Tableau online TSC.TableauAuth should be used with site_id
```
irwando: How about "To authenticate to Tableau Online, or a site other than the default within a local Tableau Server installation, the site_id parameter must be used to specify the site" just to be clear.  Otherwise, looks good!

dzucker-tab: Suggest:

```tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD', site_id='SITE_NAME')
    all_datasources, pagination_item = server.datasources.get()
    print("\nThere are {} datasources on site: ".format(pagination_item.total_available))
    print([datasource.name for datasource in all_datasources])```

```site_id``` can be omitted for singing in to the default site of a local Tableau Server installation.

File: docs/index.md [hunk 2]
```diff
@@ -74,3 +74,7 @@ with server.auth.sign_in(tableau_auth):
     print("\nThere are {} datasources on site: ".format(pagination_item.total_available))
     print([datasource.name for datasource in all_datasources])
 ```
+For Tableau online TSC.TableauAuth should be used with site_id
+```py
+tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD','SITEID')
+```
```
dzucker-tab: Even though the parameter is called "site_id" in the TSC library, in the example we should use 'SITENAME' as the placeholder label because it maps to the REST API site name element and NOT the REST API site id element.

irwando: Its even more confusing that that...

From the "QuerySites" API the value you want is in the "content-url" field.
In the Tableau Server Web Application it is called "Site ID".  

Dan makes a good point thought.  We should call it content-url and probably fix the naming in the UX.

user9747: Yeah its the site name.

user9747: More changes added .Please review them.

dzucker-tab: Somehow a previous comment I made got lost. I want to strongly suggest that we make things simpler by treating the default site of a local Server installation as the exception, rather than describing all the other cases and giving a second sub-example. Below is my proposal, to which I added a sentence to address the confusion Irwin describes:

```py
tableau_auth = TSC.TableauAuth('USERNAME', 'PASSWORD', 'SITENAME')
    all_datasources, pagination_item = server.datasources.get()
    print("\nThere are {} datasources on site: ".format(pagination_item.total_available))
    print([datasource.name for datasource in all_datasources])
```
```SITENAME``` is same as the content URL of your site (though the parameter is called site_id in some Tableau tooling). This parameter can be omitted for the Default site of a locally installed Tableau Server.

t8y8: "Locally installed" is slightly off, because it could be 'remote' to the script but it's still not Tableau Online. I would call it "On Premise" which is how we generally refer to self-installed Server.

Issue #382, opened 2019-01-21
Cannot rename a datasource


bfsaid85:
It is not possible to rename a datasource and the example in API reference didn't work :

-- do some updating 
  datasource.name = 'New Name'

-- call the update method with the data source item
  updated_datasource = server.datasources.update(datasource)

I would like rename a datasource and keep the relation with linked workbooks, how is it possible?

shinchris:
Datasources cannot be renamed. There seems to be a mistake in the API reference. We will fix the example in the API reference.

dzucker-tab:
@bfsaid85 Fix for this is merged, thanks for the find! and thanks to @irwando and @shinchris

PR #476, opened 2019-08-02
Update api-ref.md

irwando:
Updating datasource.update to be an update to ownerID instead of name as data sources can't be renamed. Also added a note that name cannot be changed in teh description.


irwando:
Fixes Issue #382

Issue #346, opened 2018-10-03
Remove the old docs from the master branch


mbren:
The docs have moved to the gh-pages branch. To reduce confusion, consider removing the old docs from the master branch.

Talvalin:
Is the gh-pages branch publicly accessible? If so, could someone provide a link please? I'd like to update the current docs with the changes that I originally made to [sign-in-out.md](https://github.com/tableau/server-client-python/blob/master/docs/docs/sign-in-out.md), as well as adding more information about logging on to a site as per #355

irwando:
@t8y8 Yes.  Just go to the page you want to edit and click edit on the top.

PR #478, opened 2019-08-07
removing doc files and adding readme to direct users to the gh-pages …

dzucker-tab:
removing doc files and adding readme to direct users to the gh-page branch for doc source


dzucker-tab:
this is in response to #346

PR #513, opened 2019-10-23
Runtime error in permissions_endpoint

martinbpeters:
Runtime error: grantee.permissions_grantee_type replaced with grantee.tag_name in the permissions_endpoint delete method.


martinbpeters:
Apologies for the delay. I've actioned those three items.

t8y8:
Merged! Thanks for the fix!

donedeal-martin:
Great! You're welcome. When do we expect this to make it into the master branch and released?

Issue #514, opened 2019-10-23
You are trying to throw an error that you never even imported


dprophet:
You are trying to throw an error

NonXMLResponseError

https://github.com/tableau/server-client-python/blob/master/tableauserverclient/server/endpoint/endpoint.py#L73

That you never imported. This totally breaks the debugging for us Corporate users that are behind all kinds of security systems. Please Fix ASAP.

dprophet:
I totally disagree with this NonXMLResponseError error approach. The errors I am experience have to do with low level proxy, firewalls, certificates of authority, and NonXMLResponseError outright masks the error. These errors happen long before any communication has happened with Tableau Server.

t8y8:
The issue I was trying to solve is that the current error handling code assumes you can parse one of our XML payloads with a specific set of tags and attributes.

ServerResponseError will throw a fairly gross stack trace if we revert this and you get an error from the Metadata API endpoints.

What I can do, is try to parse the XML, fail, and reraise the original exception

dprophet:
First thing is first. Fix the import on

https://github.com/tableau/server-client-python/blob/master/tableauserverclient/server/endpoint/endpoint.py#L1

You need to import NonXMLResponseError

t8y8:
Agreed, I pushed a fix in https://github.com/tableau/server-client-python/pull/515, it's in development now (where you can pull from to verify) -- we're discussing a patch release to master shortly as well.

bcantoni:
I'm going to close this based on the end of the discussion where Tyler added the fix to import that error properly.

PR #515, opened 2019-10-24
Quick Fix for NonXMLErrorException

t8y8:
This is a quick fix in response to #514. 

This should pass through any response body, regardless of whether or not it has one of our well-formed error XML objects in it.

However, it doesn't address all of the reporting user's issues -- for that I think we would need to re-evaluate how we catch exceptions and bubble them up.

The user requested we go back to v0.8.1 functionality, but I believe that would just throw ParseError because of the check in https://github.com/tableau/server-client-python/blob/master/tableauserverclient/server/endpoint/exceptions.py#L17

So let's take this fix now, and we can take some more time to better handle the report from #514

### DIFF
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -1,4 +1,4 @@
-from .exceptions import ServerResponseError, InternalServerError
+from .exceptions import ServerResponseError, InternalServerError, NonXMLResponseError
 from functools import wraps
 from xml.etree.ElementTree import ParseError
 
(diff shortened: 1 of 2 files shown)


tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -1,4 +1,4 @@
-from .exceptions import ServerResponseError, InternalServerError
+from .exceptions import ServerResponseError, InternalServerError, NonXMLResponseError
```
RussTheAerialist: Can you add a test for the parse error explicitly, so that this missing import is tested?

t8y8: Done!

PR #517, opened 2019-10-29
Fix filename behavior and refactor

t8y8:
There was a subtle bug where a non-folder filename would clobber the extension, for mass-download scripts this makes it hard to know what file you've downloaded (I rename to random guids to avoid name collisions).

This fixes it, and also adds tests for filesystem helpers.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ setup(
     ],
     tests_require=[
         'requests-mock>=1.0,<2.0',
-        'pytest'
+        'pytest',
+        'mock'
     ]
 )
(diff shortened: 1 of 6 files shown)


RussTheAerialist:
Should we drop python 2.7 support since it's EOL at the end of the year?

RussTheAerialist:
🚀 btw

t8y8:
@RussTheAerialist I would love to!

We could officially do it the next release, or maybe in Jan, with that being last release before back compat to 2.7 breaks. Either way works for me, once we do though, I'm going to start looking at adding some mypy types :)

PR #520, opened 2019-11-05
delete docs folder from master

jacalata:
These should only show up in the gh-pages branch


dzucker-tab:
Suggest that, as a more robust breadcrumb for those looking to contribute to docs) we leave the docs folder with the current readme.md (which points to gh-pages branch) only.

PR #523, opened 2019-11-11
Add webhooks

RussTheAerialist:
Adding webhooks for the existing api so that we it's available in the next release.


RussTheAerialist:
Just realized this isn't complete. Don't review it until I update it.

tableauserverclient/models/webhook_item.py
```diff
@@ -0,0 +1,88 @@
...
+            self.name = name
+        if url:
+            self.url = url
+        if event:
+            self.event = event
+        if owner_id:
+            self.owner_id = owner_id
+
+    @property
+    def id(self): return self._id
```
t8y8: new line here please :P

tableauserverclient/server/endpoint/webhooks_endpoint.py
```diff
@@ -0,0 +1,53 @@
...
+    @property
+    def baseurl(self):
+        return "{0}/sites/{1}/webhooks".format(self.parent_srv.baseurl, self.parent_srv.site_id)
+
+    @api(version="3.6")
+    def get(self, req_options=None):
+        logger.info('Querying all Webhooks on site')
+        url = self.baseurl
+        server_response = self.get_request(url, req_options)
+        all_workbook_items = WebhookItem.from_response(server_response.content, self.parent_srv.namespace)
```
t8y8: all_workbook_items -> all_webhook_items

t8y8: And below as well

tableauserverclient/server/request_factory.py
```diff
@@ -555,6 +555,23 @@ def empty_req(self, xml_request):
...
+    def create_req(self, xml_request, webhook_item):
+        webhook = ET.SubElement(xml_request, 'webhook')
+        webhook.attrib['name'] = webhook_item.name
+
+        source = ET.SubElement(webhook, 'webhook-source')
+        event = ET.SubElement(source, webhook_item._event)
+
+        destination = ET.SubElement(webhook, 'webhook-destination')
+        post = ET.SubElement(destination, 'webhook-destination-http')
+        post.attrib['method'] = 'POST'
```
t8y8: Is there ever a need to make this configurable? Or will it always be POST?

If the latter, this is fine, otherwise we might want a `method` attribute on the model and pass through

RussTheAerialist:
The current api is not going to change because it’s based on a bad design
that was too over-designed for features we decided to include. The method
will always be POST.

My next change will introduce the ability to use the new services endpoint
if the server supports it, but I need to get the services endpoint for
webhooks stood up first.

PR #530, opened 2019-11-19
Adds max_age param to csv and pdf request options

shinchris:
Max_age query parameter was recently added to the REST API endpoints for query view data/pdf and query workbook pdf. This PR adds max_age param to the csv and pdf request options.
It was already in place for view images.

Will make doc updates separately for that branch.

### DIFF
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -99,6 +99,7 @@ class _FilterOptionsBase(RequestOptionsBase):
 
 class CSVRequestOptions(_FilterOptionsBase):
     def __init__(self, maxage=None):
+        super(CSVRequestOptions, self).__init__()
         self.max_age = maxage
 
     def apply_query_params(self, url):
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -44,17 +44,17 @@ class RequestOptions(RequestOptionsBase):
         Asc = 'asc'
 
     def __init__(self, pagenumber=1, pagesize=100):
-        self.pagenumber = pagenumber
-        self.pagesize = pagesize
+        self.pagenumber = to_int("pagenumber", pagenumber)
+        self.pagesize = to_int("pagesize", pagesize)
         self.sort = set()
         self.filter = set()
 
     def page_size(self, page_size):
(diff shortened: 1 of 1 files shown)


tableauserverclient/server/request_options.py
```diff
@@ -159,6 +166,9 @@ def apply_query_params(self, url):
         if self.orientation:
             params.append('orientation={0}'.format(self.orientation))
 
+        if self.max_age:
+            params.append('maxAge={0}'.format(self.max_age))
```
RussTheAerialist: Should we do some verification that this is an integer at least?

tableauserverclient/server/request_options.py
```diff
@@ -146,10 +152,11 @@ class Orientation:
         Portrait = "portrait"
         Landscape = "landscape"
 
-    def __init__(self, page_type=None, orientation=None):
+    def __init__(self, page_type=None, orientation=None, maxage=None):
```
t8y8: Can we call it `max_age` here as well since it’s what you call it on the object anyway

shinchris: ImageRequestOptions already had maxage, so I wanted to keep it consistent across the 3 request objects without introducing any breaking changes. I guess we could deprecate it if we really wanted to change it?

shinchris:
oops i made the change in a different branch.

tableauserverclient/server/request_options.py
```diff
@@ -173,3 +173,11 @@ def apply_query_params(self, url):
         self._append_view_filters(params)
 
         return "{0}?{1}".format(url, '&'.join(params))
+
+
+def to_int(name, value):
+    try:
+        return int(value)
+    except ValueError:
+        print("Integer value expected for the '{}' parameter, but got {}".format(name, value))
```
RussTheAerialist: I don't think we want to just print something in this case.  We should throw an exception that tells what the error is. I believe we have something for Integer parameter errors

shinchris: We have property decorators in models for Integer parameter errors, but didn't want to bring that into here.
Above snippet prints a message and re-throws the same ValueError. It would be ideal to throw an exception with our custom message, but that results in 2 exceptions being thrown.

PR #532, opened 2019-11-23
Webhooks fixes

jacalata:
- add 'delete' argument to sample'
- actually return the value from /test


tableauserverclient/server/endpoint/webhooks_endpoint.py
```diff
@@ -60,3 +60,4 @@ def test(self, webhook_id):
         url = "{0}/{1}/test".format(self.baseurl, webhook_id)
         testOutcome = self.get_request(url)
         logger.info('Testing webhook (ID: {0} returned {1})'.format(webhook_id, testOutcome))
+        return testOutcome
```
shinchris: Just came across this while verifying the new endpoint:
I think we should parse the response, create a new WebhookTestResultItem, and return that rather than returning the raw http response. Maybe as a subclass within the webhookItem?

PR #534, opened 2019-11-25
Webhooks fixes

jacalata:
Add tests and samples for webhooks, plus some info for contributors


shinchris:
Could you change the target branch to development instead of master?

jacalata:
oops - I already had a PR into development https://github.com/tableau/server-client-python/pull/532

PR #538, opened 2019-12-10
Add support for materializeViews as schedule and task type

guodah:
Currently, the schedule and task APIs only support one type of schedule and task which is extractRefresh. The word is hardcoded in the library. The changes in this PR is to make them more general so that the caller can decide what type of schedule and task is to be used. The changes still make extractRefresh the default type so that it will not cause problems to anyone that is already using the existing APIs.

Examples:
1) Currently, `server.tasks.get()` gets all the extractRefresh tasks from a site. This will still work as is. But now users can call `server.tasks.get(task_type="materializedViews"` to get all the materializeViews tasks from a site.

2) Currently, `server.schedules.add_to_schedule(schedule_id, target_workbook)` will add a workbook to extractRefresh schedule. Again, this will remain working. But now users can call `server.schedules.add_to_schedule(schedule_id, target_workbook, schedule_type="materializeViews")` to add a workbook to a materializeViews schedule.

The API version is changed to 3.8 for Tasks.get(...) in that the new `GET /sites/site-id/tasks/materializedViews` API will be under version 3.8.


tableauserverclient/models/task_item.py
```diff
@@ -3,12 +3,18 @@
...
-    def __init__(self, id_, task_type, priority, consecutive_failed_count=0, schedule_id=None, target=None):
+    def __init__(self, id_, task_type, priority, consecutive_failed_count=0, schedule_id=None,
+                 schedule_name=None, schedule_frequency=None, schedule_next_run_at=None,
+                 schedule_last_run_at=None, target=None):
         self.id = id_
         self.task_type = task_type
         self.priority = priority
         self.consecutive_failed_count = consecutive_failed_count
         self.schedule_id = schedule_id
+        self.schedule_name = schedule_name
```
shinchris: Since these are all schedule properties, could you have one schedule_item here instead? You can leave schedule_id there, and I can create an issue to deprecate that.

tableauserverclient/models/task_item.py
```diff
@@ -20,20 +26,31 @@ def from_response(cls, xml, ns):
         parsed_response = ET.fromstring(xml)
         all_tasks_xml = parsed_response.findall(
             './/t:task/t:extractRefresh', namespaces=ns)
+        all_tasks_xml.extend(parsed_response.findall(
```
shinchris: I think it would be better to pass in the task type here and parse just the one type.

tableauserverclient/server/endpoint/schedules_endpoint.py
```diff
@@ -68,12 +68,12 @@ def create(self, schedule_item):
         return new_schedule
 
     @api(version="2.8")
-    def add_to_schedule(self, schedule_id, workbook=None, datasource=None):
+    def add_to_schedule(self, schedule_id, workbook=None, datasource=None, schedule_type='extractRefresh'):
```
shinchris: All of these schedule_type/task_type arguments should use enums rather than raw strings.

tableauserverclient/models/task_item.py
guodah: Add schedule_item there, removed other schedule_... variables, except schedule_id. Thanks for the comment!

tableauserverclient/models/task_item.py
guodah: Changed it as suggested.

tableauserverclient/models/task_item.py
```diff
@@ -1,39 +1,52 @@
...
-        schedule_element = element.find('.//t:schedule', namespaces=ns)
+        last_run_at = None
         workbook_element = element.find('.//t:workbook', namespaces=ns)
         datasource_element = element.find('.//t:datasource', namespaces=ns)
-        if schedule_element is not None:
-            schedule = schedule_element.get('id', None)
+        last_run_at_element = element.find('.//t:lastRunAt', namespaces=ns)
+
+        schedule_item_list = ScheduleItem.from_element(element, ns)
+        if len(schedule_item_list) >= 1:
```
shinchris: Just out of curiosity, can a task ever return more than 1 schedule?

tableauserverclient/server/endpoint/schedules_endpoint.py
guodah: Added TaskItem.Type. 

Here is an inconsistency. When calling [get_extract_refresh_tasks](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#get_extract_refresh_tasks), the word for extract refresh is `extractRefreshes". However, the word in the request body for [add_workbook_to_schedule ](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#add_workbook_to_schedule) is extractRefresh. I added a normalize function in tasks_endpoint.py to deal with this inconsistency.

tableauserverclient/models/task_item.py
guodah: No, a task can have only one schedule. Here is an example: https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#add_workbook_to_schedule

PR #542, opened 2019-12-12
Add support for materializeViews as schedule and task type

guodah:
Currently, the schedule and task APIs only support one task type which is extractRefresh. The word is hardcoded in the library. The changes in this PR is to make them more general so that the caller can decide what type of schedule and task is to be used. The changes still make extractRefresh the default type so that it will not cause problems to anyone that is already using the existing APIs.

Examples:

- Currently, server.tasks.get() gets all the extractRefresh tasks from a site. This will still work as is. But now users can call server.tasks.get(task_type=TaskItem.Type.MaterializeViews to get all the materializeViews tasks from a site.
- Currently, server.schedules.add_to_schedule(schedule_id, target_workbook) will add a workbook to extractRefresh schedule. Again, this will remain working. But now users can call server.schedules.add_to_schedule(schedule_id, target_workbook, task_type=TaskItem.Type.MaterializeViews) to add a workbook to a materializeViews task to schedule.

The API version is changed to 3.8 for Tasks.get(...) in that the new GET /sites/site-id/tasks/materializedViews API will be under version 3.8.


tableauserverclient/server/endpoint/tasks_endpoint.py
```diff
@@ -9,26 +9,41 @@
...
+        """
+            The word for extract refresh used in API URL is "extractRefreshes".
+            It is different than the tag "extractRefresh" used in the request body.
+        """
+        if task_type == TaskItem.Type.ExtractRefresh:
+            return '{}es'.format(task_type)
+        else:
+            return task_type
+
+    @api(version='3.8')
```
shinchris: We don't want to change the version, because this endpoint works for extractRefreshes for version 2.6+. I think the best way for now is to call assert_at_least_version() from the parent server class if tasktype is materalizeviews.

tableauserverclient/server/endpoint/tasks_endpoint.py
```diff
@@ -38,17 +53,19 @@ def run(self, task_item):
...
+        url = "{0}/{1}/{2}/runNow".format(self.baseurl,
+                                          self.__normalize_task_type(TaskItem.Type.ExtractRefresh), task_item.id)
         run_req = RequestFactory.Task.run_req(task_item)
         server_response = self.post_request(url, run_req)
         return server_response.content
 
     # Delete 1 task by id
-    @api(version="3.6")
-    def delete(self, task_id):
+    @api(version="3.8")
```
shinchris: Same comment from above.

tableauserverclient/server/endpoint/tasks_endpoint.py
guodah: Thanks a lot for pointing it out! Changed it as advised.

tableauserverclient/server/endpoint/tasks_endpoint.py
guodah: Changed as advised. thanks!

shinchris:
🚀  Looks good to me!

PR #544, opened 2019-12-18
Fix minor bug in request factory

guodah:
Currently, the materialized views setting for a workbook can only be updated when materialized_views_enabled and run_materialization_now are both true. This prevents the workbooks from enabled and disabled correctly. The fix is to check if the materialized view config is not null and contains materialized_views_enabled and run_materialization_now.

### DIFF
--- a/tableauserverclient/server/request_factory.py
+++ b/tableauserverclient/server/request_factory.py
@@ -487,8 +487,9 @@ class WorkbookRequest(object):
             materialized_views_element = ET.SubElement(workbook_element, 'materializedViewsEnablementConfig')
             materialized_views_element.attrib['materializedViewsEnabled'] = str(materialized_views_config
                                                                                 ["materialized_views_enabled"]).lower()
-            materialized_views_element.attrib['materializeNow'] = str(materialized_views_config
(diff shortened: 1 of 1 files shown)


tableauserverclient/server/request_factory.py
```diff
@@ -480,8 +480,9 @@ def update_req(self, workbook_item):
         if workbook_item.owner_id:
             owner_element = ET.SubElement(workbook_element, 'owner')
             owner_element.attrib['id'] = workbook_item.owner_id
-        if workbook_item.materialized_views_config['materialized_views_enabled'] \
-                and workbook_item.materialized_views_config['run_materialization_now']:
+        if workbook_item.materialized_views_config is not None and \
```
shinchris: Do both 'materialized_views_enabled' and 'run_materialization_now' fields have to be present to be able to update MaterializedViewsEnablementConfig for workbooks?

guodah: Thanks for pointing it out! Changed "run_materialization_now" to optional which is consistent with the server side code.

shinchris:
🚀

PR #545, opened 2019-12-24
Fix minor bug in request factory due to bool(time(0,0))==False in Python 3.4

guodah:
Currently, it will fail if someone tries to create a schedule with an end_time being time(0,0) using Python 3.4. Looks there is a bug in Python 3.4 (https://stackoverflow.com/questions/28116931/datetime-time0-0-evaluates-as-false-in-boolean-context) which evaluates bool(time(0,0)) to False. 

But https://tableau.github.io/server-client-python/docs/ states it supports Python 3.3 or higher.

The fix is just to explicitly check if end_time is None.


shinchris:
Could you also add the change to line 305, in the update request?
Otherwise, looks good 🚀

PR #551, opened 2020-01-16
Add warnings when adding workbook to schedule

guodah:
The recent changes in the server side code may add warnings when add a workbook to a schedule. Here is an example response with such warnings:

<tsResponse>
    <warnings>
        <warning message="Workbook with id 29,354 uses only embedded extract(s) and does not need explicit scheduling for acceleration. It was enabled for acceleration but not attached to the given schedule."/>
    </warnings>
</tsResponse>

This PR add client code for receiving such warnings.

### DIFF
--- a/tableauserverclient/server/endpoint/schedules_endpoint.py
+++ b/tableauserverclient/server/endpoint/schedules_endpoint.py
@@ -8,8 +8,8 @@ from collections import namedtuple
 
 logger = logging.getLogger('tableau.endpoint.schedules')
 # Oh to have a first class Result concept in Python...
-AddResponse = namedtuple('AddResponse', ('result', 'error', 'warnings'))
-OK = AddResponse(result=True, error=None, warnings=None)
+AddResponse = namedtuple('AddResponse', ('result', 'error', 'warnings', 'task_created'))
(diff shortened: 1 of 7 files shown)


tableauserverclient/server/endpoint/schedules_endpoint.py
```diff
@@ -70,15 +71,24 @@ def create(self, schedule_item):
     @api(version="2.8")
     def add_to_schedule(self, schedule_id, workbook=None, datasource=None,
                         task_type=TaskItem.Type.ExtractRefresh):
+        def read_warnings(response):
```
shinchris: Can this be moved out of the endpoint class? I thought I saw it in the last MR.

guodah: Thanks for the comment! Yeah, I thought about it. The same code was in ScheduleItem in the last MR. I was not sure I should factor it out to a static function in ScheduleItem for reading warnings. It seems reading warnings should be common to all rest api calls. But I did not find a good place for this kind of "common" things. Ideas?

shinchris: Thanks for moving it to the item object

shinchris:
Looks good to me 🚀

Issue #562, opened 2020-02-05
server.workbooks.refresh not accepting the correct parameter


nicoleLaczny:
So my code is refreshing data sources and workbooks, I am able to refresh the data source with no issues doing: 
`all_datasources, pagination_item = server.datasources.get()
        try:
            location = [datasource.name for datasource in all_datasources].index(dataSourceName)
            datasource = all_datasources[location]
            refreshData = server.datasources.refresh(datasource)`

BUT applying the same logic to:
`all_workbooks, pagination_item = server.workbooks.get()
try:
            location = [workbook.name for workbook in all_workbooks].index(workbookName)
            workbook = all_workbooks[location]
            workbook2 = server.workbooks.refresh(workbook)
`

i get this error: 
`404006: Resource Not Found
                Workbook '<tableauserverclient.models.workbook_item.WorkbookItem object at 0x000002970D233358>' could not be found.`

and when looking at `workbook = server.workbooks.refresh()`  this happens: `TypeError: refresh() missing 1 required positional argument: 'workbook_id'`

but in the documentation (https://tableau.github.io/server-client-python/docs/api-ref#workbooks) for workbooks refresh, the pass in parameter should be **workbook_item**

why are the arguments different? shouldn't server.workbooks.refresh() require **workbook_item** not **'workbook_id'**?

jorwoods:
@nicoleLaczny I ran into exactly the same issue. I just created #563 to resolve this problem while still allowing backward compatibility.

jorwoods:
@nicoleLaczny also, I found that calling `server.workbooks.get` or `server.datasources.get` doesn't necessarily return all items, only the first 100, as it paginates.

What you should do instead is something like:
`all_workbooks = list(TSC.Pager(server.workbooks))`

PR #563, opened 2020-02-06
Allow wb/ds item or id in refresh

jorwoods:
Allows refresh method on datasource and workbook endpoints to accept either the object or their id. Allows for a flexible and consistent api.


jorwoods:
I had an issue similar to #562 and this change should allow both the datasource and workbook refresh calls to accept either the item or their id. This also should be a non-breaking change for anyone's existing code.

jorwoods:
Closing PR to retarget development.

Issue #546, opened 2020-01-01
Datasource refresh status


s-arjun:
Hi,
I was trying to refresh datasources from the tsc package.
My Tableau Server version is **2019.3**
I used the following code:
```
selected_datasource = server.datasources.get_by_id(dat_id)
server.datasources.refresh(selected_datasource)
```
It refreshed successfully.
I was able to get all the jobs in a similar fashion:
```
request_options = tsc.RequestOptions(pagesize=50)
request_options.sort.add(tsc.Sort(tsc.RequestOptions.Field.CreatedAt, tsc.RequestOptions.Direction.Desc))
all_jobs, job_pagination_item = server.jobs.get(request_options)
```
When I give `job.status` I could get whether the job is success/in progress.
But I could not obtain more details of the job like the one shown below from the Tableau Server webUI.

![image](https://user-images.githubusercontent.com/30429495/71641693-1d016980-2cc6-11ea-920c-f8f1d20ddee3.png)

The **job content** is the one i require, that is here it is the datasource item. 
Is there a way to get this detail as of now?

jwalin:
Hi @s-arjun , from where I will get datasource_id?

jwalin:
I got it

jacalata:
This should have been addressed in https://github.com/tableau/server-client-python/pull/571

PR #564, opened 2020-02-06
Return notes on JobItem from XML response

jorwoods:
Extract notes from the xml response and return them as a property on a JobItem. Should help with #546


jorwoods:
Closing PR to retarget development.

PR #571, opened 2020-02-20
Create notes property from XML response

jorwoods:
Adds note information to JobItem objects. Should resolve #564

### DIFF
--- a/tableauserverclient/models/job_item.py
+++ b/tableauserverclient/models/job_item.py
@@ -6,7 +6,7 @@ from ..datetime_helpers import parse_datetime
 
 class JobItem(object):
     def __init__(self, id_, job_type, progress, created_at, started_at=None,
-                 completed_at=None, finish_code=0, notes=[]):
+                 completed_at=None, finish_code=0, notes=None):
         self._id = id_
         self._type = job_type
         self._progress = progress
@@ -14,7 +14,7 @@ class JobItem(object):
         self._started_at = started_at
         self._completed_at = completed_at
         self._finish_code = finish_code
-        self._notes = notes
+        self._notes = notes or []
 
     @property
     def id(self):
--- a/tableauserverclient/models/job_item.py
+++ b/tableauserverclient/models/job_item.py
@@ -5,7 +5,8 @@ from ..datetime_helpers import parse_datetime
 
 
 class JobItem(object):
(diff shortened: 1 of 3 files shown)


jorwoods:
Added a test for the notes property

t8y8:
Hey @jorwoods sorry for forever delays, open source 🤦 .

On a re-review, I realized that the note value and type can be very interesting depending on the job -- I'm thinking of cases that want to programmatically take action based on the value, which is more stable that the note text.

What do you think about a simple NoteItem class, or at least a dictionary that holds some of the other attributes on the note element?

jorwoods:
What else can the `note` element contain besides text? `statusNotes` which is a different element can have other attributes, but for the purposes of solving the problem stated in the issue, as well as the problem I was trying to solve, those were contained with the `note.text()` instead of somewhere within `statusNotes`.

t8y8:
Ah-ha, you're right, I've confused the note with the statusNote. I feel like I was PM back when we added this to the API, so I can only blame myself for the overlapping names.

tableauserverclient/models/job_item.py
```diff
@@ -5,14 +5,16 @@
 
 
 class JobItem(object):
-    def __init__(self, id_, job_type, progress, created_at, started_at=None, completed_at=None, finish_code=0):
+    def __init__(self, id_, job_type, progress, created_at, started_at=None,
+                 completed_at=None, finish_code=0, notes=[]):
```
t8y8, line 9: Set `notes` equal to `None` in the `__init__` and then if you want to to be an empty list, you can do that in the assignment to `self._notes` with `if notes is None` but I think you can just leave this as `None` generally. Up to you!

jorwoods: I set the default as an empty list so that it would always be iterable without having to check for a `NoneType` first. Ultimately I see that difference as a matter of preference. I can make a change to initialize it as `None` and return an empty iterable if it is `None`, though. That seems like an extra step to the same end. Is the preference that it not always be iterable?

t8y8: Following advice from here https://nikos7am.com/posts/mutable-default-arguments/ <- I think it might not be an issue in this particular case but it is the recommend Pythonic way to do it, and the style I generally follow.

As for iterable by default, makes sense to me for making it a `[]` after object initialization

PR #574, opened 2020-02-24
Adds long description to setup.

RussTheAerialist:
Read the long description into setup so that it updates on pypi.

This will not work under python2 which is fine since 2 is EOL.

Based on https://packaging.python.org/guides/making-a-pypi-friendly-readme/


shinchris:
🚀

PR #576, opened 2020-03-01
Replacing the feature name "materialized views" with "data acceleration"

guodah:
The feature "Materialized Views" has been replaced with "Data Acceleration" in the server code, including the rest api schema and urls. This PR is to make the name change in the client side.


shinchris:
The changes themselves look good 😄 
My only concern is that the sample might be too complex to be considered a useful sample for our users. Samples should capture specific and frequent scenarios that should be easy for a new user to refer to for examples of using the library. 
Would it be possible to break this sample into specific scenarios? Doesn't have to be part of this MR, though, since this is for 2020.2.

guodah:
Thanks for the speedy review! Our team wants to provide users with a "comprehensive" script to use our feature with different options as our feature does not have a UI yet. 

Let me discuss with my team. :)

shinchris:
Thanks @guodah !

Issue #587, opened 2020-03-20
PDFRequestOptions - Unspecified


sfarr15:
Hi, we are trying to download a number of daily Tableau reports in PDF format. Unfortunately there is no option to define the page_type as 'unspecified' using the PDFRequestOptions function - confining us to first generate an image and then convert this image to PDF (losing quality and adding to file size)
Can this be added as this would help us out greatly! 
Regards,
Steve

sfarr15:
Hi @shinchris,
This request is on behalf of comeon group. We would appreciate if this could be resolved as quickly as possible!

Steven-Comeon:
Any updates on this please? @shinchris

shinchris:
@sfarr15 @Steven-Comeon Which version of Tableau server do you use?

VitaliyBb:
Anyone having any luck finding a working around for this issue?

bhlmn:
This is also something that my team would like! Anyone have any luck with this?

sfarr15:
Hi,

We ended up generating the images using the unspecified format and converting them to pdf. 

This worked out fine although it adds an extra step.

bhlmn:
Thanks so much @sfarr15. We'll give that a shot.

jacalata:
Addressed in https://github.com/tableau/server-client-python/pull/589

PR #589, opened 2020-03-26
Unspecified PDF request option

sfarr15:
This is the only pagetype not available via tableau api for PDF. The 'Unspecified' pagetype is available for download using tableau online and desktop so I really don't see an issue adding this. Can this kindly be amended as the workaround for this requires downloading an image and converting to pdf which results in poorer quality and larger data size.

### DIFF
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -141,6 +141,7 @@ class PDFRequestOptions(_FilterOptionsBase):
         Note = "note"
         Quarto = "quarto"
         Tabloid = "tabloid"
+        Unspecified = "unspecified"
 
     class Orientation:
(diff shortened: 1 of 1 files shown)


sfarr15:
Any news on this please?

shinchris:
This would actually require a server-side change. Passing in 'unspecified' will throw an error because it's not an expected value in the REST API. I'll add this to our internal backlog and update the original issue #587 with any future updates.

jacalata (approved):

jacalata:
Confirmed that it's now available through the API so this is ready to go.

bcantoni:
@anthonysfancher since this is an older closed issue, could you please write a new issue with details of what you're seeing?

Issue #39, opened 2016-09-29
Add support for Favorites


benlower:
Once the REST API enables GET Favorites we can add complete support for Favorites.

jorwoods:
The REST API supports GET Favorites with API version 2.5.

jorwoods:
I tried to implement this myself, but ran into difficulties due to the fact that the favorites API returns views, datasources, workbooks, and projects in a single API call. Combine that with inconsistencies in each Item's `_set_values` method, it was much more difficult to implement than I was expecting. I have a branch in my repo with the attempt if anyone wants to build off of that.

jacalata:
I'm going to call this done, although I think there are now new types that can be returned as favorites. We can add that later.

PR #591, opened 2020-03-31
Support favorites retrieval and modification

jorwoods:
Solves #39 

This may not be the intended implementation, or be how it was intended to be solved. I took a pass at it to give myself a route to modify favorites for my users. After I wrote this, I noticed stubs for it in the users endpoint, but wasn't sure if that was still the plan as favorites has a separate base url. I haven't written extensive tests for this yet.

Also wasn't sure if "Favorites" technically should have their own model, or just an endpoint. Sort of ambiguous as a favorites object would just be a wrapper around a workbook/view/project/datasource.

### DIFF
--- a/tableauserverclient/models/user_item.py
+++ b/tableauserverclient/models/user_item.py
@@ -105,7 +105,7 @@ class UserItem(object):
         if self._favorites is None:
             error = "User item must be populated with favorites first."
             raise UnpopulatedPropertyError(error)
-        return self._favorites()
+        return self._favorites
 
     def to_reference(self):
         return ResourceReference(id_=self.id, tag_name=self.tag_name)
--- a/tableauserverclient/models/user_item.py
+++ b/tableauserverclient/models/user_item.py
@@ -42,6 +42,7 @@ class UserItem(object):
         self._id = None
         self._last_login = None
         self._workbooks = None
+        self._favorites = None
         self.email = None
         self.fullname = None
         self.name = name
@@ -98,6 +99,13 @@ class UserItem(object):
             error = "User item must be populated with workbooks first."
             raise UnpopulatedPropertyError(error)
         return self._workbooks()
+
+    @property
+    def favorites(self):
+        if self._favorites is None:
+            error = "User item must be populated with workbooks first."
+            raise UnpopulatedPropertyError(error)
+        return self._favorites()
 
     def to_reference(self):
         return ResourceReference(id_=self.id, tag_name=self.tag_name)
--- a/tableauserverclient/models/user_item.py
+++ b/tableauserverclient/models/user_item.py
@@ -103,7 +103,7 @@ class UserItem(object):
     @property
     def favorites(self):
         if self._favorites is None:
-            error = "User item must be populated with workbooks first."
+            error = "User item must be populated with favorites first."
             raise UnpopulatedPropertyError(error)
         return self._favorites()
 
--- a/tableauserverclient/server/endpoint/favorites_endpoint.py
+++ b/tableauserverclient/server/endpoint/favorites_endpoint.py
@@ -1,6 +1,7 @@
 from .endpoint import Endpoint, api
 from .exceptions import MissingRequiredFieldError
-from .. import RequestFactory, WorkbookItem, ViewItem, ProjectItem, DatasourceItem
+from .. import RequestFactory
+from ...models import get_favorites
 from ..pager import Pager
 import xml.etree.ElementTree as ET
 import logging
@@ -20,30 +21,8 @@ class Favorites(Endpoint):
         logger.info('Querying all favorites for user {0}'.format(user_item.name))
         url = '{0}/{1}'.format(self.baseurl, user_item.id)
         server_response = self.get_request(url, req_options)
-        parsed_response = ET.fromstring(server_response.content)
-        favorites = []
-        for workbook in parsed_response.findall('.//t:favorite/t:workbook', self.parent_srv.namespace):
-            fav_workbook = WorkbookItem('')
-            fav_workbook._set_values(*fav_workbook._parse_element(workbook, self.parent_srv.namespace))
-            if fav_workbook:
-                favorites.append(fav_workbook)
-        for view in parsed_response.findall('.//t:favorite[t:view]', self.parent_srv.namespace):
-            fav_view = ViewItem()
-            fav_view.from_xml_element(view, self.parent_srv.namespace)
-            if fav_view:
-                favorites.append(fav_view)
-        for datasource in parsed_response.findall('.//t:favorite/t:datasource', self.parent_srv.namespace):
-            fav_datasource = DatasourceItem('')
-            fav_datasource._set_values(*fav_datasource._parse_element(datasource, self.parent_srv.namespace))
-            if fav_datasource:
-                favorites.append(fav_datasource)
-        for project in parsed_response.findall('.//t:favorite/t:project', self.parent_srv.namespace):
-            fav_project = ProjectItem('p')
-            fav_project._set_values(*fav_project._parse_element(project))
-            if fav_project:
-                favorites.append(fav_project)
 
-        return favorites
+        user_item._favorites = get_favorites(server_response, self.parent_srv.namespace)
 
     @api(version="2.0")
     def add_favorite_workbook(self, user_item, workbook_item):
--- a/tableauserverclient/server/endpoint/favorites_endpoint.py
+++ b/tableauserverclient/server/endpoint/favorites_endpoint.py
@@ -20,8 +20,6 @@ class Favorites(Endpoint):
         logger.info('Querying all favorites for user {0}'.format(user_item.name))
         url = '{0}/{1}'.format(self.baseurl, user_item.id)
         server_response = self.get_request(url, req_options)
-        # pagination_item = PaginationItem.from_response(server_response.content, self.parent_srv.namespace)
-        # all_favorite_items = UserItem.from_response(server_response.content, self.parent_srv.namespace)
         parsed_response = ET.fromstring(server_response.content)
         favorites = []
         for workbook in parsed_response.findall('.//t:favorite/t:workbook', self.parent_srv.namespace):
--- a/tableauserverclient/server/endpoint/favorites_endpoint.py
+++ b/tableauserverclient/server/endpoint/favorites_endpoint.py
@@ -23,25 +23,25 @@ class Favorites(Endpoint):
         parsed_response = ET.fromstring(server_response.content)
         favorites = []
         for workbook in parsed_response.findall('.//t:favorite/t:workbook', self.parent_srv.namespace):
-            i = WorkbookItem('')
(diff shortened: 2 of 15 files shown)


tableauserverclient/models/user_item.py
```diff
@@ -98,6 +99,13 @@ def workbooks(self):
             raise UnpopulatedPropertyError(error)
         return self._workbooks()
 
+    @property
+    def favorites(self):
+        if self._favorites is None:
+            error = "User item must be populated with workbooks first."
```
t8y8, line 105: The string should say something like "favorites" or "content" since they can be more than just Workbooks

tableauserverclient/server/endpoint/favorites_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+
+logger = logging.getLogger('tableau.endpoint.favorites')
+
+
+class Favorites(Endpoint):
+    @property
+    def baseurl(self):
+        return "{0}/sites/{1}/favorites".format(self.parent_srv.baseurl, self.parent_srv.site_id)
+
+    # Gets all users
```
t8y8, line 17: copy/paste error?

tableauserverclient/server/endpoint/favorites_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+    @api(version="2.5")
+    def get(self, user_item, req_options=None):
+        logger.info('Querying all favorites for user {0}'.format(user_item.name))
+        url = '{0}/{1}'.format(self.baseurl, user_item.id)
+        server_response = self.get_request(url, req_options)
+        # pagination_item = PaginationItem.from_response(server_response.content, self.parent_srv.namespace)
```
t8y8, line 23: remove

tableauserverclient/server/endpoint/favorites_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+        # all_favorite_items = UserItem.from_response(server_response.content, self.parent_srv.namespace)
+        parsed_response = ET.fromstring(server_response.content)
+        favorites = []
+        for workbook in parsed_response.findall('.//t:favorite/t:workbook', self.parent_srv.namespace):
+            i = WorkbookItem('')
```
t8y8, line 28: Use something more descriptive, like `new_workbook` or `fav_workbook`

tableauserverclient/server/endpoint/favorites_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+            i._set_values(*i._parse_element(workbook, self.parent_srv.namespace))
+            if i:
+                favorites.append(i)
+        for view in parsed_response.findall('.//t:favorite[t:view]', self.parent_srv.namespace):
+            i = ViewItem()
```
t8y8, line 33: ditto

tableauserverclient/server/endpoint/favorites_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+            i.from_xml_element(view, self.parent_srv.namespace)
+            if i:
+                favorites.append(i)
+        for datasource in parsed_response.findall('.//t:favorite/t:datasource', self.parent_srv.namespace):
+            i = DatasourceItem('')
```
t8y8, line 38: ditto

tableauserverclient/server/endpoint/favorites_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+            i._set_values(*i._parse_element(datasource, self.parent_srv.namespace))
+            if i:
+                favorites.append(i)
+        for project in parsed_response.findall('.//t:favorite/t:project', self.parent_srv.namespace):
+            i = ProjectItem('p')
```
t8y8, line 43: ditto

tableauserverclient/server/endpoint/favorites_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+            i._set_values(*i._parse_element(project))
+            if i:
+                favorites.append(i)
+
+        return favorites
+
+    @api(version="1.0")
```
t8y8, line 50: Minimum version is 2.0

tableauserverclient/server/request_factory.py
```diff
@@ -145,6 +145,34 @@ def publish_req_chunked(self, datasource_item, connection_credentials=None, conn
...
+        xml_request = ET.Element('tsRequest')
+        favorite_element = ET.SubElement(xml_request, 'favorite')
+        target = ET.SubElement(favorite_element, target_type)
+        favorite_element.attrib['label'] = label
+        target.attrib['id'] = id_
+
+        return ET.tostring(xml_request)
+
+    def add_datasource_req(self, id_, name):
+        return self._add_to_req(id_, 'datasource', name)
```
t8y8, line 164: For permissions content types, where we have a similar pattern, we use a (fake) enum for these types.
For grantees, we pull the type from the model itself with a 'tag_name' property that extracts from the XML.

I think here the former ( an enum ) is a little better than the raw strings

tableauserverclient/models/user_item.py
jorwoods: Made the object name more generic.

tableauserverclient/server/endpoint/favorites_endpoint.py
jorwoods: Yes. Corrected.

tableauserverclient/server/endpoint/favorites_endpoint.py
jorwoods: Commented lines were removed.

tableauserverclient/server/endpoint/favorites_endpoint.py
jorwoods: Made the variable name more descriptive.

tableauserverclient/server/endpoint/favorites_endpoint.py
jorwoods: Made the variable name more descriptive.

tableauserverclient/server/endpoint/favorites_endpoint.py
jorwoods: Made the variable name more descriptive.

tableauserverclient/server/endpoint/favorites_endpoint.py
jorwoods: Made the variable name more descriptive.

tableauserverclient/server/endpoint/favorites_endpoint.py
jorwoods: Updated the API version number.

tableauserverclient/server/request_factory.py
jorwoods: Made an enum like class, and used that instead. Let me know if that is not what you intended.

tableauserverclient/server/endpoint/favorites_endpoint.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class Favorites(Endpoint):
+    @property
+    def baseurl(self):
+        return "{0}/sites/{1}/favorites".format(self.parent_srv.baseurl, self.parent_srv.site_id)
+
+    # Gets all favorites
+    @api(version="2.5")
+    def get(self, user_item, req_options=None):
```
t8y8, line 19: Thanks for the quick turnaround!

Revisiting your question about should there be a favorite_item class: I think so.

This xml parsing logic is fairly harmless here, but if we ever swapped to JSON or something like that, we'd be leaking that into the endpoint, rather than keeping it contained in just the models.

I can think of 3 ways to do this:
1. Have a FavoriteItem class per content type `WorkbookFavoriteItem.from_request` etc
2. Have a single `FavoriteItem` with 4 parsing methods `FavoriteItem.workbooks_from_request`
3. Have a single `FavoriteItem` with 1 parsing method that returns a list of lists, or a dict of lists, or a series of tuples, `FavoriteItem.from_request`

@jorwoods I'm not sure which is best, do you have any thoughts? (You may look to TaskItem for inspiration, I think it has a similar wrapper pattern)

jorwoods: Agreed that it likely need to be pulled out of endpoint. The way I'm calling out to the class methods for each of the types feels flimsy too, but I could not see a better path that didn't involve altering a few of the classes substantially.

1.  This feels like it would fracture the favorites endpoint substantially. It also wouldn't reflect the REST API that returns all objects. May be preferable though? May be the clearest option as a python user decides they only want favorites of a certain type. REST API though would return everything, and the object would just filter it down.

2. Again, the REST API returns all objects, so while having individual dispatch is clearer from a python standpoint, there would have to be something calling out to each method.

3. This would probably be my preference, with a dict of lists.  Would sort the objects into easily addressable sets, while also being consistent with the behavior of the REST API.

t8y8: #3 sounds best to me as well.

(You could also have a favorites object with properties for each. This is what Permissions wraps as a list of PermissionRules... but this case is simpler and that'd be no better than a dict.)

So...

```
class FavoritesList(TypedDict):
  workbooks: List[WorkbookItem]
  flows: List[FlowItem]
  datasources: List[DatasourceItem]
  views: List[Views]
```

jorwoods: To clarify, because of the snippet you posted: You have `FavoritesList` inherit from TypedDict. Should I put in type hinting? Or was that purely for demonstrative purposes?

jorwoods: And this invites other questions: Would it be best for them to be accessed via __getattr__ or __getitem__? A simple enough hack could make it both.

If a user has no favorites of a given item, should it return None, or an empty list? Alternatively it could just not be present on the object, but I think None or empty list is better.

t8y8: Just illustrative, we don't have type hinting in the code base yet.

If it wasn't clear, I think a simple dict is the way to go. No need for the wrapper object.

`None` or `[]` ... I like the idea of an empty list so for loops work just as well. I don't think we're always consistent here throughout the library though.

```
for fav in favorites.workbooks:
    print(fav.id)
    # if empty nothing happens, but no error
```

jorwoods:
Added testing, and made the favorites load in as a dictionary attached to the user item.

tableauserverclient/server/endpoint/favorites_endpoint.py
jorwoods: Marking this as resolved as it now has tests and writes to a dictionary as requested.

jorwoods:
Rebased on current development branch

Fixed code style issues
Readded Favorites import
Removed deprecated calls from unrelated tests

jorwoods:
@t8y8 Anything else need to change?

Issue #590, opened 2020-03-28
URI Resource Not Found When Update Project Default Permission for Workbook and Datasource


doulam:
Hi,

The URI for updating the project default permission for Workbook and Datasource sections is incorrect. I got the following error:

	404003: Resource Not Found
		Unknown resource '/3.6/sites/cf5e5d82-ddba-4df4-b595-59db361ab0db/projects/b0ee5d45-6ed3-4514-b6b1-8fcd92b3969b/default-permissions/workbook' specified in URI.

It seems like an "s" is missing in the URI for **Workbook** and **Datasource**.

The correct URI are: 

/api/api-version/sites/site-id/projects/project-id/default-permissions/**datasources**
/api/api-version/sites/site-id/projects/project-id/default-permissions/**workbooks**

PR #592, opened 2020-03-31
Fix default permission uri

doulam:
fix for issue #590


doulam:
@shinchris Any idea on the ETA for this PR to get reviewed, approved and merged?

tableauserverclient/server/endpoint/default_permissions_endpoint.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, parent_srv, owner_baseurl):
         self.owner_baseurl = owner_baseurl
 
     def update_default_permissions(self, resource, permissions, content_type):
-        url = '{0}/{1}/default-permissions/{2}'.format(self.owner_baseurl(), resource.id, content_type)
+        url = '{0}/{1}/default-permissions/{2}'.format(self.owner_baseurl(), resource.id, content_type + 's')
```
shinchris, line 30: I think it would be better to append the 's' to the 5 Permission.Resource enums defined in permissions.py, rather than adding it to the URL here (@t8y8 FYI). Pluralization is also needed for the populate and delete methods, so changing the enum would make be cleaner.

t8y8: There was some reason I did this here in the URLs and not in the enum, I don't remember exactly why, but I think it's because the enum gets used elsewhere where you don't want the `s`

shinchris: Not sure if it is being used elsewhere currently, but yeah it could be in the future. Within the domain of permissions only plural resource names are used, hence the suggestion, but I'm okay with this.

t8y8: I'm ok either way, I think I lean slightly towards the 's' being part of the url logic and not the resource name itself, but not sure I care a ton.

Will this create a merge conflict with the View Permissions PR? Should that one go in first?

shinchris: I don't think it will create a merge conflict. I'm waiting on the CLA for this one.

t8y8: I rethought it, and I think I changed my mind, maybe the enum should have the plural form.

Let's open a new item on that to do them all at once though

shinchris: Sorry to keep going back and forth on this, but looking more into this, making the enums plural will require more changes than I originally thought. Especially with the _set_default_permissions() method in project_item.py that sets fields by their name. 

I'll resolve this conversation. This PR is good to be merged whenever the CLA has been signed 🚀

qingfei1994:
@shinchris @doulam  I'm wondering when this could be merged and released as I kind of need this fix to update default permission of workbooks in my program..... I could fix it if @doulam don't have time to sign the CLA.......

shinchris:
We have PR #603 that includes this fix and a couple more fixes, as well as some samples.

PR #596, opened 2020-04-07
Adding DataAccelerationReport end point and item

guodah:
On the server side, a new end point dataAccelerationReport has been added. It returns the page loading time before and after certain sheets were accelerated. It will only show such comparison for pages that are accelerated AND loaded by users.

This PR contains the corresponding changes on the client side.


tableauserverclient/server/endpoint/data_acceleration_report_endpoint.py
```diff
@@ -0,0 +1,30 @@
...
+        super(DataAccelerationReport, self).__init__(parent_srv)
+
+        self._permissions = _PermissionsEndpoint(parent_srv, lambda: self.baseurl)
+        self._default_permissions = _DefaultPermissionsEndpoint(parent_srv, lambda: self.baseurl)
+
+    @property
+    def baseurl(self):
+        return "{0}/sites/{1}/dataAccelerationReport".format(self.parent_srv.baseurl, self.parent_srv.site_id)
+
+    @api(version="3.8")
```
shinchris, line 23: Just noticed this here and on the server side. Is this new for 2020.3? 2020.3 is v3.9.

guodah: Thanks for checking! This is actually for 2020.2. :)

PR #603, opened 2020-04-24
Fixes default permissions urls and adds return value to update endpoints

shinchris:
Default permissions urls for update and delete were missing an 's' after the content types. Also noticed that the update permission methods weren't returning the response from server, while the delete methods were unnecessarily returning something. 

Changes:
- Corrected URLs for update and delete default permissions by appending an s to the content type
- Removed return statements for the delete methods
- Added return statements for the update methods
- Fixed delete default permissions test and added a update default permissions test

### DIFF
--- a/tableauserverclient/server/endpoint/databases_endpoint.py
+++ b/tableauserverclient/server/endpoint/databases_endpoint.py
@@ -93,7 +93,7 @@ class Databases(Endpoint):
 
     @api(version='3.5')
     def delete_permission(self, item, rules):
-        return self._permissions.delete(item, rules)
+        self._permissions.delete(item, rules)
 
     @api(version='3.5')
     def populate_table_default_permissions(self, item):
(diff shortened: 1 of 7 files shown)


shinchris:
I'm also adding some samples. Will push it as soon as it's done. 
Since we don't have docs yet, the samples should hopefully help in the meantime.

tableauserverclient/server/endpoint/databases_endpoint.py
```diff
@@ -93,15 +93,15 @@ def update_permission(self, item, rules):
 
     @api(version='3.5')
     def delete_permission(self, item, rules):
-        return self._permissions.delete(item, rules)
+        self._permissions.delete(item, rules)
```
t8y8, line 96: I reviewed the underlying methods, and I know what I was going for, but thanks for pointing out it didn't work :P

DELETE should get a 204 back from the Server, but we don't return it (https://github.com/tableau/server-client-python/blame/master/tableauserverclient/server/endpoint/endpoint.py#L89). 

I think we should though the plumbing might not be worth it, I'll take a look in another item

shinchris: It was working, I just noticed that the underlying methods weren't returning anything so I removed it here. I think returning the 204 would be nice too, since currently it's a silent operation unless theres is an error.

PR #608, opened 2020-04-28
Fix logger statement in User.add

rmagier1:
The logger statement is using the parameter to output the id of the user, but that isnt set until line 67 and saved to a new variable. We want the logger statement to use that new user because it currently always prints None. 

Issue link #609


shinchris:
Thanks for the PR @rmagier1!
Before we merge in the change, could you sign our [Contributor License Agreement](http://tableau.github.io/contributing.html)?

RussTheAerialist:
CLA Received

Issue #354, opened 2018-11-05
Hide Views when Publishing a Workbook


twozeroseven:
A workbook with worksheets set to hidden - which is not uncommon for Workbooks with 1 dashboard view and many hidden worksheets.   Upon publishing via workbook.publish() method the sheets show up on Tableau server as visible.

sks786:
Hi Team, 

I also have same concern.
When publishing workbook to the server, hidden sheets are being visible, which we don't want?

Any help on this concern.

Regards,
Subodh kumar

sks786:
Hi twozeroseven,

Did you get any solution?
If yes, please help me out.

Regards,
Subodh Kumar

twozeroseven:
Is there a better way to setup a workbook to not have hidden sheets?

irwando:
This was added to REST API, need to plumb it through:

https://onlinehelp.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_workbooksviews.htm#publish_workbook

irwando:
Duplicate of #315

julien10:
Any update on that matter ?

shinchris:
This has been added by @scuml in #614.

PR #614, opened 2020-05-01
Adds hidden_views parameter to publish()

scuml:
Resolves Issue #354 

Adds `hidden_views` as a parameter to publish().

```
new_workbook = server.workbooks.publish(
        workbook_item=new_workbook, file_path=filename, mode="CreateNew",
        hidden_views=["Sheet 44", "Test View"],
) 
```


shinchris:
Thanks for submitting a pull request!
Before we can merge this change, could you read and sign our [Contributor License Agreement](http://tableau.github.io/contributing.html)?

scuml:
Send it in on Friday afternoon.

RussTheAerialist:
Can confirm receipt of the CLA

Sambhawam07:
Is there a way to get the list of hidden views using the api?
In order to get the data from the hidden views.

jmeisele:
Same question ^

bcantoni:
@jmeisele if you don't mind, could you create a new issue for that question? The we can address it over there instead of getting lost on this closed PR.

cody2094:
@bcantoni Did that new issue get created for that question? I have a similar need

bcantoni:
Hi @cody2094 I don't see one - if you don't mind could you create an issue with some details?

cody2094:
@bcantoni I started to, but after I looked into this a little more, it seems to me as though the views aren't necessarily "hidden", they are just not in Tableau server at all. I'm not sure it warrants an issue/feature request if the data isn't even in Tableau Server. Seems like this would be a big feature request, you mind confirming that something like that is even reasonable for me to post an issue for?

SaraUris:
Is there a way to hide all worksheets instead of giving a list of names?

PR #615, opened 2020-05-05
Django-style filters

scuml:
This adds a django-style filtering and sorting mechanism to tableau which significantly decreases implementation code complexity.

For instance, this:
```python
request_options = tsc.RequestOptions()
request_options.filter.add(tsc.Filter(
    tsc.RequestOptions.Field.ProjectName, tsc.RequestOptions.Operator.Equals, project_name
))
request_options.sort.add(tsc.Sort(
    tsc.RequestOptions.Field.ProjectName, TSC.RequestOptions.Direction.Desc))
all_workbooks, pagination_item = server.workbooks.get(request_options)
```

...can now be written in one line:
```
all_workbooks = server.workbooks.filter(project_name=project_name).sort("-project_name")
```

This method returns a queryset, which can be iterated to access members.
```
for workbook in all_workbooks:
    print(workbook)
```

Full example...
```
# return queryset with no filters
workbooks = server.workbooks.all() 

# filters can be appended in new lines
workbooks = workbooks.filter(project_name=project_name)

# sort can take multiple args, with desc direction added as a (-) prefix 
workbooks = workbooks.sort("project_name", "-created_at")

# pagination take these two keywords
workbooks = workbooks.paginate(page_size=10, page_number=2) 

# query is executed at time of access
for workbook in workbooks: 
    print(workbook)

# pagination item is still accessible  
print(workbooks.total_available, workbooks.page_size, workbooks.page_number) 

# methods can be chained
all_workbooks = server.workbooks.filter(project_name=project_name).sort("-project_name")

# operators are implemented using dunderscore
all_workbooks = server.workbooks.filter(project_name__in=["Project A", "Project B"])


```

Documentation will need to be updated if accepted.


scuml:
Looking for repo owners to take a look at this and comment/merge.

jorwoods:
This will be really handy, and remove the need for some custom code to enact filters.

bcantoni:
@shinchris and @RussTheAerialist please review.

shinchris:
Overall, looks good to me. It will definitely make querying a lot simpler, especially with filtering and sorting. And allows the old methods to continue to work.

@RussTheAerialist what are you thoughts on style for querying?

scuml:
@RussTheAerialist Would you be able to look at this?

RussTheAerialist:
Sorry for the delay in reviewing, at the surface this looks great, I need to spend a little more time to think through implications around compatibility with the original implementation.  I also validated that we have a valid CLA signed for this change.

RussTheAerialist (approved):
Yup, I have no concerns on backcompat, so I'm going to merge this in. Thank you @scuml for the amazing contribution!

abubakar343:
This code filters the views. But how can we filter a specific view? For example, I have view which has multiple regions and I want to fetch the view where regions=='UK' for example.

jorwoods:
Take a look at this sample. 

https://github.com/tableau/server-client-python/blob/master/samples/export.py

PR #623, opened 2020-05-18
Simple Paging Endpoint for GraphQL/Metadata API

t8y8:
I took a simple approach for now, unlike `Pager` which takes an endpoint and works some magic, this just runs a query and collects the results.


t8y8:
Note, I thought of a way to do testing based on https://requests-mock.readthedocs.io/en/latest/response.html#response-lists

tableauserverclient/server/endpoint/metadata_endpoint.py
```diff
@@ -1,12 +1,43 @@
...
+                    arr.append(v)
+        elif isinstance(obj, list):
+            for item in obj:
+                extract(item, arr, key)
+        return arr
+
+    results = extract(obj, arr, key)
+    return results
+
+    def get_page_info(result):
```
shinchris, line 35: Should this method be un-indented?

tableauserverclient/server/endpoint/metadata_endpoint.py
```diff
@@ -1,12 +1,43 @@
...
-from .exceptions import GraphQLError
-
+from .exceptions import GraphQLError, InvalidGraphQLQuery
 import logging
 import json
 
 logger = logging.getLogger('tableau.endpoint.metadata')
 
 
+def is_valid_paged_query(parsed_query):
```
shinchris, line 9: Are these methods outside of the Metadata class because they are static?

tableauserverclient/server/endpoint/metadata_endpoint.py
t8y8: Yes! Fixed.

tableauserverclient/server/endpoint/metadata_endpoint.py
t8y8: Yup, I thought about `staticmethods` but they're really just internal helpers for parsing the json, so I moved them to module functions

t8y8:
Ok getting those tests in there was an adventure. I still don't do any error checking on nth calls, only the first. I'll think about if that matters, or for not just make this endpoint "all or nothing"

PR #635, opened 2020-06-16
Adds in maxage param to csv and pdf export options

shinchris:
Closing #530 and opening this one. 
This PR is to add maxage params to the csv and pdf export options.

Issue #628, opened 2020-05-22
In views.populate_csv,  consider support for the max_age parameter.


ashwinpn:
When querying view data the REST API supports passing a max age parameter to avoid hitting the cache when requesting the data for a view. The views.populate csv method, or the CSVRequestOptions class, is not currently used.

Are there any current plans to implement something along these lines?

razze76:
I would need this as well. Any plans for implementing anytime soon.

shinchris:
Implemented with #635 and #639 in the `development` branch.

Issue #581, opened 2020-03-09
Add support for the max_age parameter for views.populate_csv


timothyvermeiren:
The REST API supports passing a max_age parameter when [querying view data](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#query_view_data) to avoid hitting the cache when requesting the data for a view. Currently, the [method](https://tableau.github.io/server-client-python/docs/api-ref#viewspopulate_csv) `views.populate_csv` or the [CSVRequestOptions class](https://tableau.github.io/server-client-python/docs/api-ref#csvrequestoptions-class) do not.

Is it on the roadmap to add that capability? There is [this workaround](https://github.com/tableau/server-client-python/issues/568#issuecomment-596151576), but native support would be interesting.

rookxyz:
The same should be added to view.populate_pdf.

ndolash:
@timothyvermeiren @rookxyz 

I made a change to the TSC library so maxage (maxAge as in API spec) parameter is used in PDFRequestOptions class, however I still receive the cached PDF, so I am not sure if this is relevant for the caching solution. 

I cannot even try different pagetype.

rookxyz:
To verify cache update I used the following test:
1. Generate PDF
2. update data source, wait maxAge minutes
3. Generate PDF - this should come with the new data (updates cache for around 75% of time for me, which was better than users clicking in the Web UI). Not perfect but acceptable.

PR #639, opened 2020-06-26
Bugfix for maxage

shinchris:
With #635, I realized that 0 is actually a valid value. Adding -1 as the default value for when no maxAge is set. 240 is the maximum value server allows. Doc change to follow.

Image export maxage is available in all supported versions of Tableau server. PDF and CSV maxage params are supported for versions 2019.2.13+. 

Addresses #628 and #581.

### DIFF
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -103,7 +103,7 @@ class _FilterOptionsBase(RequestOptionsBase):
 
 
 class CSVRequestOptions(_FilterOptionsBase):
-    def __init__(self, maxage=None):
+    def __init__(self, maxage=-1):
         super(CSVRequestOptions, self).__init__()
         self.max_age = maxage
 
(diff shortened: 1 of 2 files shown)


tableauserverclient/server/request_options.py
```diff
@@ -112,13 +112,13 @@ def max_age(self):
         return self._max_age
 
     @max_age.setter
-    @property_is_int(range=(0, 240))
+    @property_is_int(range=(0, 240), allowed=[-1])
```
t8y8, line 115: I added `allowed` for cases where the allowed value is non-contiguous. Like a range of 1-100, and -1. Here you could just use a range of -1 to 240.

That said, this sort of makes it more readable, that -1 is a 'different' kind of value, so I think it's fine either way

shinchris: That's a really good point. I will keep it since it does make it more readable

PR #640, opened 2020-06-26
Update docs with maxage info

shinchris:
Docs for #635 and #639

### DIFF
--- a/docs/api-ref.md
+++ b/docs/api-ref.md
@@ -1597,13 +1597,13 @@ Name  |  Description
 ```py
 CSVRequestOptions(maxage=-1)
 ```
-Use this class to specify view filters to be applied when the CSV data is generated. This class can also be used to specify the maxage of the cached CSV data on server. See `views.populate_csv`.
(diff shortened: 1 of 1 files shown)


docs/api-ref.md
```diff
@@ -4042,4 +4055,49 @@ server.workbooks.update_conn(workbook, connection)
...
 <br> 
 
+#### workbooks.populate_pdf
+```
+workbooks.populate_pdf(workbook_item, req_options=None)
+```
+
+Populates the PDF content of the specified workbook.
+
+This method uses the `id` field to query the PDF content, and populates the content as the `pdf` field.
```
dzucker-tab, line 4065: _Not sure I understand this, but taking a crack based on my understanding of the REST endpoint. Also suggest this change for line 3354 in this review._

Suggest:

This method populates a PDF with image(s) of the workbook view(s) you specify.

shinchris:
Thanks @dzucker-tab for the detailed review!

PR #644, opened 2020-07-09
Adds a sample for publishing datasources

shinchris:


samples/publish_datasource.py
```diff
@@ -0,0 +1,81 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(description='Publish a datasource to server.')
+    parser.add_argument('--server', '-s', required=True, help='server address')
+    parser.add_argument('--site', '-i', help='site name')
+    parser.add_argument('--username', '-u', required=True, help='username to sign into server')
+    parser.add_argument('--filepath', '-f', required=True, help='filepath to the datasource to publish')
+    parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
+                        help='desired logging level (set to error by default)')
+    parser.add_argument('--as-job', '-a', help='Publishing asynchronously', action='store_true')
```
RussTheAerialist, line 32: We should just call this "--async" because "as job" doesn't communicate what is happening, imo.

samples/publish_datasource.py
```diff
@@ -0,0 +1,81 @@
...
+import logging
+
+import tableauserverclient as TSC
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Publish a datasource to server.')
+    parser.add_argument('--server', '-s', required=True, help='server address')
+    parser.add_argument('--site', '-i', help='site name')
+    parser.add_argument('--username', '-u', required=True, help='username to sign into server')
```
RussTheAerialist, line 28: We should do this using the personal access token instead of username/password to get people to start using them.

samples/publish_datasource.py
```diff
@@ -0,0 +1,81 @@
...
+        new_datasource = TSC.DatasourceItem(project_id="")
+
+        # Create a connection_credentials item if connection details are provided
+        new_conn_creds = None
+        if args.conn_username:
+            new_conn_creds = TSC.ConnectionCredentials(args.conn_username, args.conn_password,
+                                                       embed=args.conn_embed, oauth=args.conn_oauth)
+
+        # Define publish mode - Overwrite, Append, or CreateNew
+        publish_mode = TSC.Server.PublishMode.Overwrite
```
RussTheAerialist, line 65: Should we make this a command line option?

samples/publish_datasource.py
shinchris: Yeah, I agree with this. Fixed in new commit

samples/publish_datasource.py
shinchris: Mainly wanted to demonstrate using the enum rather than strings

RussTheAerialist:
🚀

PR #646, opened 2020-07-10
Development to master for v0.12

shinchris:
v0.12
* Added hidden_views parameter to workbook publish method (#614)
* Added simple paging endpoint for GraphQL/Metadata API (#623)
* Added endpoints to Metadata API for retrieving backfill/eventing status (#626)
* Added maxage parameter to CSV and PDF export options (#635)
* Added support for querying, adding, and deleting favorites (#638)
* Added a sample for publishing datasources (#644)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 0.12 (10 July 2020)
+
+* Added hidden_views parameter to workbook publish method (#614)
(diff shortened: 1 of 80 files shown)


nandoroz:
Hi all, is it possible call an incremental refresh with refresh.py script?

bcantoni:
@nandoroz could you raise that question under Issues? Then it will get more visibility than here at the end of this pull request.

PR #650, opened 2020-07-21
Remove urllib3 pinned version

t8y8:
We should support all the right things now for quoting.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,6 @@ setup(
     setup_requires=pytest_runner,
     install_requires=[
         'requests>=2.11,<3.0',
-        'urllib3>=1.24.3,<2.0'
     ],
     tests_require=[
         'requests-mock>=1.0,<2.0',


setup.py
```diff
@@ -33,7 +33,6 @@
     setup_requires=pytest_runner,
     install_requires=[
         'requests>=2.11,<3.0',
-        'urllib3>=1.24.3,<2.0'
```
shinchris, line 36: Do we want to remove it completely or leave it to grab the latest version. Otherwise I think users would have to separately download urllib3 to use TSC.

t8y8: I believe requests will install it's own dependancies if we remove this line so we'll respect whatever it's pin is. I'll double check the commit when I added this to see.

t8y8: Yup, we used to not specify at all, and let requests manage itself -- assuming we're all good on the quotes, we should be ok now. (I wouldn't merge this until @gaoang2148 confirms all is well on the Server side though)

shinchris: Oh okay that makes sense!

PR #655, opened 2020-07-24
Enable switch site functionality

jorwoods:
Adds capability for a user to switch sites using the REST API.

### DIFF
--- a/tableauserverclient/server/endpoint/auth_endpoint.py
+++ b/tableauserverclient/server/endpoint/auth_endpoint.py
@@ -1,5 +1,5 @@
 from ..request_factory import RequestFactory
-
+from .exceptions import ServerResponseError
 from .endpoint import Endpoint, api
 import xml.etree.ElementTree as ET
 import logging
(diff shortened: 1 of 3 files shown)


jorwoods:
@RussTheAerialist I have already signed and sent in the CLA.

RussTheAerialist:
@jorwoods I wasn't able to find you in our list of signed people, though I see you've contributed in the past, So it's probably a record-keeping error on our part. I'm double-checking with our legal team who keeps the hard copies to make sure we have it. Thanks for understanding!

RussTheAerialist:
@jorwoods Our legal team couldn't find any record of it either, I'm afraid.  Do you mind resending it to github@tableau.com ? I  am sorry that we seem to have lost track of the original CLA, and I hope this isn't too much of an inconvenience.

jorwoods:
@RussTheAerialist Not a problem. Resent it. Let me know if that doesn't satisfy the need.

RussTheAerialist:
Got it now! Thank you for responding so quickly!

jorwoods:
I know this already got approved, but since it hasn't been merged yet, one looming question: It technically produces an error if the user tries to switch sites to their current site. Should the user handle that within their code, or should that be handled quietly within TSC?

t8y8:
@jorwoods Do we really throw an error on that, happen to know which HTTP code we throw? (I might have even been the PM for the API at that time, so it's maybe my fault).

My first thought is just to swallow it and ignore, I could see certain user code switching to self just to be safe in certain scenarios and not wanting an error... what do you think?

If we do want to go that route, there are a few examples of how we do that today. The preference is to create a a new specific exception, then catch it and swallow, and rethrow anything else.

PR #658, opened 2020-08-06
Doc pages cleanup and minor editing

bcantoni:
- minor content edits
- update Tableau doc links
- update dev community forum links
- add missing samples to samples.md list


docs/versions.md
```diff
@@ -55,17 +52,17 @@ The current version of TSC only supports the following REST API and Tableau Serv
...
-|2.7|10.4|
-|2.8|10.5|
-|3.0|2018.1|
-|3.1|2018.2|
-|3.2|2018.3|
-|3.3|2019.1|
-|3.4|2019.2|
-|3.5|2019.3|
-|3.6|2019.4|
 |3.7|2020.1|
```
shinchris, line 55: Could you also add 3.8 - 2020.2 here?

docs/versions.md
```diff
@@ -25,14 +25,13 @@ For example, the code might display version `2.3`.
 
 ## Use the REST API version supported by the server
 
-To always use the latest version of the REST API that is supported by the instance of Tableau Server you are connecting to, set the version using the `use_server_version()` method. You can do this when you create the server instance by specifying `use_server_version=True` as one of the arguments.
```
shinchris, line 28: Both methods are still valid, we just didn't have an example for the first method. It would looks like this:
```
server.use_server_version()
```

docs/api-ref.md
```diff
@@ -3742,7 +3742,7 @@ Deletes a workbook with the specified ID.
 To specify the site, create a `TableauAuth` instance using the content URL for the site (`site_id`), and sign in to that site.  See the [TableauAuth class](#tableauauth-class).  
 
 
-REST API: [Delete Workbook](https://onlinehelp.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#Delete_Workbook31){:target="_blank"}  
+REST API: [Delete Workbook](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#Delete_Workbook31)
```
shinchris, line 3745: This url should be: https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#delete_workbook1

docs/api-ref.md.rej
```diff
@@ -0,0 +1,68 @@
+***************
```
RussTheAerialist, line 1: yeah we definitely don't want to have this as part of the repo.

docs/dev-guide.md
```diff
@@ -65,21 +66,20 @@ This section will get you started with the basic workflow, describing how to cre
...
 ### Commit changes to your fork and open a PR
 
-1. Make a PR as described [here](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) against the **development** branch for code changes, or against **gh-pages** for documentation updates.
+1. Make a PR as described [here](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork) against the **development** branch for code changes, or against **gh-pages** for documentation updates.
 
 1. Wait for a review and address any feedback.
    While we try and stay on top of all issues and PRs it might take a few days for someone to respond. Politely pinging
    the PR after a few days with no response is OK, we'll try and respond with a timeline as soon as we are able.
 
-1. That's it! When the PR has received ![](https://assets-cdn.github.com/images/icons/emoji/unicode/1f680.png){:height="5%" width="5%"} (:rocket:'s) from members of the core team they will merge the PR
```
RussTheAerialist, line 78: awww getting rid of the rocket ship! could we do "approval 🚀" at least?

docs/versions.md
RussTheAerialist: And there are reasons you might want to use the separate version rather than the argument version.  For instance, if you want to disable certificate checking, you have to do that after you create the server object, but before you make calls to the server (use_server_version makes a call to the server).

docs/versions.md
bcantoni: done

docs/versions.md
bcantoni: got it - my earlier understanding was wrong, but I've rolled this back in now

docs/api-ref.md
bcantoni: done, good catch

docs/api-ref.md.rej
bcantoni: yea i fat-fingered a few files there in this PR :(

docs/dev-guide.md
bcantoni: ah, the current url was a 404 but i found a new one and will bring it back!

bcantoni:
@RussTheAerialist @shinchris I've incorporated all the feedback including bringing back the 🚀 😄

PR #659, opened 2020-08-06
Minor cleanup pass

bcantoni:
See individual commits for details - this is mostly a lightweight editing pass and fixing a couple of test issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "3.5"
   - "3.6"
   - "3.7"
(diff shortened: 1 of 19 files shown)


bcantoni:
@RussTheAerialist @shinchris FYI, the test code bit in `test/test_workbook.py` was actually failing under Python 3.8. For some reason in the request body the name and hidden attributes were swapping in 3.8, causing the test to fail. I switch to using regex there so the test now works in all Python versions.

PR #667, opened 2020-08-13
Data alert endpoint and item

jorwoods:
Adds in the Data Alert endpoint and item to match functionality of REST API.

### DIFF
--- a/tableauserverclient/models/data_alert_item.py
+++ b/tableauserverclient/models/data_alert_item.py
@@ -1,0 +1,190 @@
+import xml.etree.ElementTree as ET
+
+from .property_decorators import property_not_empty
+from .user_item import UserItem
+from .view_item import ViewItem
+
+
+class DataAlertItem(object):
+    def __init__(self):
+        self._id = None
+        self._subject = None
+        self._creatorId = None
+        self._createdAt = None
+        self._updatedAt = None
+        self._frequency = None
+        self._public = None
+        self._owner_id = None
+        self._owner_name = None
+        self._view_id = None
+        self._view_name = None
+        self._workbook_id = None
+        self._workbook_name = None
+        self._project_id = None
+        self._project_name = None
+        self._recipients = None
+
+    def __repr__(self):
+        return "<Data Alert {} subject={} frequency={} public={}>".format(self.id, self.subject, 
+                                                                          self.frequency, self.public)
+
+    @property
+    def id(self):
+        return self._id
+
+    @property
+    def subject(self):
+        return self._subject
+
+    @subject.setter
+    @property_not_empty
+    def subject(self, value):
+        self._subject = value
+
+    @property
+    def frequency(self):
+        return self._frequency
+
+    @frequency.setter
+    @property_not_empty
+    def frequency(self, value):
+        self._frequency = value
+
+    @property
+    def public(self):
+        return self._public
+
+    @public.setter
+    @property_not_empty
+    def public(self, value):
+        self._public = value
+
+    @property
+    def creatorId(self):
+        return self._creatorId
+
+    @property
+    def recipients(self):
+        return self._recipients or list()
+
+    @property
+    def createdAt(self):
+        return self._createdAt
+
+    @property
+    def updatedAt(self):
+        return self._updatedAt
+
+    @property
+    def owner_id(self):
+        return self._owner_id
+
+    @property
+    def owner_name(self):
+        return self._owner_name
+
+    @property
+    def view_id(self):
+        return self._view_id
+
+    @property
+    def view_name(self):
+        return self._view_name
+
+    @property
+    def workbook_id(self):
+        return self._workbook_id
+
+    @property
+    def workbook_name(self):
+        return self._workbook_name
+
+    @property
+    def project_id(self):
+        return self._project_id
+
+    @property
+    def project_name(self):
+        return self._project_name
+
+    def _set_values(self, id, subject, creatorId, createdAt, updatedAt,
+                    frequency, public, recipients, owner_id, owner_name,
+                    view_id, view_name, workbook_id, workbook_name, project_id,
+                    project_name):
+        if id is not None:
+            self._id = id
+        if subject:
+            self._subject = subject
+        if creatorId:
+            self._creatorId = creatorId
+        if createdAt:
+            self._createdAt = createdAt
+        if updatedAt:
+            self._updatedAt = updatedAt
+        if frequency:
+            self._frequency = frequency
+        if public:
+            self._public = public
+        if owner_id:
+            self._owner_id = owner_id
+        if owner_name:
+            self._owner_name = owner_name
+        if view_id:
+            self._view_id = view_id
+        if view_name:
+            self._view_name = view_name
+        if workbook_id:
+            self._workbook_id = workbook_id
+        if workbook_name:
+            self._workbook_name = workbook_name
+        if project_id:
+            self._project_id = project_id
+        if project_name:
+            self._project_name = project_name
+        if recipients:
+            self._recipients = recipients
+
+    @classmethod
+    def from_response(cls, resp, ns):
+        all_alert_items = list()
+        parsed_response = ET.fromstring(resp)
+        all_alert_xml = parsed_response.findall('.//t:dataAlert', namespaces=ns)
+
+        for alert_xml in all_alert_xml:
+            kwargs = cls._parse_element(alert_xml, ns)
+            alert_item = cls()
+            alert_item._set_values(**kwargs)
+            all_alert_items.append(alert_item)
+
+        return all_alert_items
+
+    @staticmethod
+    def _parse_element(alert_xml, ns):
+        kwargs = dict()
+        kwargs['id'] = alert_xml.get('id', None)
+        kwargs['subject'] = alert_xml.get('subject', None)
+        kwargs['creatorId'] = alert_xml.get('creatorId', None)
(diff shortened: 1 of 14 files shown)


tableauserverclient/server/endpoint/data_alert_endpoint.py
```diff
@@ -0,0 +1,88 @@
...
+        # DELETE /api/api-version/sites/site-id/dataAlerts/data-alert-id/users/user-id
+        url = "{0}/{1}".format(self.baseurl, dataalert_id)
+        self.delete_request(url)
+        logger.info('Deleted single dataalert (ID: {0})'.format(dataalert_id))
+
+    @api(version="3.2")
+    def delete_user_from_alert(self, dataalert, user):
+        dataalert_id = getattr(dataalert, 'id', dataalert)
+        user_id = getattr(user, 'id', user)
+        if not dataalert_id:
```
shinchris, line 56: Could you also add this check for user id?

tableauserverclient/server/endpoint/data_alert_endpoint.py
```diff
@@ -0,0 +1,88 @@
...
+            error = "Dataalert ID undefined."
+            raise ValueError(error)
+        # DELETE /api/api-version/sites/site-id/dataAlerts/data-alert-id/users/user-id
+        url = "{0}/{1}/users/{2}".format(self.baseurl, dataalert_id, user_id)
+        self.delete_request(url)
+        logger.info('Deleted single dataalert (ID: {0})'.format(dataalert_id))
```
shinchris, line 62: Logging message is copied from the other delete method. Same with the commented line above

tableauserverclient/server/endpoint/data_alert_endpoint.py
```diff
@@ -0,0 +1,88 @@
...
+    @api(version="3.2")
+    def add_user_to_alert(self, dataalert_item, user):
+        if not dataalert_item.id:
+            error = "Dataalert item missing ID."
+            raise MissingRequiredFieldError(error)
+
+        url = "{0}/{1}/users".format(self.baseurl, dataalert_item.id)
+        update_req = RequestFactory.DataAlert.add_user_to_alert(dataalert_item, user)
+        server_response = self.post_request(url, update_req)
+        logger.info('Updated dataalert item (ID: {0})'.format(dataalert_item.id))
```
shinchris, line 73: Logging message should be for adding user to alert

tableauserverclient/server/endpoint/data_alert_endpoint.py
```diff
@@ -0,0 +1,88 @@
...
+        if not dataalert_id:
+            error = "Dataalert ID undefined."
+            raise ValueError(error)
+        # DELETE /api/api-version/sites/site-id/dataAlerts/data-alert-id/users/user-id
+        url = "{0}/{1}/users/{2}".format(self.baseurl, dataalert_id, user_id)
+        self.delete_request(url)
+        logger.info('Deleted single dataalert (ID: {0})'.format(dataalert_id))
+
+    @api(version="3.2")
+    def add_user_to_alert(self, dataalert_item, user):
```
shinchris, line 65: Since we only grab the id from the user item, I think you can have the 'user' variable be either the item or the id, like you did in delete_user_from_alert().

tableauserverclient/models/data_alert_item.py
```diff
@@ -0,0 +1,190 @@
...
+    @property_not_empty
+    def subject(self, value):
+        self._subject = value
+
+    @property
+    def frequency(self):
+        return self._frequency
+
+    @frequency.setter
+    @property_not_empty
```
shinchris, line 49: I think it would be really nice to have a frequency enum that lists out the 5 valid values. You can also use the enum property decorator to ensure that a user passes in a valid frequency value. Similar to how it's done in schedule_item.py

jorwoods: Created enum and used decorator to validate values.

tableauserverclient/server/endpoint/data_alert_endpoint.py
jorwoods: Done.

tableauserverclient/server/endpoint/data_alert_endpoint.py
jorwoods: Done.

tableauserverclient/server/endpoint/data_alert_endpoint.py
jorwoods: Done.

tableauserverclient/server/endpoint/data_alert_endpoint.py
jorwoods: Done.

tableauserverclient/models/data_alert_item.py
```diff
@@ -0,0 +1,197 @@
+import xml.etree.ElementTree as ET
+
+from .property_decorators import property_not_empty, property_is_enum
+from .user_item import UserItem
+from .view_item import ViewItem
+
+
+class DataAlertItem(object):
+    class Frequency:
```
shinchris, line 9: The enums are intended to allow users to pass in `TSC.DataAlertItem.Frequency.Once`, rather than a string value. With the way the `@property_is_enum` annotation is implemented, the string values listed here have to match the enum names exactly for it to work (e.g. `Once = 'Once'`). But server only accepts lowercase values for this particular field, so we need to convert it to lowercase in request_factory.

Could you change these string values to start with capitalized letters and also convert frequency into lowercase in request_factory.py?

tableauserverclient/server/endpoint/data_alert_endpoint.py
```diff
@@ -0,0 +1,94 @@
...
+logger = logging.getLogger('tableau.endpoint.dataalerts')
+
+
+class DataAlerts(Endpoint):
+    def __init__(self, parent_srv):
+        super(DataAlerts, self).__init__(parent_srv)
+
+    @property
+    def baseurl(self):
+        return "{0}/sites/{1}/dataalerts".format(self.parent_srv.baseurl, self.parent_srv.site_id)
```
t8y8, line 19: Is the url case sensitive? If not, can we satisfy my OCD and say dataAlerts?

tableauserverclient/server/endpoint/data_alert_endpoint.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @api(version="3.2")
+    def get(self, req_options=None):
+        logger.info('Querying all dataalerts on site')
+        url = self.baseurl
+        server_response = self.get_request(url, req_options)
+        pagination_item = PaginationItem.from_response(server_response.content, self.parent_srv.namespace)
+        all_dataalert_items = DataAlertItem.from_response(server_response.content, self.parent_srv.namespace)
```
t8y8, line 27: nit: camelCase `dataAlert`

tableauserverclient/server/endpoint/data_alert_endpoint.py
```diff
@@ -0,0 +1,94 @@
...
+            raise ValueError(error)
+        logger.info('Querying single dataalert (ID: {0})'.format(dataalert_id))
+        url = "{0}/{1}".format(self.baseurl, dataalert_id)
+        server_response = self.get_request(url)
+        return DataAlertItem.from_response(server_response.content, self.parent_srv.namespace)[0]
+
+    @api(version="3.2")
+    def delete(self, dataalert):
+        dataalert_id = getattr(dataalert, 'id', dataalert)
+        if not dataalert_id:
```
t8y8, line 44: `dataAlert`

tableauserverclient/models/data_alert_item.py
```diff
@@ -0,0 +1,197 @@
...
+    def frequency(self, value):
+        self._frequency = value
+
+    @property
+    def public(self):
+        return self._public
+
+    @public.setter
+    @property_not_empty
+    def public(self, value):
```
shinchris, line 66: This one can be `@property_is_boolean`

tableauserverclient/server/request_factory.py
```diff
@@ -86,6 +86,27 @@ def update_req(self, column_item):
...
+        user_element = ET.SubElement(xml_request, 'user')
+        user_element.attrib['id'] = user_id
+
+        return ET.tostring(xml_request)
+
+    def update_req(self, alert_item):
+        xml_request = ET.Element('tsRequest')
+        dataAlert_element = ET.SubElement(xml_request, 'dataAlert')
+        dataAlert_element.attrib['subject'] = alert_item.subject
+        dataAlert_element.attrib['frequency'] = alert_item.frequency
```
shinchris, line 101: this is where we have to convert to lowercase

tableauserverclient/server/endpoint/data_alert_endpoint.py
jorwoods: Patched casing throughout entire item. Should be consistent now.

tableauserverclient/server/request_factory.py
jorwoods: `.lower()` call added

tableauserverclient/models/data_alert_item.py
jorwoods: Done.

tableauserverclient/models/data_alert_item.py
jorwoods: Capital cased the string values.

PR #672, opened 2020-08-19
Extract operations to match tabcmd

jacalata:
Added endpoints to match tabcmd functionality around create/delete extracts on datasource/workbook, and encrypt/decrypt/reencrypt extracts for a site. 

- a bunch of whitespace fixes to come that will make pycodestyle happier

### DIFF
--- a/tableauserverclient/server/endpoint/sites_endpoint.py
+++ b/tableauserverclient/server/endpoint/sites_endpoint.py
@@ -111,8 +111,7 @@ class Sites(Endpoint):
             raise ValueError(error)
         url = "{0}/{1}/encrypt-extracts".format(self.baseurl, site_id)
         empty_req = RequestFactory.Empty.empty_req()
-        self.post_request(url,empty_req)
-
+        self.post_request(url, empty_req)
 
     @api(version="3.5")
     def decrypt_extracts(self, site_id):
@@ -121,8 +120,7 @@ class Sites(Endpoint):
             raise ValueError(error)
         url = "{0}/{1}/decrypt-extracts".format(self.baseurl, site_id)
         empty_req = RequestFactory.Empty.empty_req()
-        self.post_request(url,empty_req)
-
+        self.post_request(url, empty_req)
 
     @api(version="3.5")
     def re_encrypt_extracts(self, site_id):
@@ -130,6 +128,6 @@ class Sites(Endpoint):
             error = "Site ID undefined."
             raise ValueError(error)
         url = "{0}/{1}/reencrypt-extracts".format(self.baseurl, site_id)
-        
+
         empty_req = RequestFactory.Empty.empty_req()
-        self.post_request(url,empty_req)
+        self.post_request(url, empty_req)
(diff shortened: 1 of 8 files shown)


tableauserverclient/server/endpoint/sites_endpoint.py
```diff
@@ -103,3 +103,31 @@ def create(self, site_item):
...
+    def decrypt_extracts(self, site_id):
+        if not site_id:
+            error = "Site ID undefined."
+            raise ValueError(error)
+        url = "{0}/{1}/decrypt-extracts".format(self.baseurl, site_id)
+        empty_req = RequestFactory.Empty.empty_req()
+        self.post_request(url, empty_req)
+
+    @api(version="3.5")
+    def re_encrypt_extracts(self, site_id):
```
RussTheAerialist, line 126: What's the difference from a conceptual level between encrypt and re-encrypt? Does the user need to really care about the difference or should we switch between the two based on inspection?

test/test_workbook.py
```diff
@@ -470,8 +470,11 @@ def test_publish_with_hidden_view(self):
                                                          hidden_views=['GDP per capita'])
 
             request_body = m._adapter.request_history[0]._request.body
-            self.assertTrue(re.search(rb'<views><view.*?hidden=\"true\".*?\/><\/views>', request_body))
-            self.assertTrue(re.search(rb'<views><view.*?name=\"GDP per capita\".*?\/><\/views>', request_body))
+            # order of attributes in xml is unspecified
```
RussTheAerialist, line 473: I don't think this change clearly communicates what is being tested. In the previously written tests, it reads as "this attribute exists in the response". In your change it reads as "the response is exactly this."  What happens if the views as a third property added, now you need 9 variations of the check. If you don't want to use regexes, then parse the xml and validate that the attributes hidden and name exist. I don't feel like the current change is an improvement in test strategy.

bcantoni: Yea in 12aacb9c7e1bac0a9c76b1e01ef3bb55548e77fa I switched to the dual regex checks here specifically because in Python 3.8 the order of attributes changed.

jacalata: This isn't intended as a change in strategy, this was just to make the tests pass. The previous test failed because the xml was reordered on my machine.

tableauserverclient/server/endpoint/sites_endpoint.py
jacalata: Hm, good question.  They're exposed as separate commands in tabcmd, but there's not any real justification for that. I'll check with the feature team.

jacalata: encrypt will only act on extracts that are not yet encrypted. re-encrypt will rotate the key for extracts that are already encrypted and do nothing for unencryped extracts.

jacalata: @RussTheAerialist how's that explanation? I think in our other help docs they are presented as two different operations, I don't think it would help to change that model here.

RussTheAerialist: Yup, totally makes sense! I mean, maybe I don't 100% agree with their design decisions, but 100% agree that we shouldn't change the model here.

PR #674, opened 2020-08-23
AD groups

jacalata:
This change implements creation/update of AD groups, as well as adding the license and site-role attributes to local groups.

### DIFF
--- a/tableauserverclient/server/request_factory.py
+++ b/tableauserverclient/server/request_factory.py
@@ -238,13 +238,36 @@ class GroupRequest(object):
         user_element.attrib['id'] = user_id
         return ET.tostring(xml_request)
 
-    def create_req(self, group_item):
+    def create_local_req(self, group_item):
         xml_request = ET.Element('tsRequest')
         group_element = ET.SubElement(xml_request, 'group')
         group_element.attrib['name'] = group_item.name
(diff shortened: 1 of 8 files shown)


tableauserverclient/server/request_factory.py
```diff
@@ -238,21 +238,46 @@ def add_user_req(self, user_id):
         user_element.attrib['id'] = user_id
         return ET.tostring(xml_request)
 
-    def create_req(self, group_item):
+    def create_local_req(self, group_item):
         xml_request = ET.Element('tsRequest')
         group_element = ET.SubElement(xml_request, 'group')
         group_element.attrib['name'] = group_item.name
+        if (group_item.license_mode is not None):
```
t8y8, line 245: Nit: no `()` around if statements (unless it's multi-line is the general rule, but I also think nested is ok)

tableauserverclient/server/request_factory.py
```diff
@@ -238,21 +238,46 @@ def add_user_req(self, user_id):
...
+            group_element.attrib['grantLicenseMode'] = group_item.license_mode
+        if (group_item.minimum_site_role is not None):
```
t8y8, line 247: `()`

PR #681, opened 2020-08-29
Fixed Assert in line 459 for test_workbook.py

raposo-dev:
When testing for the first time was getting an assert error in test_workbook.py line 459.

Replaced the request_body message and now all tests pass.


bcantoni:
@raposo-dev you probably already discovered this (since you closed your PR), but in the latest `development` branch we changed that test assertion to work properly in all supported versions of Python including 3.8.

Issue #578, opened 2020-03-04
the function vf() (view filter) does not handle with special character as "&"


ChadyNassar:
I am using the TSC API on python to generate multiple pdf files (one for each client).
When using  pdf_req_option.vf(parameter,value)
Where parameter = 'client' and value = the name of each client
All the pdf are generated correctly except for one client that have a '&' in his name.
For this client the PDF is generated but without any data inside.
It is the same result than if  I gave a fake value that dosen't exist in my list of 'client' parameter.

divinorum-webb:
Hey Chady, 

I encountered a similar issue very recently. You can work around this successfully, and it is worth noting that even if TSC scrubbed the '&' value from your filter value, your dashboard would not display the correct data due to the values no longer matching -- so you must fix this on your end.

First, this is not an error of the TSC library -- the '&' character is used to append additional URL params to your URL string, so when the URL string encounters the '&' in your client's name it assumes the text that follows is an entirely new URL parameter.

Even though this cannot be fixed from TSC's side, here is how you can resolve the issue:
1) build a new version of your field in Tableau which removes the '&' character from the field
2) be sure to use this new field as a filter in your dashboard / view that is published
3) reference the new field as your vf_ filter value

Hope that helps,
Elliott

ChadyNassar:
Hi Elliott,

Thank you for your quick response. It's very clear.

Finally I found a solution thanks to URL-encoding Reference using
*string.replace("&","%26)*

Have a good day.

Chady

vignesh1609:
Hey all, My filters are not working with .vf() i am getting pdf with the filters which is published . can you suggest me how to resolve this issue.

TimMoyce:
Is there any way of passing multiple parameters to a single view
I currently have 
pdf_req_option.vf(filter_criteria1, Parameter_StoreBrand)
but cannot find a solution to add a second criteria. I am aware how to do this using URL parameters but not the TSC module
Thanks

The-Loud:
> Is there any way of passing multiple parameters to a single view
> I currently have
> pdf_req_option.vf(filter_criteria1, Parameter_StoreBrand)
> but cannot find a solution to add a second criteria. I am aware how to do this using URL parameters but not the TSC module
> Thanks

Hi @TimMoyce,

If you want to add multiple values such as in a drop-down selection box, you can do something like
`pdf_req_open.vf('filter_name', 'value1,value2,value3')`

[...]

Issue #516, opened 2019-10-29
query string not escaping '+'


kurtbrose:
1- create user with name='a+b'

2- fetch user with filter query set to name='a+b'; returns no items

3- fetch user with filter query set to name='a%2b'; returns user

I don't know how or where, but seems pretty clear somewhere it is doing `?name={name}` in a query string, and it should be encoding but isn't

t8y8:
We aren't doing encoding ourselves anywhere, it's handled by the http library we use.

I can take a look when I have time, but I'm not sure what options we have available to us over the lower layer's encoding

kurtbrose:
I fixed this by doing `urlencode()` on values before passing them to the API

kurtbrose:
dude, you are doing string munging to generate query params

https://github.com/tableau/server-client-python/blob/b260ebcafec6e6ebe89597f13e1e18ed9ca55406/tableauserverclient/server/request_options.py#L95-L97

https://github.com/tableau/server-client-python/blob/master/tableauserverclient/server/request_options.py#L164

```python
        self._append_view_filters(params)

        return "{0}?{1}".format(url, '&'.join(params))
```

kurtbrose:
that took me less than 5 minutes looking at the code

kurtbrose:
sorry for being a little short, trying to get tableau integration working has been my life for the past week or so

t8y8:
Right, but our http library itself encodes those values when it makes the request (I had a bug in the opposite direction recently, maybe this is actually related to that)

We can look at encoding it at the places you just mentioned, I just want to make sure nothing gets double encoded

kurtbrose:
makes sense -- maybe it is possible to pass the query string to HTTP as a key/value dict instead of doing the string construction yourself?

kurtbrose:
that seems to be the recommended way to use requests: https://requests.kennethreitz.org/en/master/user/quickstart/#passing-parameters-in-urls

kurtbrose:
https://github.com/tableau/server-client-python/blob/b260ebcafec6e6ebe89597f13e1e18ed9ca55406/tableauserverclient/server/endpoint/endpoint.py#L45 looking here it tries to manually smoosh up the URL, but instead it good just say `parameters['params'] = request_object.get_query_params()` or similar

there's no issue -- body data is `data={...}`; so even on a post request you can set `params={...}` and everything works fine

jacalata:
This works today.

Issue #418, opened 2019-04-03
RequestOptions Filter does not properly escape strings.


1oglop1:
Hi, I tried to find user by user name using request options however I still got errors until I tried to `quote_plus` the user name string.

## Task:
Find user via `user name`
`username=1oglop1+24@gmail.com`

### Steps
```python
import tableauserverclient as tsc
# .. server sign in etc..
request_options = tsc.RequestOptions()
request_options.filter.add(
    tsc.Filter(
        tsc.RequestOptions.Field.Name, 
        tsc.RequestOptions.Operator.Equals, 
        '1oglop1+24@gmail.com'
        )
)
users, pagination_item = server.users.get(request_options)
print(users)
```

### Expected result
`[<User ada95cbc-ca59-4087-bbfe-acbcf55d88a4 name=1oglop1+24@gmail.com role=Creator>]`

### Actual result
`[]`

This is due to the character `+` in `username`.
I'm using `quote_plus` as a workaround.

```
import tableauserverclient as tsc
from urllib.parse import quote_plus
# .. server sign in etc..
request_options = tsc.RequestOptions()
request_options.filter.add(
    tsc.Filter(
        tsc.RequestOptions.Field.Name,
        tsc.RequestOptions.Operator.Equals,
        quote_plus('1oglop1+24@gmail.com')
        )
)
users, pagination_item = server.users.get(request_options)
print(users)
# [<User ada95cbc-ca59-4087-bbfe-acbcf55d88a4 name=1oglop1+24@gmail.com role=Creator>]
```

rhelenius:
Quote plus worked for me as well except for when the % character is present.

jacalata:
This works today

PR #686, opened 2020-09-04
Removes manual query param generation

shinchris:
Removing manual query param generation in req_options, and delegating that step to the requests library.
By doing this, the requests library will encode the url for us, fixing #418, #516, #578.

Some endpoints have their own query params added, like workbook publish overwrite flag. The requests library will append new query params to the existing one in this case.

### DIFF
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -2,6 +2,22 @@ from ..models.property_decorators import property_is_int
 
 
 class RequestOptionsBase(object):
+    def apply_query_params(self, url):
+        import warnings
+        warnings.simplefilter('always', DeprecationWarning)
+        warnings.warn('apply_query_params is deprecated, please use get_query_params instead.', DeprecationWarning)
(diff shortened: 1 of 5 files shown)


tableauserverclient/server/request_options.py
```diff
@@ -2,7 +2,7 @@
 
 
 class RequestOptionsBase(object):
-    def apply_query_params(self, url):
```
t8y8, line 5: Since this isn't an _ method it is technically "public" -- do we want to preserve it by just having it throw a deprecation warning and then call get_query_params?

Or at a minimum throw a deprecation warning

t8y8 (changes requested):

tableauserverclient/server/request_options.py
shinchris: Thanks for pointing that out. I put back the old method, preserving the same behavior with a deprecation warning.

t8y8 (approved):

Issue #675, opened 2020-08-25
Delete tag does not work


tshouri:
I found that if the tag contains encodable characters such as '#' (which is quite commonly used for tagging!), the delete tag request neither succeeds nor throws an exception.

https://github.com/tableau/server-client-python/blob/43e1b063d83d3271b262d02e94445a930c161cb6/tableauserverclient/server/endpoint/resource_tagger.py#L28

bcantoni:
@tv-terminator there could be an issue there (I think we have or had a similar bug with encoding special characters). Can you give a fuller code snippet example of how you're trying to delete tags? Also what type of resource are the tags associated with?

tv-terminator:
I have tried both workbooks & datasources; ended up with same result. Heres a full log with code:

`# w = workbook_item, t = tableau connection/hook
>>> w=t.workbooks_get(get_by_content_url='POC')[0]
>>> w
<tableauserverclient.models.workbook_item.WorkbookItem object at 0x7f1116314e10>
>>> w.tags
set()
>>> w.tags.add('#test')
>>> w.tags
{'#test'}
>>> t.workbooks_update(w)
[2020-08-28 19:43:18,196] {{resource_tagger.py:47}} INFO - Updated tags to {'#test'}
[2020-08-28 19:43:18,283] {{workbooks_endpoint.py:89}} INFO - Updated workbook item (ID: xxxxxx-xxxxxx-xxxxx-xxxxx)
<tableauserverclient.models.workbook_item.WorkbookItem object at 0x7f1118fa3f28>
>>> w=t.workbooks_get(get_by_content_url='POC')[0]
>>> w.tags
{'#test'}
>>> w.tags.remove('#test')
>>> w.tags
set()
>>> t.workbooks_update(w)
[2020-08-28 19:44:10,195] {{resource_tagger.py:47}} INFO - Updated tags to set()
[2020-08-28 19:44:10,259] {{workbooks_endpoint.py:89}} INFO - Updated workbook item (ID: xxxxxx-xxxxxx-xxxxx-xxxxx)
<tableauserverclient.models.workbook_item.WorkbookItem object at 0x7f1116f73400>
>>> w=t.workbooks_get(get_by_content_url='POC')[0]
>>> w.tags
{'#test'}
`

Works fine with a plain text tag:
`
>>> w.tags.add('test')
>>> w.tags
{'#test', 'test'}
>>> t.workbooks_update(w)
[...]

Issue #622, opened 2020-05-14
ViewItem Tag Bug


erich-trabus:
https://github.com/tableau/server-client-python/blob/188be71e0fbe9d87ac36628af21d2f63af32d020/tableauserverclient/models/view_item.py#L161

should be:

``` python
import copy
...
view_item._initial_tags = copy.copy(tags)
```
otherwise tags won't update/delete properly. Is this bug worth a pull request?

bcantoni:
@erich-trabus thanks for reporting this; it's been fixed in the pull request mentioned above.

PR #687, opened 2020-09-04
Encode tag-name before deleting

shinchris:
Encodes tag-name in the request URL for deleting a tag, #675.
Note that tag names with forward slashes will not work.

Also fixes tag add/delete bug for views, #622.


t8y8 (approved):
🚀

Issue #682, opened 2020-08-31
'Groups' object has no attribute 'get_by_id'


victorumn:
I'm trying to get group item by id. using the sample code it returns 

AttributeError: 'Groups' object has no attribute 'get_by_id'

mygroup = server.groups.get_by_id('1a2b3c4d-5e6f-7a8b-9c0d-1e2f3a4b5c6d')

bcantoni:
@victorumn you're correct; [Groups](https://tableau.github.io/server-client-python/docs/api-ref#groups) does not have a `get_by_id()` method.

Instead, you could simply get all groups and then iterate through to find the one you want.

You can also use the request options to effectively "get by name":

```py
options = TSC.RequestOptions()
options.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
                                TSC.RequestOptions.Operator.Equals,
                                'All Users'))

all_groups, pagination_item = server.groups.get(req_options=options)
print("\nThere are {} groups on site: ".format(pagination_item.total_available))
for grp in all_groups:
    print(grp.id, grp.name)
    server.groups.populate_users(grp)
    for user in grp.users:
        print(" {}".format(user.name))
```

victorumn:
Thanks. I did use a similar workaround for that.

Since all other objects have the method, I thought this could be a bug. Also, the documentation (https://tableau.github.io/server-client-python/docs/) showed an example that it has that method.

bcantoni:
Good catch - I had missed that the example for `groups.remove_user` shows `server.groups.get_by_id`. I'll reopen this as a docs bug so we can clear that up.

bcantoni:
Doc/example cleaned up with changes by Chris^^

PR #688, opened 2020-09-04
Removes an example using groups.get_by_id

shinchris:
Groups does not have a query-by-id endpoint, but was included in an example (#682)


bcantoni (approved):
+1

PR #689, opened 2020-09-10
Added endpoints for revoking all server admins tokens

jessicachen79:


tableauserverclient/server/endpoint/auth_endpoint.py
```diff
@@ -73,3 +73,12 @@ def switch_site(self, site_item):
         self.parent_srv._set_auth(site_id, user_id, auth_token)
         logger.info('Signed into {0} as user with id {1}'.format(self.parent_srv.server_address, user_id))
         return Auth.contextmgr(self.sign_out)
+
+    @api(version="3.10")
+    def revoke_all_server_admin_tokens(self):
+        url = "{0}/{1}".format(self.baseurl, 'revokeAllServerAdminTokens')
+        if not self.parent_srv.is_signed_in():
```
shinchris, line 80: This check is necessary, since it'll just be a silent operation if a user isn't signed in. The post_request call below should throw an error telling the user to first sign in.

tableauserverclient/server/endpoint/auth_endpoint.py
```diff
@@ -73,3 +73,12 @@ def switch_site(self, site_item):
...
+            return
+        self.post_request(url, '')
+        self.parent_srv._clear_auth()
```
shinchris, line 83: I don't think we need this line, since it'll end the user's current session. From my understanding, this endpoint will revoke all server admin PATs, but not log the calling user out.

shinchris (changes requested):

tableauserverclient/server/endpoint/auth_endpoint.py
shinchris: sorry, I meant to say that the check is not necessary 😓 it's redundant since post_request() will check anyways

jessicachen79: I just removed the check from the most recent commit. Thanks for addressing that!

shinchris (approved):
Looks good, thanks for adding this to TSC!

PR #700, opened 2020-09-28
Add 'Execute' Capability

elsherif:
This Capability was missing and is needed to be able to set permission "Run Flow" on prep flows.

### DIFF
--- a/tableauserverclient/models/permissions_item.py
+++ b/tableauserverclient/models/permissions_item.py
@@ -31,6 +31,7 @@ class Permission:
         ViewUnderlyingData = 'ViewUnderlyingData'
         WebAuthoring = 'WebAuthoring'
         Write = 'Write'
+        Execute = 'Execute'
 
     class Resource:
         Workbook = 'workbook'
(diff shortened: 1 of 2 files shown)


tableauserverclient/models/permissions_item.py
```diff
@@ -31,6 +31,7 @@ class Capability:
         ViewUnderlyingData = 'ViewUnderlyingData'
         WebAuthoring = 'WebAuthoring'
         Write = 'Write'
+        Execute = 'Execute'
```
shinchris, line 34: Could you move this up to line 23? to keep it alphabetized

shinchris (changes requested):
Thanks @elsherif! I just have one minor comment on the change.

Also, I had to rebase to make the change into development, rather than master. It brought in an unrelated change to a sample, but you can ignore it.

tableauserverclient/models/permissions_item.py
elsherif: done!

t8y8 (approved):
🚀 

It looks like something snuck in to the samples as well, is that something you added or it's grabbing it from the branch?

shinchris (approved):

PR #704, opened 2020-10-09
File objects for publish workbook

nnevalainen:
This is a proposal for an enhancement for workbook.publish to support a file object parameter. 

## User story
As a TSC user I want to publish workbooks by providing a file object, so that I don't have to store temporary files to a file system

Specific use case
1. Ad-hoc conversion of a non-packaged workbook to a packaged workbook in order to embed supporting files
2. Ad-hoc conversion is needed because workbook is produced dynamically by populating parameters for a parametrised template workbook

## Acceptance criteria
- Workbook can be published by providing either a file path or a file object
- Parameter type (file path vs file object) is automatically determined
- When providing a file object
    - Workbook name must be always provided (can't be interpreted from file name)
    - Workbook type (.twb vs .twbx) is automatically determined by sniffing the beginning of the file object


nnevalainen:
@t8y8 @shinchris How do you feel about the concept/implementation?

t8y8:
I'm not opposed to the idea, but I need to take a deeper look.

Do I understand the primary use case as being "I need to hold a workbook xml in memory, possibly modifying it, and then eventually publish it"?

nnevalainen:
Yes that's about it. Modified xml could then be published either packaged or non-packaged without writing it to the filesystem. Thanks for looking into this!

tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -291,22 +300,32 @@ def publish(
             url += '&{0}=true'.format('asJob')
 
         # Determine if chunking is required (64MB is the limit for single upload method)
-        if os.path.getsize(file_path) >= FILESIZE_LIMIT:
-            logger.info('Publishing {0} to server with chunking method (workbook over 64MB)'.format(filename))
-            upload_session_id = Fileuploads.upload_chunks(self.parent_srv, file_path)
+        if is_file_path:
+            file_size = os.path.getsize(file)
+        else:
+            file.seek(0, os.SEEK_END)
```
t8y8, line 306: I had to look this up, but I see what’s going on here — is it possible to safely use os.fstat(fd).st_size? If not, can you extract this into a little utility function named `get_file_size` or something similar?

tableauserverclient/filesys_helpers.py
```diff
@@ -20,3 +20,15 @@ def make_download_path(filepath, filename):
         download_path = filepath + os.path.splitext(filename)[1]
 
     return download_path
+
+
+def file_is_compressed(file):
```
t8y8, line 25: This is the main hang up I have, I have a feeling that peeking into file contents just to get the extension feels wrong.

But I know that the publish requires you specify the file type and I don’t have an alternative in mind to suggest.

Curious, did you have any other ideas you rejected for this issue?

tableauserverclient/server/endpoint/fileuploads_endpoint.py
```diff
@@ -39,23 +39,28 @@ def append(self, xml_request, content_type):
...
-    def read_chunks(self, file_path):
-        with open(file_path, 'rb') as f:
-            while True:
-                chunked_content = f.read(CHUNK_SIZE)
-                if not chunked_content:
-                    break
-                yield chunked_content
+    def read_chunks(self, file):
+        is_file_path = isinstance(file, str)
+        f = open(file, 'rb') if is_file_path else file
```
t8y8, line 44: Can you preserve the `with` here?

t8y8 (changes requested):
Great idea — I understand the use case and think it’s worth pursuing.

I left some specific comments, mostly a small nit.

I do have a larger concern about how we would sprinkle a bunch of if path checks in the already somewhat if statement heavy code. I’m wondering if there’s a way to avoid that?

t8y8:
@nnevalainen Alright, a quick review! I need to ponder a bit on the if checks question, but overall it feels close/the right direction.

tableauserverclient/filesys_helpers.py
nnevalainen: I would also prefer not to specify file type in the API call but can't really get around that. As the file object is not saved we neither have any metadata to look into. Only alternative that I could think of was to pass file type as an optional parameter for the publish method and require it when dealing with file objects. The reason I didn't go for this was that I preferred to restrict the changes to function internals as opposed to changing its interface as well. Now thinking about it, adding one optional parameter probably wouldn't  hurt if you would prefer it over sniffing the file contents.

tableauserverclient/server/endpoint/workbooks_endpoint.py
nnevalainen: As I understood it, in memory binary streams such as BytesIO() don't have a file descriptor that is needed for os.fstat. I'll extract the logic to a utility function
```
from io import BytesIO()
import os

with open('file.txt', 'wb') as f:
    f.write('Hello World!'.encode())

f1 = open('file.txt')

f2 = BytesIO()
f2.write('Hello World!'.encode())

print('f1 size', os.fstat(f1.fileno()).st_size)
print('f2 size', os.fstat(f2.fileno()).st_size)

> f1 size 12
> Traceback (most recent call last):
>  File "<stdin>", line 1, in <module>
> io.UnsupportedOperation: fileno
```

nnevalainen:
> I do have a larger concern about how we would sprinkle a bunch of if path checks in the already somewhat if statement heavy code. I’m wondering if there’s a way to avoid that?

Should we try to get rid of the "if path" checks altogether and adopt duck typing approach? We could group all operations using file path together like

```
try:
    filename = os.path.basename(file)
    file_extension = os.path.splitext(filename)[1][1:]
    file_size = os.path.getsize(file)
    ...
except SpecificException:
    file_size = get_file_size() # utility function returning file object size
    ...

# Perform common oprations

```

tableauserverclient/filesys_helpers.py
t8y8: Alright, fair enough.

Let's keep the peeking approach

t8y8:
> > I do have a larger concern about how we would sprinkle a bunch of if path checks in the already somewhat if statement heavy code. I’m wondering if there’s a way to avoid that?
> 
> Should we try to get rid of the "if path" checks altogether and adopt duck typing approach? We could group all operations using file path together like
> 
> ```
> try:
>     filename = os.path.basename(file)
>     file_extension = os.path.splitext(filename)[1][1:]
>     file_size = os.path.getsize(file)
>     ...
> except SpecificException:
>     file_size = get_file_size() # utility function returning file object size
>     ...
> 
> # Perform common oprations
> ```

Yes, I like that approach, if we need to add another method in the future, or update something, we can keep it contained to the try block. Ship it!

shinchris:
Other than the requested changes from @t8y8, this looks good to me.
Would there be use cases for allowing this for datasource.publish as well?

t8y8:
Yeah, I think the use case is still valid for TDS/TDSX, and maybe even flows (TFL/TFLX).

@nnevalainen would you be interested in doing that as well? 
You can say no, or at a minimum defer it to a second PR :)

nnevalainen:
I can take a look at that as well in a separate PR after finishing this one

nnevalainen:
Alright finished with the the changes. Let me know what you think

shinchris (approved):

t8y8 (approved):
🚀

shinchris:
@nnevalainen forgot to ask earlier, but would you be able to read and sign our [CLA](https://tableau.github.io/contributing.html)?

PR #705, opened 2020-10-09
Updates datasource_item with new fields

shinchris:
Datasource item has been updated to support some new fields. Below are the 6 new fields and the endpoints it affects:
**ask_data_enablement**: publish, update, get_by_id 
**description**: publish, get, get_by_id (only returned if it's set to something)
**encrypt_extracts**: update, get, get_by_id
**has_extracts**: get, get_by_id (2020.2+, v3.8+ with next MR)
**use_remote_query_agent**: publish, get, get_by_id
**webpage_url**: get, get_by_id

This change also alphabetizes the fields in datasource_item to make it a bit more readable. Doc updates coming


t8y8 (approved):
🚀

PR #707, opened 2020-10-19
Adds new datasource fields to docs

shinchris:
This PR adds all the new datasource fields to our docs.

I'm not sure if we should include information about which fields can and cannot be used for which endpoints. For example, some fields are update-able but not publish-able. Some are returned by get, some are returned by list.


bcantoni (approved):
+1 I think this looks good. Also double-checked and confirmed all the links.

I think the level of detail we have here is good for now; we can always add more if questions come up.

bcantoni:
For context, these doc changes go with code changes in #705 and #708.

PR #711, opened 2020-10-28
Removes constraint that required interval when updating schedule

shinchris:
Attributes of a schedule can be updated without needing an interval attribute. We previously had a check to make sure that a schedule item contained an interval item when updating. This PR removes that check.

### DIFF
--- a/tableauserverclient/server/endpoint/schedules_endpoint.py
+++ b/tableauserverclient/server/endpoint/schedules_endpoint.py
@@ -43,9 +43,6 @@ class Schedules(Endpoint):
         if not schedule_item.id:
             error = "Schedule item missing ID."
             raise MissingRequiredFieldError(error)
-        if schedule_item.interval_item is None:
-            error = "Interval item must be defined."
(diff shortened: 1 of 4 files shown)


bcantoni:
@shinchris overall looks good. Are there any test changes or additions to consider? For example, do we already cover setting the state? (Like the example from the bug reporter.)

shinchris:
We have a general unit test that verifies the parsing of an update response, but doesn't cover setting the state. We can add state to that existing test, and also add another test case that tests the removal of the constraint.

I also just noticed that we are missing schedules.get_by_id(), which we can add in a separate PR.

t8y8:
What was the issue here? I don't think a schedule can be interval-less.

Is the problem that update requests required it in addition to creates?

shinchris:
The main issue is that the schedules.get call does not return interval. So when you try to update the same object with a changed state, it would fail in TSC because it was interval-less. Server doesn't have that constraint for update though - partial, interval-less updates should go through.

PR #713, opened 2020-10-30
Add all fields for users.get

shinchris:
This adds supports for getting all fields for users.get(). For now, this is a 'private' request option (only used by users.get) because we don't have the parser logic in place to parse all the fields we get back in the other item classes.

This is to fix the fullname and email being None, despite it being parsed and documented for users.get. Once we have a solution for expanding our parsers, we can expose this publicly or just add it by default for all the get calls.


bcantoni (approved):
+1

PR #714, opened 2020-10-31
File objects for publish datasource

nnevalainen:
Extend support for publishing file objects to datasource.publish as discussed in #704. I followed the same approach as with workbooks. Let me know what you think :)


tableauserverclient/server/endpoint/datasources_endpoint.py
```diff
@@ -173,22 +173,37 @@ def delete_extract(self, datasource_item):
...
-            datasource_item.name = os.path.splitext(filename)[0]
-        if file_extension not in ALLOWED_FILE_EXTENSIONS:
-            error = "Only {} files can be published as datasources.".format(', '.join(ALLOWED_FILE_EXTENSIONS))
+        except TypeError:
+
+            if not datasource_item.name:
+                error = "Datasource item must have a name when passing a file object"
+                raise ValueError(error)
+
+            file_extension = 'tdsx' if file_is_compressed(file) else 'tds'
```
t8y8, line 201: I think we might have a problem here -- you can directly publish a `tde` or `hyper` file, both of which are actually compressed file formats themselves (I'm not sure if they're both zip or some other format)

t8y8: This might not be an issue practically -- if they're loading a tde or hyper file in memory it might just fail on publish.

I wonder if there is a way to warn the caller early though.

nnevalainen: I modified the file_peeking to return file_type instead of boolean value and extended it to identify .xml, .tde, and .hyper. I couldn't find official sources for .tde or .hyper file signatures but was able to deduce them by creating few different files and peeking their content. 

Now if users provides '.tde' or '.hyper' file they are informed that these are not-supported file types. Additionally, they will be shown an error if they're providing file types we can't identify. Let me know what you think

tableauserverclient/filesys_helpers.py
```diff
@@ -30,13 +30,30 @@ def get_file_object_size(file):
     return file_size
 
 
-def file_is_compressed(file):
-    # Determine if file is a zip file or not
+def get_file_type(file):
     # This reference lists magic file signatures: https://www.garykessler.net/library/file_sigs.html
 
-    zip_file_signature = b'PK\x03\x04'
+    magic_bytes = {
```
t8y8, line 36: Nit: Make this all CAPS since it's kind of like a constant.

tableauserverclient/filesys_helpers.py
```diff
@@ -30,13 +30,30 @@ def get_file_object_size(file):
...
+        'zip': bytes.fromhex("504b0304"),
+        'tde': bytes.fromhex("20020162"),
+        'xml': bytes.fromhex("3c3f786d6c20"),
```
t8y8, line 39: A small comment to remind readers that our twb/tds are XML

t8y8 (approved):
Two small nits but approved.

The tests are excellent, thank you!

nnevalainen:
Made the changes pointed out. Should really learn to use pycodestyle locally. I've basically run into broken travis builds every time I've pushed something out 😅  

It could be nice to force some local testing in the form of pre-push hooks or similar.

t8y8 (approved):
🚀

PR #701, opened 2020-09-29
Doc updates including new page for troubleshooting

bcantoni:
This PR includes a batch of doc updates including:
- add new troubleshooting page showing how to capture REST API traffic
- rewrite sign in / sign out page to be more clear, in particular server vs online, and to - mention switch_site
- dev-guide, index, samples, versions: some general edits and improvements
- api-ref: added auth.switch_site (from #655)

Nothing in here is tied to any code changes, so it can be merged into gh-pages when ready. I'll squash all the commits before merging.

PR #715, opened 2020-11-01
Update documentation with switch site and Django style filters

jorwoods:
Updates documentation to have details for site switching and the django style filtering


bcantoni:
@jorwoods thanks for adding all this. I actually already added docs for switch site in #701, so I'm not sure why they aren't in your branch. You might need another merge from (or rebase on top of) gh-pages.

jorwoods:
@bcantoni It was indeed out of date. I did another merge and removed my duplicate entries for the switch site documentation

bcantoni:
@jorwoods thanks for this contribution!

Issue #740, opened 2020-11-19
Updating workbook fields (such as name) is failing with "Payload is either malformed or incomplete" error


bcantoni:
TSC: latest (v0.14.0)
Tableau server versions tested:
* exhibit the failure: 2019.1, 2019.3
* works correctly: 2020.4

Sample script:
```py
tableau_auth = TSC.TableauAuth(username, password, site_id=site_url)
tableau_server = TSC.Server(server_url, use_server_version=True)
tableau_server.add_http_options({'verify': False})

with tableau_server.auth.sign_in(tableau_auth):
	workbook_item = tableau_server.workbooks.get_by_id(workbook_luid)
	new_name = "{}_{}".format(workbook_item.name, "Test")
	workbook_item.name = new_name
	tableau_server.workbooks.update(workbook_item)
```

Result:
```
Traceback (most recent call last):
  File "c:\dev\- Tests\workbook_actions.py", line 30, in <module>
    main()
  File "c:\dev\- Tests\workbook_actions.py", line 27, in main
    tableau_server.workbooks.update(workbook_item)
  File "C:\Python39\lib\site-packages\tableauserverclient\server\endpoint\endpoint.py", line 127, in wrapper
    return func(self, *args, **kwargs)
  File "C:\Python39\lib\site-packages\tableauserverclient\server\endpoint\workbooks_endpoint.py", line 107, in update
    server_response = self.put_request(url, update_req)
  File "C:\Python39\lib\site-packages\tableauserverclient\server\endpoint\endpoint.py", line 93, in put_request
    return self._make_request(self.parent_srv.session.put, url,
  File "C:\Python39\lib\site-packages\tableauserverclient\server\endpoint\endpoint.py", line 55, in _make_request
    self._check_status(server_response)
  File "C:\Python39\lib\site-packages\tableauserverclient\server\endpoint\endpoint.py", line 70, in _check_status
    raise ServerResponseError.from_response(server_response.content, self.parent_srv.namespace)
tableauserverclient.server.endpoint.exceptions.ServerResponseError:

        400000: Bad Request
                Payload is either malformed or incomplete	
```

bcantoni:
This is fixed in the current `development` branch; will close the issue once the next release happens.

shinchris:
Fix (#741) went in v0.14.1 release.

PR #741, opened 2020-11-19
Fixes data_acceleration field always in workbook update payload

shinchris:
Improves data_acceleration_config checks when generating workbook update requests. WorkbookItem itself enforces valid values, so it was redundant to have checks in the request_factory. 

Addresses #740


bcantoni (approved):
I tested on a 2019.1.19 server to first confirm the original failure and then confirmed this branch works correctly. In the fixed version the whole `<dataAccelerationConfig accelerationEnabled="none" accelerateNow="none" />` element is omitted as expected from the update PUT.

shinchris:
Thanks for confirming!

PR #743, opened 2020-11-20
Improve debug logging by including put/post request contents (limited to first 1000 chars)

bcantoni:
- show contents of methods such as put, post
- for empty responses, don't try to print them (avoid [Truncated File Contents] which might appear like an error)

### DIFF
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -50,6 +50,10 @@ class Endpoint(object):
         if content is not None:
             parameters['data'] = content
 
+        logger.debug(u'request {}, url: {}'.format(method.__name__, url))
+        if content:
+            logger.debug(u'request content: {}'.format(content))
+
         server_response = method(url, **parameters)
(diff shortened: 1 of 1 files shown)


tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -50,6 +50,10 @@ def _make_request(self, method, url, content=None, request_object=None,
         if content is not None:
             parameters['data'] = content
 
+        logger.debug(u'request {}, url: {}'.format(method.__name__, url))
+        if content:
```
shinchris, line 54: For publishing requests, this would print out the whole file content. I think we should limit the length, like its done [here for workbook publish](https://github.com/tableau/server-client-python/blob/master/tableauserverclient/server/endpoint/workbooks_endpoint.py#L348) and also remove that duplicate log line

bcantoni: Good point, I'll work on adjusting that!

shinchris (approved):
🚀

Issue #732, opened 2020-11-17
Filter query breaks in 0.14 with Tableau Server 2019.4


barscm:
When connecting to Tableau Server 2019.4, tableauserverclient works fine on 0.13, however when moving to 0.14 the Filter breaks. When connecting to Tableau Server 2020.2, all functionality works fine. Is this a regression for older versions of Tableau Server? There was a change in the ChangeLog referring to RequestOptions so I wondered if that could be the cause. I note there are mutliple issues fixed by this change, which relate to url encoding, however I would expect this code should still work, as it does not use any reserved characters.

With 0.13:
`(tsc) ...\tsc>py filter_test.py`
Output:
`0`

With 0.14:
`(tsc_014) ...\tsc_014>py ..\tsc\filter_test.py`
Output:
```
(tsc_014) ...\tsc_014>py ..\tsc\filter_test.py
Traceback (most recent call last):
  File "..\tsc\filter_test.py", line 12, in <module>
    ds_list, p = server.datasources.get(req_option)
  File "...\tsc_014\lib\site-packages\tableau
serverclient\server\endpoint\endpoint.py", line 127, in wrapper
    return func(self, *args, **kwargs)
  File "...\tsc_014\lib\site-packages\tableau
serverclient\server\endpoint\datasources_endpoint.py", line 40, in get
    server_response = self.get_request(url, req_options)
  File "...\tsc_014\lib\site-packages\tableau
serverclient\server\endpoint\endpoint.py", line 85, in get_request
    return self._make_request(self.parent_srv.session.get, url, auth_token=self.
parent_srv.auth_token,
  File "...\tsc_014\lib\site-packages\tableau
serverclient\server\endpoint\endpoint.py", line 55, in _make_request
    self._check_status(server_response)
  File "...\tsc_014\lib\site-packages\tableau
serverclient\server\endpoint\endpoint.py", line 70, in _check_status
    raise ServerResponseError.from_response(server_response.content, self.parent
_srv.namespace)
tableauserverclient.server.endpoint.exceptions.ServerResponseError:

        400065: Bad Request
                None
```

With `...\tsc\filter_test.py`:
```python
import tableauserverclient as TSC

tableau_auth = TSC.PersonalAccessTokenAuth(token_name="poc", personal_access_token="ACCESS_TOKEN", site_id="MY_SITE")  # redacted
server = TSC.Server('https://tableau.example.com', use_server_version=True)  # redacted

[...]

PR #745, opened 2020-11-20
Server versions before 2020.1 do not accept encoded query param delimiters

shinchris:
Request options are only applied for GET calls, so I've moved the query param generation to the GET-specific request call.
In there, it checks the server version to determine how to append the query params:
- 2020.1 and above - let requests library handle all query params and encoding
- 2019.4 and below - we manually generate the query param string - do not encode delimiters

Addresses #732


t8y8 (approved):
🚀

PR #757, opened 2020-12-09
Fixes issue #754 by moving file read logic inside generator

### LINKED ISSUE #754: Error publishing large twbx files
While publishing a large twbx file with latest 0.14 version of tableau-server-client using the below code,

wb_item = server.workbooks.publish(wb_item,target_file,'Overwrite')

I am getting the following error.

 File "/usr/local/lib/python3.9/site-packages/tableauserverclient/server/endpoint/fileuploads_endpoint.py", line 45, in read_chunks

  chunked_content = file.read(CHUNK_SIZE)

ValueError: read of closed file

The same code works for 0.13 and below versions. And also it works for smaller workbooks, even with extracts.

shinchris:
Issue was that read_chunks() returns a generator, so the function doesn't actually get executed until we try to iterate through the collection. By the time we try to iterate through it, the file is already closed in the generator method. So moving the open logic into the generator itself will prevent this.

Fixes #754

### DIFF
--- a/tableauserverclient/server/endpoint/fileuploads_endpoint.py
+++ b/tableauserverclient/server/endpoint/fileuploads_endpoint.py
@@ -40,10 +40,18 @@ class Fileuploads(Endpoint):
         return FileuploadItem.from_response(server_response.content, self.parent_srv.namespace)
 
     def read_chunks(self, file):
+        file_opened = False
+        try:
+            file_content = open(file, 'rb')
+            file_opened = True
+        except TypeError:
+            file_content = file
 
         while True:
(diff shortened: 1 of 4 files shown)


t8y8:
I think it's a smell and a sign that we aren't using `with` block or a `try/finally` in the right place, but this fix works for now if you need to get it out quickly. I remember thinking about that when reviewing the PR that changed the behavior but I guess I missed it.

bcantoni:
+1 changes look good. It works correctly now with my test script for .twbx workbooks under and over 64MB.

bcantoni (approved):

PR #767, opened 2020-12-29
Add example of using pre-commit hook to check code style

bcantoni:


shinchris:
This looks great!
Can we add the pre-commit hook to the repo itself to have it run by default on all commits?

bcantoni:
@shinchris I'm not totally sure whether or how to do that as part of the repo. Most of what I've read points to hooks being client-side only. We could look more at that in the future or consider using the [pre-commit](https://pre-commit.com/) package which seems quite popular.

shinchris (approved):

Issue #730, opened 2020-11-14
Group Item; add ability to poulated domain_name


graham-marcus:
In order to create active directory group the domain_name is not an input field available when creating the group item to add.
In group_item.py the option is not available  can  this be added as optional
def __init__(self, name=None):
        self._domain_name = None

graham-marcus:
Take this through to use
newgroupitem = server.groups.create_AD_group(newgroupitem)
where the newgroupitem object has the group name, import, domain, minimum site role options available to populate

bcantoni:
Thanks for reporting/suggesting this -- I'll tag as a feature enhancement.

shinchris:
Fix (#770) released with v0.15.0.

PR #770, opened 2021-01-04
Improves group creation for both local and AD

shinchris:
- Adds domain name to group_item constructor (#730)
- Adds enum for license modes
- Removes licenseMode from local group creation (not used by server, need to also update rest api docs)
- Fixes field name for siteRole (need to also update rest api docs)

TSC documentation - #771 

Internal 1210383


t8y8 (approved):
Do we have tests for AD groups?

t8y8:
🚀

PR #772, opened 2021-01-08
Fixes groups.update to match server requests/responses

shinchris:
Included in this PR:

1. Response parsing: [Rest api docs](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_usersgroups.htm#update_group) are wildly wrong about the local group update request/response (will file a doc bug) starting 3.9+ (2020.3+). Import element is returned starting 3.9+, for both local and AD groups. Query calls will return a domain element for all versions, but won't return it for create/update calls.
2. Update endpoint was taking in a `default_site_role`, which would always override the minimum_site_role field of the group_item with Unlicensed. TSC standard is to set that value in the resource item itself.
3. Updating local groups don't support async, so it would error in the parser if you tried.
4. Allows local groups to update minimum site role now

### DIFF
--- a/tableauserverclient/server/request_factory.py
+++ b/tableauserverclient/server/request_factory.py
@@ -299,17 +299,30 @@ class GroupRequest(object):
         return ET.tostring(xml_request)
 
     def update_req(self, group_item, default_site_role=None):
+        # (1/8/2021): Deprecated starting v0.15
         if default_site_role is not None:
+            import warnings
+            warnings.simplefilter('always', DeprecationWarning)
+            warnings.warn('RequestFactory.Group.update_req(...default_site_role="") is deprecated, '
+                          'please set the minimum_site_role field of GroupItem',
+                          DeprecationWarning)
             group_item.minimum_site_role = default_site_role
+
         xml_request = ET.Element('tsRequest')
         group_element = ET.SubElement(xml_request, 'group')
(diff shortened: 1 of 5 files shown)


tableauserverclient/server/request_factory.py
```diff
@@ -299,17 +299,30 @@ def create_ad_req(self, group_item):
...
-            project_element.attrib['siteRole'] = group_item.minimum_site_role
-            project_element.attrib['grantLicenseMode'] = group_item.license_mode
+        if group_item.domain_name is not None and group_item.domain_name != 'local':
+            # Import element is only accepted in the request for AD groups
+            import_element = ET.SubElement(group_element, 'import')
+            import_element.attrib['source'] = "ActiveDirectory"
+            import_element.attrib['domainName'] = group_item.domain_name
+            import_element.attrib['siteRole'] = group_item.minimum_site_role
+            if group_item.license_mode is not None:
+                import_element.attrib['grantLicenseMode'] = group_item.license_mode
```
jacalata, line 321: Did we still have an underlying REST bug on changing the license mode?

shinchris: Yes, there is still that bug where you can't change license mode from onLogin -> onSync. Internal 1214605

jacalata (approved):

PR #777, opened 2021-01-18
Update site properties and functions

tjones-commits:
Tableau server 2020.3 api version 3.9 makes available several configuration options for the site endpoint. The changes here are meant to align the site object and its functions with the api. 

Linked to Issue https://github.com/tableau/server-client-python/issues/776

PR #778, opened 2021-01-19
Add functionality for tierLicenseCapacity on sites

jorwoods:
This adds support to the SiteItems to allow it to be used to manage the license level specific counts.


tableauserverclient/models/property_decorators.py
```diff
@@ -8,6 +8,8 @@
     # In case we are in python 3 the string check is different
     basestring = str
 
+literal_range = range
```
t8y8, line 11: I didn't realize I was shadowing the builtin before, I actually prefer just renaming the old variable, maybe call it "valid_int_range" and then let range stay the function

t8y8 (changes requested):

tableauserverclient/models/property_decorators.py
jorwoods: Renamed the keyword argument and removed the copy of builtin.

t8y8 (approved):
🚀

shinchris:
#777 covers tierLicenseCapacity support, though it doesn't include the property_is_int annotation and user-quota limitation. Do you mind waiting until that one is checked in? 

Also I think the integer annotation on these are a bit tricky - I'm not sure if there is a limit to the value and -1 is a valid value to indicate reset

jorwoods:
This can wait until #777 is in. I don't like having the hardcoded upper bound in there either. I can replace it with a `sys.maxsize` and replace the lower limit with -1.

shinchris:
That works! I would keep the lower limit at 0 and pass in -1 as the `allowed` parameter, to keep it consistent with how it's done in request_options.py.

jorwoods:
@shinchris Question regarding the -1 to indicate reset: I had tested on my own server by passing `None` to the tiered capacities when I wanted to remove them. Should it be -1 instead? It seems that unittest doesn't like trying to evaluate a range with `sys.maxsize` as a bound (even though via ipython it has the same run time as any other range evaluation). I had to add a range to `property_is_int` to be able to add `None` to the allowed kwarg. 

If you think it should be able to accept None, I can either:

1. Modify `property_is_int` to evaluate in a different order to be able to remove the range restriction, 
2. Add a `nullable_int` type checker
3. Remove `None` as an acceptable value, and change the check against user quota

shinchris:
@jorwoods I think it should be -1 instead. Passing in `None` does remove the tier capacity limits, but it must be all 3 tiers at the same time. For example, if you just wanted to set tierViewerCapacity back to its default, you need to pass in -1 for that one value. Setting just the 1 tier to `None` (omitted from api request) will not be accepted by server.

For the `property_is_int` decorator, perhaps we can make it handle a no-max range. If the max value in the tuple is None or some specific value, we don't check the upper bound.

tableauserverclient/server/endpoint/sites_endpoint.py
```diff
@@ -67,6 +67,14 @@ def update(self, site_item):
                 error = 'You cannot set admin_mode to ContentOnly and also set a user quota'
                 raise ValueError(error)
 
+        reset = (None, -1,)
```
jacalata, line 70: Why is this variable named "reset"? It could have always been empty, right?

jacalata (approved):

PR #784, opened 2021-01-26
Fetch project owner on get

jorwoods:
Adds `owner_id` to `ProjectItem` and populates it via the `projects.get` endpoint. Adds stubs for updating owner via create/update calls, but leaves them commented out pending server-side enhancement. Raises a `NotImplementedError` when trying to set `owner_id` on a `ProjectItem` as that is currently not supported server side.

### DIFF
--- a/tableauserverclient/models/project_item.py
+++ b/tableauserverclient/models/project_item.py
@@ -11,13 +11,14 @@ class ProjectItem(object):
         LockedToProject = 'LockedToProject'
         ManagedByOwner = 'ManagedByOwner'
 
-    def __init__(self, name, description=None, content_permissions=None, parent_id=None):
+    def __init__(self, name, description=None, content_permissions=None, parent_id=None, owner_id=None):
         self._content_permissions = None
         self._id = None
         self.description = description
         self.name = name
         self.content_permissions = content_permissions
         self.parent_id = parent_id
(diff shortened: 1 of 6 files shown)


t8y8:
I think grabbing owner and populating it is great, the NotImplemented exception is fine but I'm not a fan of the comments hanging out in the code.

Would you be willing to nix the comments but keep the rest :)

jorwoods:
I can understand not liking the commented code there. I did it with the idea of referencing what needs to be enabled once the issue is resolved, but removed it.

tableauserverclient/models/project_item.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, name, description=None, content_permissions=None, parent_id=N
         self.name = name
         self.content_permissions = content_permissions
         self.parent_id = parent_id
+        # self.owner_id = owner_id
```
t8y8, line 21: Missed one :)

t8y8 (changes requested):

tableauserverclient/models/project_item.py
jorwoods: Pushed patch

t8y8 (approved):
🚀

shinchris:
Thanks @jorwoods!

PR #787, opened 2021-02-01
Type hinting

jorwoods:
Adds type hinting to package. I have noticed difficulty myself as well as confusion with others in github issues on what gets returned or what values are expected. This should hopefully make usage of the package a little clearer.

### DIFF
--- a/tableauserverclient/models/connection_credentials.py
+++ b/tableauserverclient/models/connection_credentials.py
@@ -1,4 +1,8 @@
 from .property_decorators import property_is_boolean
+import xml.etree.ElementTree as ET
+from typing import Dict, TypeVar
+
+T = TypeVar('T', bound='ConnectionCredentials')
 
 
 class ConnectionCredentials(object):
@@ -9,7 +13,8 @@ class ConnectionCredentials(object):
 
     """
 
-    def __init__(self, name, password, embed=True, oauth=False):
+    def __init__(self, name: str, password: str, embed: bool = True,
+                 oauth: bool = False) -> T:
         self.name = name
         self.password = password
         self.embed = embed
@@ -21,7 +26,7 @@ class ConnectionCredentials(object):
 
     @embed.setter
     @property_is_boolean
-    def embed(self, value):
+    def embed(self, value: bool):
         self._embed = value
 
     @property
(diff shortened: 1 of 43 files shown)


t8y8:
If we're going to do this, I think we should tackle it in 3 parts:

1. Get a type checker running in our CI pipeline (Did you use mypy to check these or some other?)
2. Add the type hints places
-- I'm not sure we want to type hint all the things; I like starting with just endpoints and not any of the models. I also have some other thoughts we might want to work through (We have a bunch of decorators that do psuedo-type-checking/hinting already, should we keep them since they work for runtime issues, or should we nix them.
3. Resolve / refactor any issues the type hints show a problem or a smell.

@jorwoods would you be interested in taking a crack at #1? If not, we were talking about moving to github actions internally so maybe now is the time to do that, and add a mypy run.

tableauserverclient/models/connection_credentials.py
```diff
@@ -30,11 +35,11 @@ def oauth(self):
 
     @oauth.setter
     @property_is_boolean
-    def oauth(self, value):
+    def oauth(self, value: bool):
         self._oauth = value
 
     @classmethod
-    def from_xml_element(cls, parsed_response, ns):
+    def from_xml_element(cls, parsed_response: ET.ElementTree, ns: Dict) -> T:
```
t8y8, line 42: There's something going on here, when I run mypy on your PR, there are a bunch of complaints about ElementTree not having a `get` method. I think that's because those are `Element` not `ElementTree` but I thought it should be smart enough to infer that is what `.find` returns?

jorwoods: I'm not entirely sure what is going on here. I admittedly should have started with mypy, but instead I did it by looking through some of my application code. Mypy complains about a wide array of things when I try to run it, and it will take some time to sort out. I wanted to get a PR up, perhaps too early, to get a reading on whether or not this kind of thing would be accepted. Seems like so if I can iron out the issues in it.

t8y8: Now that we only support 3.5 and above, I would be open to type hints for sure.

Maybe we want to start with just 1 vertical, like DatasourceItem, DatasourceEndpoints, and use that vertical to sort out some of the general patterns we may need (like Element/ElementTree) or if we want type_checking IF statements etc

shinchris: Yeah, I agree with Tyler on starting out with 1 vertical slice.
Type hints would be great to have!

t8y8:
@jorwoods because I'm stalking your fork and I think we both googled the same issue, I guess `# type: ignore` is the best we can do for now with all of our properties. It looks like we might want to rename some of the double-named functions (like `content_permissions`).

Also let's use `Optional[Thing]` over `Union[Thing, None]`

There are also some strange type hints generated, like Union[str,str], I think that can just be a happy str :)

jorwoods:
@t8y8 Yeah, I don't like having to ignore those errors, but it definitely seems that there is not another option.

I've got datasource item/endpoint hinted in my fork, and a few sticking points figured out. The auto-generated hints aren't sensible, so I'll be moving things from this branch over to the other branch in phases.

As for the `content_permissions` errors, if we're thinking of the same issue, that was resolved by simply rearranging the code and moving the setter to immediately following the getter instead of having other entries between them.

PR #794, opened 2021-02-11
[Subscriptions] Add new fields and ability to update

shinchris:
Went through subscriptions and added some new fields, as well as a new endpoint for updating them (internal 1201628).
The commits are separated by steps I took to make the changes.

Working on doc changes now.


jacalata (approved):
looks good to me. Would it make sense to have a "user object" attribute on the subscription, instead of just referring to the user-name and user-id?

shinchris:
That's a good point, though it would be an incomplete user object (only id and name). But this reminded me that we usually only keep the user/owner id, since you can query the user by the id for the full list of attributes - same for schedule name.

PR #795, opened 2021-02-11
[Docs] Updates documentation for subscriptions

shinchris:
Doc updates go with #794 

Adds the new fields in there, as well as a new update endpoint.
The example code snippet shows all the fields you can set for the create and update methods.

### DIFF
--- a/docs/api-ref.md
+++ b/docs/api-ref.md
@@ -2569,15 +2569,32 @@ SubscriptionItem(subject, schedule_id, user_id, target)
 Name | Description
 :--- | :---
 `id` |   The id of the subscription on the site.
+`attach_image` | Setting this `True` will cause the subscriber to receive mail with .png images of workbooks or views attached to it. By default, this value is set to `True`. If subscriptions to attachments are disabled in Tableau server or site settings, then making a request that sets `attach_image` to `False` will cause an error.
+`attach_pdf` | Setting this `True` will cause the subscriber to receive mail with a .pdf file containing images of workbooks or views attached to it. By default, this value is set to `False`. If subscriptions to attachments are disabled in Tableau server or site settings, then making a request that sets `attach_pdf` to `True` will cause an error.
+`message` | The text body of the subscription email message.
+`page_orientation` | The orientation of the pages produced when `attach_pdf` is `True`. If this parameter is not present the page orientation will default to `Portrait`. See chart below for a full list of values.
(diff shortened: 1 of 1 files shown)


docs/api-ref.md
```diff
@@ -2569,15 +2569,32 @@ SubscriptionItem(subject, schedule_id, user_id, target)
 Name | Description
 :--- | :---
 `id` |   The id of the subscription on the site.
+`attach_image` | Setting this `True` will cause the subscriber to receive mail with .png images of workbooks or views attached to it. By default, this value is set to `True`. If subscriptions to attachments are disabled in Tableau server or site settings, then making a request that sets `attach_image` to `False` will cause an error.
```
bcantoni, line 2572: Should this end instead with "...making a request that sets attach_image to **True** will cause an error"?

bcantoni (changes requested):
Comparing this to the matching code changes, it looks like `SubscriptionItem` does not actually have `schedule_name` or `user_name`, just the "_id" for each of those. Am I missing something or should they be dropped from the list here?

docs/api-ref.md
shinchris: Thanks for catching this. This is directly from the rest api doc and it's a bit confusing, but it is correct - though I can improve the wording a bit here. Image is not considered an 'attachment', so 'subscriptions to attachments' setting refers to pdf only.

In other words, you can never have both `attach_image` and `attach_pdf` set to false. If you have 'subscription attachments' site setting disabled, you can only get image attachments (can't set `attach_pdf` to true). Which also means that `attach_image` can't be set to false in this case.

shinchris:
Ah yes, `schedule_name` and `user_name` were removed after making this doc PR, as per review feedback. Will update this PR.

docs/api-ref.md
jacalata: It sounds like we shouldn't even expose this option, because it's entirely controlled by site/server settings. Is there any case where the user has a choice of valid values to use?

jacalata: @shinchris ?

shinchris: @jacalata it's not entirely controlled by site settings. If site settings allows subscription attachments, these fields give users two choices: 1) get both image and pdf, or 2) get image only.

shinchris: Looking more at this, there is no server/site setting for subscription attachments. Made the field descriptions more clear.
To sum it up, user choices: turn both image/pdf on, turn pdf on but image off

PR #796, opened 2021-02-12
[Tasks] Translate task type from server to TSC enum

shinchris:
We have enums in task_item.py to represent the 2 task types supported by the REST API - these are used in the URLs.
However, the server response returns different values, so we need to translate them into TSC enums. 

Not quite sure why the test xml responses had different values - checked old versions of tableau server but didn't see them used anywhere.


t8y8 (approved):
🚀

jacalata:
So the server is sending 'MaterializeViews'? It was called that for a period during development but renamed before release, might be why?

shinchris:
Ah that's right, used to be materialized views before data acceleration

PR #797, opened 2021-02-12
[Docs] Adds docs for groups.update

shinchris:
Goes with #772

### DIFF
--- a/docs/api-ref.md
+++ b/docs/api-ref.md
@@ -1144,6 +1144,65 @@ None. The user is removed from the group.
 <br>
 <br>
 
+
+#### groups.update
+
+```py
+groups.update(group_item, as_job=False)
+```
+
+Updates a group in Tableau Server.
+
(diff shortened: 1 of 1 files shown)


bcantoni (changes requested):
New content looks good. I tested the sample on a couple servers.

Since we're showing how to use license_mode and minimum_site_role, I think we should add those under the [GroupItem class](http://127.0.0.1:4000/server-client-python/docs/api-ref#groups). Under license_mode we should explain the 2 options onLogin and onSync. (Possible values for minimum_site_role are already explained under UserItem.)

shinchris:
Thanks @bcantoni! 
Those were added with #771 and are live in our [docs](https://tableau.github.io/server-client-python/docs/api-ref#groups).

bcantoni (approved):
Ah, that's great! I had checked out this branch so I had missed it. Looks good to go.

PR #798, opened 2021-02-12
Removes travis and adds linting/testing into github action

shinchris:
With every push and pull requests, it should now run our tests and linter automatically. See the checks below!
Runs much faster than travis so having checks on both push and pull_request should be okay.


bcantoni:
@shinchris overall this looks great. Two things to consider:
1. Now that Python 3.9 is out, let's add that to the test matrix. Technically 3.5 is EOL but I would leave it on there for now just to make sure we don't accidentally break.
2. For the setup.py changes, will that change anything else in the way people use that now? For example does `python setup.py test` still work the same?

t8y8:
Since it's fast it might not matter, but my OCD is sad that everything is duplicated -- what's the reason for both PR and Push?

Also, since we're thinking about it, I think I will add a mypy step in, but it will allow failures (there's a `continue-on-error` option)

shinchris:
@bcantoni thanks for pointing these out. I added 3.9 to the script, and yeah I wouldn't mind keeping 3.5 for a bit longer unless it starts to cause problems. 
Also updated setup.py to work with the old way as well. `python setup.py test` is deprecated, but we can phase that out so I added that back in - i'll update the developer guide docs

@t8y8 no reason other than seeing more green checks 😄 but yeah I think it's excessive, so I changed it to push only. And mypy step would be great - not too familiar with it yet but looks nice to have

jacalata (approved):
Nice!

PR #802, opened 2021-02-17
Add Mypy to CI runs

t8y8:
Add a mypy run, just on the main project for the moment, but ignore errors, so it won't need to 'pass'.

@jorwoods is working on a first run, I think once we have the main set of errors resolved and and endpoint checking, we can flip this to required, but it should still pass even of we don't add annotations everywhere.

Then we we can take a "leave better than you found it" approach for adding new hints as we add other features or fix other bugs


shinchris (approved):
Nice addition!

PR #833, opened 2021-04-21
[WIP] Dqws

t8y8:
This is badly a wip but it's my hackathon

right now reads and writes work for dbs, need tests and to add the other types, and probably stop hard coding strings and etc


t8y8:
@ugamarkj

PR #835, opened 2021-04-21
Format the world with black

t8y8:
Using black, format everything with a line length of 120.

This doesn't include workflow/actions to automate this as a pre-commit hook, but that can come after some more investigation


jacalata:
skimmed it - mostly newlines and double quotes, it looks like

PR #836, opened 2021-04-21
Data Quality Warning Support

t8y8:
Add DQW support to TSC.

DataQualityWarnings are a bit weird -- they are something like permissions where their url depends on the content type receiving the dqw, but they have a top level path.

You can also query and delete them by ID, but there are no list calls for them, so I've ignore those methods for now.


ugamarkj:
I'll be a very happy person when this gets merged =)

jacalata:
so much code 😅 
looks ok to me

PR #843, opened 2021-05-05
841 enhancement rename datasource

annematronic:
Changes include:

- Update request factory to add name attribute when present.  
- Update Datasource tests to assert name (and some other properties that were not already asserted)
- Updated other Datasource tests which were passing 'project_id' and 'name' in the incorrect order to the DatasourceItem constructor

### DIFF
--- a/test/test_datasource.py
+++ b/test/test_datasource.py
@@ -1,6 +1,8 @@
 import unittest
+import sys
 from io import BytesIO
 import os
+sys.path.append('d:\\dev\\git\\server-client-python\\.eggs\\requests_mock-1.9.1-py3.7.egg')
 import requests_mock
 import xml.etree.ElementTree as ET
 from zipfile import ZipFile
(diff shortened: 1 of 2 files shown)


test/test_datasource.py
```diff
@@ -1,6 +1,8 @@
 import unittest
+import sys
 from io import BytesIO
 import os
+sys.path.append('d:\\dev\\git\\server-client-python\\.eggs\\requests_mock-1.9.1-py3.7.egg')
```
shinchris, line 5: Is this necessary for this test to run properly?

shinchris (commented):
Minor comment in test, but otherwise looks good!

test/test_datasource.py
annematronic: Nope!  Meant to remove that, thanks for catching.

shinchris (approved):

PR #847, opened 2021-05-14
Fixes revision limit field not propagating to the request

shinchris:
Revision limit was not set from the constructor - user couldn't construct a new site item with a specific revision limit.

### DIFF
--- a/tableauserverclient/models/site_item.py
+++ b/tableauserverclient/models/site_item.py
@@ -31,7 +31,7 @@ class SiteItem(object):
         disable_subscriptions=False,
         subscribe_others_enabled=True,
         revision_history_enabled=False,
-        revision_limit=None,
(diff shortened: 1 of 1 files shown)


shinchris:
Github actions having issues https://www.githubstatus.com/incidents/kdvly1jxxdpl

t8y8:
Shouldn't the setter on line 188 allow this?

shinchris:
> Shouldn't the setter on line 188 allow this?

Yep, but the issue was when a user was creating a new site item, the revision_limit passed into `new TSC.Site(... revision_limit=100)` was not being used at all. The setter on line 188 was letting them set it to an existing site item, but not at creation time.

t8y8:
I see, we were clearing it to `None`

t8y8 (approved):

Issue #849, opened 2021-05-23
Pass timeout value for POST request


Udit107710:
**Feature Request**
While querying on the Metadata API I want to pass the timeout value. I am unable to do so because the code structure restricts it.

PR #850, opened 2021-05-23
feat: accept parameters for post endpoint

Udit107710:
This will allow request parameters like timeout
to be set by user. Because the metadata api use
post_request endpoint, that endpoint will also
support adding request parameters.
This feature was already present in the GET
endpoint

Related issue: #849


bcantoni:
Hi @Udit107710 thanks for the pull request!

While we wait for a review we'll also need to ask you to submit a Contributor License Agreement - details are listed here: http://tableau.github.io/contributing.html

t8y8:
Generally speaking, this is probably a good idea!

However, curious the use case -- the GraphQL API itself has a server-side timeout of 60 seconds by default, is there a reason you want to set that on the client side to a lower value?

Udit107710:
@t8y8 I did not know about the 60 sec timeout. But even then, I think it will be good to have common functionality across the endpoints

bcantoni:
Thanks Udit for the CLA; updated tags here to reflect.

t8y8:
@Udit107710 fair enough -- but yeah the metadata api times out 60 seconds by default, and returns an error on the `error` payload.

Still, change is reasonable -- can you target it against `development` instead of `master`? I'm also not sure if the formatting will match the formatter or not; you can run `black` against the file to see.

Udit107710:
@jacalata I think you have incorrectly merged development into my branch

Udit107710:
@t8y8 Ran the check by pycodestyle and that passed. Didn't know about the check by black. Checking and updating branch

t8y8:
This PR doesn't have the "auto mergeable" checkmark which makes me somewhat nervous -- I think because it was branched from master, it includes the issue template (which we don't need in dev). Can you rebase on development rather than retarget the PR? That should fix it... I think.

Otherwise looks good.

bcantoni:
Thanks @t8y8 good catch - @Udit107710 I think if you rebase this PR against the current `development` branch here, that should fix it.

bcantoni:
@Udit107710 reminder ping on this one - could you please rebase your branch from the current `tableau:development` branch? Then we should be able to merge.

jacalata:
I fixed the conflict, it was an update in content type.

Udit107710:
> 
> 
> I fixed the conflict, it was an update in content type.

Thanks @jacalata

PR #859, opened 2021-07-01
Be more clear that always using the latest REST API version is recommended

bcantoni:


t8y8:
We could make `use_server_version=True` the default on a new client instance, vs today where it's false by default

bcantoni:
@t8y8 that's an interesting idea - but might that cause backwards-compability issues for people?

t8y8:
It shouldn't -- if they're upgrading their server already, they could still do so. They can also still manually set it after the fact and it'll override ServerInfo's call.

I don't think there are any calls that stop working outright between 2.3 and 3.X are there?

The version detection logic works all the way back to 9.0, if it's newer than API version 2.4 it will use `ServerInfo` but otherwise there's a legacy way to detect version.

The only reason we didn't do it originally is because we didn't want to "issue silent HTTP calls" but I think that's a very small price to pay to get people using the latest version.

They can also explicitly set it to `False` if for some reason they don't want it to fire off.

shinchris (approved):

bcantoni:
Decided to include just the original text change here as an incremental improvement, moving the broader idea of changing the default to a new issue #876 so we can consider it separately.

Issue #873, opened 2021-08-23
List workbooks error when Personal Space workbooks exist (missing project_id)


pschuemann:
**Describe the bug**
Listing all existing workbooks is failing with an error.

We usually run the following code to find dashboards by name for refresh operations. Since yesterday its failing with a ValueError exception. After some investigation it looks like a workbook in a "Personal Space" is causing it as it doesn't have a project_id nor a project_name. It can be circumvented by patching the code with:
```
if not project_id:
  continue
```
to workbook_item.py line 278 and additionally ignoring the server error
```
tableauserverclient.server.endpoint.exceptions.ServerResponseError:
	400006: Bad Request
		Invalid page number '7'
```
following it.

**Versions**
Details of your environment, including:
 - Tableau Online
 - Python v3.9.4
 - tableauserverclient-0.16.0

**To Reproduce**
Here is the code. It worked fine before the user added the personal workbook:
```
import tableauserverclient
tableau_auth = tableauserverclient.TableauAuth(...)
server = tableauserverclient.Server(server_address='https://dub01.online.tableau.com/', use_server_version=True)
with server.auth.sign_in(tableau_auth)
  refresh_objs = [o for o in tableauserverclient.Pager(server.workbooks)]
```

**Results**
> ValueError: project_id must be defined.

bcantoni:
Thanks @pschuemann for reporting this. I too can reproduce it on Tableau Online. We'll do some more digging here to understand what changed.

bcantoni:
I confirmed with Query Workbooks for Site REST calls directly that workbooks in the new personal space feature do not have `project_id` or `project_name` defined. Looks like we'll need to properly handle that in the TSC library and see if any other objects are affected in a similar way.

adeturckheim:
Hi there, facing this problem too.
No issue with Projects, Users and Views

TomKlotzPro:
Hello @bcantoni, thanks for the fix!
When do you plan to release this fix?

jacalata:
This was released in 0.17 in October https://github.com/tableau/server-client-python/releases/tag/v0.17.0

PR #875, opened 2021-08-23
Add handling for workbooks in personal spaces which will not have pro…

bcantoni:
WIP testing a fix for #873


tableauserverclient/models/workbook_item.py
```diff
@@ -275,6 +276,10 @@ def from_response(cls, resp, ns):
                 data_acceleration_config,
             ) = cls._parse_element(workbook_xml, ns)
 
+            # workaround for Personal Space workbooks which won't have a project
+            if not project_id:
```
jacalata, line 280: So this just inserts a meaningless project id to avoid code crashing on the empty value, is that right? Is it possible to put in a human-readable string that makes this more obvious, instead of a uuid?

t8y8:
Does the REST API return anything for the container type, I know in the metadata API we say "containerType" and "containerName" -- it would be nicer if we just modeled that rather than had fake strings.

We could also just make project_id not required on the model and allow it to be none, then handle that in the Serializer.

Fake rando strings feels like my 3rd choice.

(Side note, I seem to have lost my commit bit, was that intentional?)

tableauserverclient/models/workbook_item.py
bcantoni: Yes that's right. And upon further thought this is not a great way to solve it, especially because it will populate project_id with something that _looks like_ a project ID but really isn't.

bcantoni:
@t8y8 I don't see anything in the REST response like that. Instead maybe it's simpler to change the class instantiation from using project_id to just the workbook id (which should always be there and should be unique).

Also I'm not sure where you were dropped, it may have been from team membership changes. In any case I've added you back here with permissions.

t8y8:
> @t8y8 I don't see anything in the REST response like that. Instead maybe it's simpler to change the class instantiation from using project_id to just the workbook id (which should always be there and should be unique).

Yeah, that seems reasonable, and it cleans up `WorkbookItem` a bit. The lowest impact change it to make `project_id` be `project_id=None` in the signature, and then handle it later down in WorkbookRequest serializers, if it's none don't add a Project element. A larger refactor is to get rid of project_id all together, or move it as a keyword-only argument, and the new way to set a project would just be `workbook.projectI_id = ProjectItem or id string`. I think both are fine approaches, one is riskier.

On the side, it's probably worth pinging whatever team owns that now (Personal Spaces, I guess) because it's a miss, imo, that the REST API doesn't include the information. We could update WorkbookItem to handle personal spaces at a later time.

matthieucan:
Hello,
I believe this is a fix for a bug that is hitting our prod since ~1 week. We are unable to run any of our Tableau-related jobs, which all try to retrieve lists of workbooks.
Would it be possible to provide us with an ETA for a resolution and a bugfix release?
Many thanks in advance!

ChristophGmeiner:
> Hello,
> I believe this is a fix for a bug that is hitting our prod since ~1 week. We are unable to run any of our Tableau-related jobs, which all try to retrieve lists of workbooks.
> Would it be possible to provide us with an ETA for a resolution and a bugfix release?
> Many thanks in advance!

Same here!

jacalata:
The fake uuid is not ideal but will definitely get people unblocked here, so worth going ahead with.

matthieucan:
> The fake uuid is not ideal but will definitely get people unblocked here, so worth going ahead with.

Sounds good, when can we expect it?

adeturckheim:
Hi there, anything blocking this merge request? That's a blocker for us

jacalata (approved):

matthieucan:
Hi @jacalata,
Thank you for merging :pray: Could you give an estimate on when the next release including this fix is likely to happen?

alexandruandrei90:
Hi, thank you for merging the change - do you know when this will be released?

jacalata:
It should be this week.

alexandruandrei90:
Hi, are there any updates on the release?

PR #878, opened 2021-09-01
Update publish_datasource.py

smarvin:
Edited to reflect a changed filename in the REST API docs.


bcantoni (approved):

bcantoni:
@smarvin thanks for catching and fixing this. We referenced that link in a couple places in the API docs as well. Could you peek at #879 and give it a +1 to confirm?

bcantoni:
Merged.

I'm not sure why the Slack check step showed as failed. This PR did come through to our Slack just fine.

PR #880, opened 2021-09-01
Upgrade to newer Slack action provider

bcantoni:
Testing to see if upgrading this will fix the error seen in #878


bcantoni:
@jacalata please review when you get a chance. It's possible the Checks in #878 were just a glitch, but in any case it's probably a good idea to be on the latest version here.

jacalata (approved):

PR #881, opened 2021-09-04
Jorwoods/query endpoint paging

jorwoods:
Tweaks the `QuerySet` added by @scuml to enable automatic pagination through all results. @scuml - Was the original intent to have users explicitly paginate through results? Would handling the pagination implicitly within the `__iter__` break anything?

Also adds the ability to use `QuerySet` operations on Groups and Flows.


scuml:
> Was the original intent to have users explicitly paginate through results?

Yes - this looks good.  Thanks for the improvement.

jorwoods:
@bcantoni Seems the slack notifier is failing when I push on this PR. The error message is empty so I am not sure what is needed to resolve it.

jorwoods:
This to me is a pretty important PR that should definitely be merged before the next release. Any use of the `QuerySet` methods without this fix will only yield the first 100 items, regardless of how many exist.

jacalata:
I think I'm missing something here - what does this add to the existing Pager/Pagination functionality? In the sample list.py, I can retrieve over 100 views already.

jorwoods:
Mostly it’s syntactic sugar. This corrects an issue with the QuerySet introduced in #615 .

It makes filtering easier than building a`RequestOptions` object to pass in. Searching for a given user, for example, becomes:

`server.users.filter(name=jorwoods’)`

jorwoods:
@jacalata I should clarify a little more. The `QuerysetEndpoint` that was introduced by #615 would only handle the first page. This PR fixes the QuerysetEndpoint so that it can work on more than the first page.

tableauserverclient/server/query.py
```diff
@@ -15,11 +16,54 @@ def __init__(self, model):
...
+            self._fetch_all()
+            yield from self._result_cache
 
     def __getitem__(self, k):
-        return list(self)[k]
+        page = self.page_number
+        size = self.page_size
+
+        page_range = range((page - 1) * size, page * size)
+
```
jacalata, line 35: could you add some comments in here, or maybe rename the index variables?  it's a bit of an unexpected wall of text, I think some framing would make it easier to understand.

jacalata: Not gonna hold up the whole thing for that though, it can be a separate item.

Issue #860, opened 2021-07-08
Support for last refresh date of a flow


atinbansal148:
Is there an end point to get the last refresh date time for a prep flow?
Trying to build code for automatic triggering flow refresh if data source data is updated after last refresh time for flow

jorwoods:
Support for pulling information about flow runs in #884

jacalata:
This change was in https://github.com/tableau/server-client-python/releases/tag/v0.17.0

PR #884, opened 2021-09-11
Add FlowRun Item and Endpoints.

jorwoods:
Adds support to the package for getting flow run status, as well as the ability to cancel flow runs.


jorwoods:
The API documentation incorrectly has the response XML including an attribute of `flow id` instead of what it really returns, `flowId`.

https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_flow.htm#get_flow_run

jorwoods:
Adds required endpoints and items for #860

tableauserverclient/server/endpoint/flowruns_endpoint.py
```diff
@@ -0,0 +1,49 @@
+from tableauserverclient.models import flowrun_item
+from .endpoint import Endpoint, QuerysetEndpoint, api
+from .. import FlowRunItem, PaginationItem
+
+import logging
+
+logger = logging.getLogger("tableau.endpoint.flowruns")
+
+
+class FlowRuns(QuerysetEndpoint):
```
t8y8, line 10: I know we've already merged some endpoints in as Queryset endpoints, but I've been skeptical of the approach -- I worry about needing more and more combinations of endpoint types, rather than just relying on helper methods or wrappers (like `Pager`).

The ship has sailed, but we should probably decide what route we want to go and convert all the endpoints at once

tableauserverclient/server/server.py
```diff
@@ -83,6 +84,7 @@ def __init__(self, server_address, use_server_version=False):
         self.data_acceleration_report = DataAccelerationReport(self)
         self.data_alerts = DataAlerts(self)
         self._namespace = Namespace()
+        self.flowruns = FlowRuns(self)
```
t8y8, line 87: I think it should be `flow_runs` to match others like `data_alerts`

tableauserverclient/models/flowrun_item.py
```diff
@@ -0,0 +1,106 @@
...
+            self._progress = progress
+        if background_job_id is not None:
+            self._background_job_id = background_job_id
+
+
+    @classmethod
+    def from_response(cls, resp, ns):
+        all_flowrun_items = list()
+        parsed_response = ET.fromstring(resp)
+        all_flowrun_xml = itertools.chain(
```
t8y8, line 82: I'm not super familiar with flow runs -- is this because we have two versions of response or something?

tableauserverclient/models/job_item.py
```diff
@@ -24,6 +26,7 @@ def __init__(
         self._finish_code = finish_code
         self._notes = notes or []
         self._mode = mode
+        self._flowrun = flowrun
```
t8y8, line 29: `_flow_run`

tableauserverclient/models/job_item.py
```diff
@@ -66,6 +69,14 @@ def mode(self, value):
         # check for valid data here
         self._mode = value
 
+    @property
+    def flowrun(self):
```
t8y8, line 73: `flow_run` and same below

t8y8 (changes requested):
Minor changes but looks good

tableauserverclient/server/endpoint/flowruns_endpoint.py
jorwoods: The `Queryset` methods I think are a nice convenience functionality for building up `RequestOptions` in a way that is a little more intuitive and fluent. I set it as the superclass for `FlowRuns` since I was working under the presumption that usually an individual querying the site will care more about getting runs for a specific flow and in a specific order than just generally querying all flows.

tableauserverclient/models/flowrun_item.py
jorwoods: Yes, the `Get Flow Run` and `Get Flow Runs` endpoints seem to return different XML. (The documentation is also a little wrong about the XML document returned. It returns `flowId` and not `flow id`.

https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_flow.htm#get_flow_run

tableauserverclient/models/job_item.py
jorwoods: Changed to flow_run.

tableauserverclient/models/job_item.py
jorwoods: Changed to flow_run.

tableauserverclient/server/server.py
jorwoods: Changed to `flow_runs`.

tableauserverclient/models/flow_run_item.py
```diff
@@ -0,0 +1,106 @@
+import xml.etree.ElementTree as ET
+from ..datetime_helpers import parse_datetime
+import itertools
+
+
+class FlowRunItem(object):
```
jacalata, line 6: is this a type of JobItem? It looks very similar, might be nice not to have several almost-identical job items.

vogelsgesang: It is indeed pretty similar (`started_at`, ...) but also has some unique properties (`flow_id`, ...)
Not sure what to do about this...
Maybe we could factor out those common fields into a base class - but then again: is it worth it?

@jacalata do you have some particular solution in mind here?

tableauserverclient/models/flowrun_item.py
vogelsgesang: @dzucker-tab FYI re
> The documentation is also a little wrong about the XML document returned. It returns flowId and not flow id.

vogelsgesang:
@jorwoods could you please rebase this pull request on latest `development`?
There are some (hopefully trivial) conflicts. Besides that, I think the `wait_for_job` method should be refactored to also use the `ExponentialBackoffTimer` class introduced by #903

tableauserverclient/server/endpoint/flow_runs_endpoint.py
```diff
@@ -0,0 +1,76 @@
...
+            flow_run = self.get_by_id(flow_run_id)
+            logger.debug(f"\tFlowRun {flow_run_id} progress={flow_run.progress}")
+
+        logger.info("FlowRun {} Completed: Status: {}".format(flow_run_id, flow_run.status))
+
+        if flow_run.status == "Success":
+            return flow_run
+        elif flow_run.status == "Failed":
+            raise FlowRunFailedException(flow_run)
+        elif flow_run.status in ["Canceled", "Cancelled"]:
```
t8y8, line 73: Does Server really return it spelled both ways?!

PR #885, opened 2021-09-15
Update CI pipeline to test on Mac and Win, add early support for Python 3.10

bcantoni:


bcantoni:
This is an experiment to see how well our tests and everything run on Mac & Win systems. Hopefully everything "just works" :)

bcantoni:
Okay this all works and is ready for review. I'll squash the commits into one when merging.

t8y8:
What's the turnaround time for a CI run? I originally excluded OS X and Windows since this library is only ever pure-python and not making any os-specific calls. If it's nice and quick though, no reason not to test them all

bcantoni:
@t8y8 yep good question. If the last run is any indication, the whole matrix was done in under a few minutes. Even though the matrix is up to 3 OS x 6 Python, they all run in parallel and are individually pretty quick (~1 min). So for now it seems fast enough. The risk might be if/when jobs are waiting in a queue for resources, the overall time might stretch out.

t8y8 (approved):
🚀

PR #887, opened 2021-09-17
Update master from recent dev branch changes

bcantoni:
There shouldn't be any functional TSC code changes here, just catching up master with GitHub actions and other recent minor changes.


bcantoni:
@t8y8 for your review: this catches up master with development to pick up, among other things, the github actions.

bcantoni:
This PR is picking up a bunch of actual code changes now in development, so we should not merge this until we do the next release. (Which probably should happen soon-ish.)

PR #895, opened 2021-09-18
Add example for querying metadata API

vogelsgesang:


samples/metadata_query.py
```diff
@@ -0,0 +1,65 @@
...
+    parser.add_argument('--server', '-s', required=True, help='server address')
+    parser.add_argument('--site', '-S', help='site name')
+    parser.add_argument('--token-name', '-n', required=True,
+                        help='name of the personal access token used to sign into the server')
+    parser.add_argument('--token-value', '-v', required=True,
+                        help='value of the personal access token used to sign into the server')
+    parser.add_argument('--logging-level', '-l', choices=['debug', 'info', 'error'], default='error',
+                        help='desired logging level (set to error by default)')
+    # Options specific to this sample
+    parser.add_argument('datasource_name')
```
t8y8, line 26: We should probably make this resilient to a `Null`, if you make the default value `None` does the query method insert a null for the variable?

(I forgot what I wrote :) )

t8y8 (changes requested):
Minor request and then looks good.

samples/metadata_query.py
vogelsgesang: `datasource_name` is a required parameter. It does not have a default value. Invoking this script without a `datasource_name` prints

```
usage: metadata_query.py [-h] --server SERVER [--site SITE] --token-name TOKEN_NAME --token-value TOKEN_VALUE [--logging-level {debug,info,error}] datasource_name
metadata_query.py: error: the following arguments are required: datasource_name
```

t8y8: Sorry, I meant I think that it should just print all data sources if you don't supply a name. You don't have to if you really want to keep it to just 1 published data source, but I think it shows off GraphQL's flexibility to allow both.

vogelsgesang: Now I got it...
Yes, querying all data sources works, too. Updated the sample to do so

vogelsgesang:
@t8y8 I changed the sample so it is also possible to query all data sources and rebased the PR on the `development` branch. Anything else or are we good to merge this?

t8y8 (approved):
🚀 

I think the sample looks good -- we still need to fix the pipeline though. Let me see if I can do that from the outside world

t8y8:
I can't see why the python tests don't run -- they worked when @bcantoni added python 3.10 and haven't since, but I remember seeing them run on that PR itself (across linux, mac, and windows no less).

I will file an issue, this is blocking a ton of open PRs (maybe just revert the slack action?)

bcantoni:
@vogelsgesang this looks ready for merge now.

PR #900, opened 2021-09-23
Revert slack.yml to unblock open PRs

t8y8:
I don't have the time to figure out a fix for the moment, this might be the fastest way


t8y8:
And the tests are back!

bcantoni (approved):
Agreed. There must be some interaction here which is affecting Slack for PRs. Curiously, we are still getting Slack messages from new Issues, so the webhook URL must still be okay.

I'll merge this and we can dive in later.

PR #804, opened 2021-02-22
Type hinting on Datasources and Workbooks

jorwoods:
This PR enables the mypy type checking, as well as adds type hints to datasource and workbook endpoints and items. Through testing, I noticed deprecation errors on python 3.5, and that new releases of pip no longer support it, so that was removed as well.

PR #903, opened 2021-09-24
Add `jobs.wait_for_job` method

vogelsgesang:
Based on @FrancoisZim 's [suggestion](https://github.com/tableau/server-client-python/pull/891#discussion_r713791435).
Currently still a draft to get feedback on the proposed interface. Will add test cases, as soon as we got agreement on the interface & functionality.

This commit adds a `wait_for_job` method which will repeatedly poll
a job's status until that job is finished.

Internally, it uses an exponential backoff for the polling intervals.
That way, it is snappy for fast-running jobs without putting too
much load on the server for long-running jobs.

It returns the successfully finished `JobItem` object which might be
of interest to the caller, e.g. to inspect the reported `started_at`
`finished_at` times or the `notes`.

For failed jobs, `wait_for_job` raises an exception. That way, we
ensure that errors in jobs don't accidentally go unnoticed. The
`jobs` object can still be retrieved from the exception object, if
required.


t8y8:
It makes sense to me.

The back of my head is asking if the `Jobs` endpoint is the right place for logic like that -- I think this would be the first 'utility' method we have on an endpoint... Is there a different way to present it like a `TSC.Waiter` that can handle other wait-retry-backoff methods? 

Maybe that's too much, it's a bit off-the-cuff. Let me know what you think, otherwise I'll give this code as is a deeper review

FrancoisZim:
Looks good to me.  I don't really agree with @t8y8 about a generalizable TSC.Waiter class - I think all async operations spawn a background job and so this is the right place to implement this wait logic.

tableauserverclient/server/endpoint/jobs_endpoint.py
```diff
@@ -48,3 +54,35 @@ def get_by_id(self, job_id):
         server_response = self.get_request(url)
         new_job = JobItem.from_response(server_response.content, self.parent_srv.namespace)[0]
         return new_job
+
+    def wait_for_job(self, job_id, *, timeout=None):
+        id_ = getattr(job_id, "id", job_id)
```
t8y8, line 59: If the caller is always passing in the string job_id, this check will always fail and return the job_id.

If you want to allow callers to pass in a JobItem OR a job_id, this is a clever way to do it but it's a little non-obvious at first read. You could call the parameter just `job` or `job_or_job_id` to help.

vogelsgesang:
> I think this would be the first 'utility' method we have on an endpoint...

`server.filesuploads.upload` is also kind of an utility method. It chunks the file into multiple parts and uploads each chunk through a separate HTTP request. Similarly most `publish` methods (workbooks, datasources, ...) encapsulate multiple API calls into one utility method (first the upload, then the actual publish)

But I see what you mean here... To me the main question would be:

> Is there a different way to present it like a TSC.Waiter that can handle other wait-retry-backoff methods?

which makes me wonder: which other things would there in the Tableau Server API which we might want to wait on?
If there are other things we want to wait on than jobs: yes, maybe a `TSC.Waiter` makes sense...

But afaict the "jobs" concept is already Tableau's abstraction for everything async on the server, and hence introducing another layer of abstraction on the client would add only limited value

tableauserverclient/server/endpoint/jobs_endpoint.py
```diff
@@ -48,3 +54,35 @@ def get_by_id(self, job_id):
...
+                max_sleep_time = max(ASYNC_JOB_POLL_MIN_INTERVAL, timeout - elapsed)
+
+            time.sleep(min(current_sleep_interval, max_sleep_time))
+            job = self.get_by_id(id_)
+            current_sleep_interval *= ASYNC_JOB_POLL_BACKOFF_FACTOR
+            logger.debug(f"\tJob {id_} progress={job.progress}")
+
+        logger.info("Job {} Completed: Finish Code: {} - Notes:{}".format(id_, job.finish_code, job.notes))
+
+        if job.finish_code == 0:
```
t8y8, line 81: Either a quick comment or change the magic numbers to a constant so we remember for the future what the finish codes mean

t8y8 (changes requested):
Minor changes.

I also want to verify this works for AD sync, and I think it does (https://github.com/tableau/server-client-python/blob/fefd6f18d8a6617829c6323879d2c3ed77a4cda6/tableauserverclient/server/endpoint/groups_endpoint.py#L101) -- but confirm that should be poll-able as well

tableauserverclient/server/endpoint/jobs_endpoint.py
vogelsgesang: I simply copied this trick from the `jobs.cancel` endpoint and saw it in many other endpoints in TSC.
I thought, it would be considered best practice...

Given this is consistent with the remaining code of TSC, do think I should change it here?

t8y8: It looks like all instances of that trick are related to jobs/extract refresh. I don't think I wrote any of those endpoints, which might be why I'm unfamiliar with the pattern. I'm not a huge fan but I guess it's already in there.

jacalata:
The other thing we can wait for is a Flow Run, which might actually be what the BackgroundJobItem is related to now I look some more. @jorwoods how does this look to you in the context of adding the flows endpoints?

jorwoods:
Regarding any changes to #884, this would not be too difficult to implement on the `flow_runs` endpoint as well.

tableauserverclient/server/endpoint/jobs_endpoint.py
jorwoods: @t8y8 I wrote that getattr trick on the refresh endpoints. The reason for it was because the datasource endpoint was expecint a `DatasourceItem` whereas the workbook endpoint was expecting a string of `WorkbookItem.id`. That trick allowed for a consistent, but backwards compatible endpoint API, where they both accepted both. (One of the reasons I ultimately started #804 )

vogelsgesang:
argh... Ok, so flows are also de-facto jobs, but for some reason didn't use the jobs API. That's sad 😢 Such a nice abstraction on the server side, but then its inconistent by one job type not using the common REST interface... @jacalata maybe the "jobs" endpoint/abstraction should be added as a guideline to the TAG design guidelines for our REST API?

Either way, it is what it is... I will rewrite this code to pull out the "retry with exponential backoff" logic, such that it can be reused across multiple places, in particular also by @jorwoods in the flow endpoint, without having to duplicate that logic.
Will only have time for this a few days from now, though

tableauserverclient/server/endpoint/jobs_endpoint.py
vogelsgesang: So what is our strategy going forward? Do we only want to accept strings? Or only `JobItem`? Or both?
If both: Do we want to use `getattr` or do we want to use a more specific typing, e.g.

```
if isinstance(job_id, JobItem):
    job_id = job_id.id
assert isinstance(job_id, str)
```

jorwoods: If both, explicit `isinstance` checking would be better as it will play nicer with mypy if and when type hinting is implemented.

tableauserverclient/server/endpoint/jobs_endpoint.py
vogelsgesang: Added `JobItem.FinishCode` which has all finish codes as named constants

tableauserverclient/server/endpoint/jobs_endpoint.py
vogelsgesang: ok, so it's `isinstance` now

vogelsgesang:
added test cases and refactored the waiting logic into a reusable `ExponentialBackoffTimer` which can also be reused for flows.

I think this should be ready for the final review now

t8y8 (approved):
Looks great -- thanks for doing the diligence and discovering flows don't fit the pattern. (I'm not too surprised they don't, there was significant time between BG Jobs and Flows Runs).

The testing of the BackoffTimer is excellent -- I'm excited that a bonus of this implementation is that anyone can use the timer to write their own backoff between retries of non job calls, like waiting for SOLR to update or something.

🚀

PR #911, opened 2021-10-07
Stop supporting Python 3.5

vogelsgesang:
Python 3.5 is already end-of-life and no longer receives security patches for over a year now.
Also, I recently added an f-string because all Python versions which are still in support, also
support f-strings. I was unpleasantly surprised when I had to realize that we still claim to
support Python 3.5 which doesn't have f-strings...


t8y8 (approved):
🚀

jacalata:
thanks!

Issue #899, opened 2021-09-21
Can't assign description property to workbook item


dev-mkc19:
**Describe the bug**
Can't assign description property to WorkbookItem as I can with others properties. I noticed that tests skips this property for unknown reasons.

**Versions**
Details of environment, including:
 - Tableau Server version 2021.2.1 (20212.21.0712.0907) 64-bit Linux
 - Python version 3.6.8
 - TSC library version 0.16.0

**To Reproduce**
<img width="662" alt="error_screenshot" src="https://user-images.githubusercontent.com/10463072/134188421-a2ca64dd-ba77-4f6f-bf7f-10e23018b0b1.png">
print(type(wbs[0]))
wbs[0].description = 'new description'

**Results**
AttributeError: can't set attribute

jacalata:
This wasn't included in that other CL at all, oops.

jacalata:
The workbook description needs to be enabled server-side, for some reason.

PR #912, opened 2021-10-08
Adds description to datasource item

jacalata:
Shame to leave this work just sitting there


jacalata:
This is addressing https://github.com/tableau/server-client-python/issues/899

Issue #876, opened 2021-08-24
Consider switching use_server_version default to True


bcantoni:
This is a spin-out of a conversation from #859. Capturing it here for future consideration.

@t8y8 suggestion was to could make `use_server_version=True` the default on a new client instance, vs today where it's false by default:

https://github.com/tableau/server-client-python/blob/5cbfc3452453bedac2ef3039fa11b0d59fdacb86/tableauserverclient/server/server.py#L32

I had asked what impact would that have on existing clients, i.e. would this be a surprising change perhaps with compatibility issues.

Tyler's feedback:

It shouldn't -- if they're upgrading their server already, they could still do so. They can also still manually set it after the fact and it'll override ServerInfo's call.

I don't think there are any calls that stop working outright between 2.3 and 3.X are there?

The version detection logic works all the way back to 9.0, if it's newer than API version 2.4 it will use ServerInfo but otherwise there's a legacy way to detect version.

The only reason we didn't do it originally is because we didn't want to "issue silent HTTP calls" but I think that's a very small price to pay to get people using the latest version.

They can also explicitly set it to False if for some reason they don't want it to fire off.

rejuls:
It would be useful to set the default as true, would have saved me a lot of time trying to figure this out - https://github.com/tableau/server-client-python/issues/853#issuecomment-1008771146

jacalata:
While doing this, we should fix the code so that determine_latest_version pays attention to any http options passed in, like --no-certcheck or a proxy.

PR #918, opened 2021-10-19
Jac/publish samples

jacalata:
Added an optional parameter allowing users to publish sample content into a project on creation or update, as described in the REST API docs https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_projects.htm#create_project


samples/create_project.py
```diff
@@ -45,23 +45,28 @@ def main():
...
+    server = TSC.Server(args.server)
+    server.add_http_options({'verify': False})
+    server.use_server_version()
     with server.auth.sign_in(tableau_auth):
         # Use highest Server REST API version available
         server.use_server_version()
 
         # Without parent_id specified, projects are created at the top level.
-        top_level_project = TSC.ProjectItem(name='Top Level Project')
+        top_level_project = TSC.ProjectItem(name='TPL2')
```
vogelsgesang, line 56: why rename? I actually preferred the previous names...
Here and other places

samples/create_project.py
```diff
@@ -45,23 +45,28 @@ def main():
     logging.basicConfig(level=logging_level)
 
     tableau_auth = TSC.PersonalAccessTokenAuth(args.token_name, args.token_value, site_id=args.site)
-    server = TSC.Server(args.server, use_server_version=True)
+    server = TSC.Server(args.server)
+    server.add_http_options({'verify': False})
+    server.use_server_version()
```
vogelsgesang, line 50: why disable verification in the sample?
I don't think we should such an anti-pattern/workaround into a sample which many customers might use as a starting point for their own scripts

samples/create_project.py
```diff
@@ -45,7 +45,8 @@ def main():
     logging.basicConfig(level=logging_level)
 
     tableau_auth = TSC.PersonalAccessTokenAuth(args.token_name, args.token_value, site_id=args.site)
-    server = TSC.Server(args.server, use_server_version=True)
+    server = TSC.Server(args.server)
+    server.use_server_version()
```
vogelsgesang, line 49: why change this?

bcantoni (changes requested):
Ah good point @vogelsgesang. This came up in #859 and we wrote a standalone issue to discuss whether we should change the default (see #876). I think we should revert this bit as part of this PR and revisit later.

bcantoni (approved):

samples/create_project.py
jacalata: imperfect undo. But it does make it easier to insert the ssl change for local testing :)

PR #931, opened 2021-10-31
Add support to datasource and workbook revisions

elsherif:
Add support to datasource and workbook revisions (get all item revisions / download item by revision number)

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0-rc.2]
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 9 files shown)


t8y8:
@elsherif sorry for the poor timing, but we just merged a big change to some of the same underlying files that add type hints to the function signatures -- would you be able to rebase on those changes, as well as consider adding type hints as well?

elsherif:
@t8y8 I rebased but I got some errors after adding the type hints, so I removed them now.

t8y8:
The problem is that the embedded fetcher makes it so `_thing` returns something like `Callable[[], List["WorkbookItem"]]) -> None` and then the property `thing` just returns the `List[WorkbookItem]`

elsherif:
@t8y8 can someone from Dev team add a commit for that so you can merge?

t8y8:
@jorwoods since you're on a type hinting spree -- if we merge this without would you want to go through and add them in? If that's the case I can give this a review this week.

jorwoods:
Yeah, I can add the type hints later.

jacalata (approved):
looks good to me

jacalata:
merging, type hints can come later

t8y8:
Just a note we also merged this without tests which isn't ideal

PR #934, opened 2021-11-01
Jorwoods/type hint data alert

jorwoods:
Add type hinting to data alert item and endpoint

### DIFF
--- a/tableauserverclient/models/data_alert_item.py
+++ b/tableauserverclient/models/data_alert_item.py
@@ -9,44 +9,50 @@ from .user_item import UserItem
 from .view_item import ViewItem
 
 
+from typing import List, Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from datetime import datetime
+
+
 class DataAlertItem(object):
     class Frequency:
-        Once = "Once"
-        Frequently = "Frequently"
-        Hourly = "Hourly"
-        Daily = "Daily"
-        Weekly = "Weekly"
+        Once: str = "Once"
+        Frequently: str = "Frequently"
+        Hourly: str = "Hourly"
+        Daily: str = "Daily"
+        Weekly: str = "Weekly"
 
     def __init__(self):
-        self._id = None
-        self._subject = None
-        self._creatorId = None
-        self._createdAt = None
-        self._updatedAt = None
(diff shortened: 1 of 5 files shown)


tableauserverclient/models/data_alert_item.py
```diff
@@ -9,44 +9,50 @@
...
+
+
 class DataAlertItem(object):
     class Frequency:
-        Once = "Once"
-        Frequently = "Frequently"
-        Hourly = "Hourly"
-        Daily = "Daily"
-        Weekly = "Weekly"
+        Once: str = "Once"
```
t8y8, line 20: I think these enum-like classes don't really need the type hint, it's pretty obvious it's a string and isn't going to change. Mypy is smart enough to infer from the literal.

tableauserverclient/server/request_factory.py
```diff
@@ -87,22 +92,26 @@ def update_req(self, column_item):
 
 
 class DataAlertRequest(object):
-    def add_user_to_alert(self, alert_item, user_id):
+    def add_user_to_alert(self, alert_item: "DataAlertItem", user_id: str) -> bytes:
```
t8y8, line 95: This should be a string, not bytes

tableauserverclient/server/request_factory.py
```diff
@@ -87,22 +92,26 @@ def update_req(self, column_item):
...
         xml_request = ET.Element("tsRequest")
         user_element = ET.SubElement(xml_request, "user")
         user_element.attrib["id"] = user_id
 
         return ET.tostring(xml_request)
 
-    def update_req(self, alert_item):
+    def update_req(self, alert_item: "DataAlertItem") -> bytes:
```
t8y8, line 102: string, not bytes

t8y8 (changes requested):
minor changes -- though as a general rule i think we can leave hints off the "enum-like" classes.

Now that we're on 3.5 and later, we could just convert them all to enums as well...

tableauserverclient/models/data_alert_item.py
jorwoods: Removed type hints from the frequency namespace

t8y8 (approved):

PR #938, opened 2021-11-01
Jorwoods/type hint groups

jorwoods:
Add type hinting to groups


jacalata:
I'm going to merge a few hint stories and then format them in one go.

PR #939, opened 2021-11-01
Type hint Job

jorwoods:


test/test_job.py
```diff
@@ -17,7 +17,7 @@
 
 
 class JobTests(unittest.TestCase):
-    def setUp(self):
+    def setUp(self) -> None:
```
vogelsgesang, line 20: do we really want to annotate all test cases? seems a bit redundant/noisy...

tableauserverclient/server/endpoint/jobs_endpoint.py
```diff
@@ -32,21 +38,20 @@ def get(self, job_id=None, req_options=None):
...
+    def get_by_id(self, job_id: str) -> JobItem:
         logger.info("Query for information about job " + job_id)
         url = "{0}/{1}".format(self.baseurl, job_id)
         server_response = self.get_request(url)
         new_job = JobItem.from_response(server_response.content, self.parent_srv.namespace)[0]
         return new_job
 
-    @api(version="2.6")
-    def wait_for_job(self, job_id, *, timeout=None):
+    def wait_for_job(self, job_id: str, *, timeout: Optional[T] = None) -> JobItem:
```
vogelsgesang, line 54: * `job_id: str` annotation is incorrect. The code below explicitly also allows `JobItem` instances
* `timeout: Optional[T]`: why `T`?  Can't this just be `Optional[float]`

tableauserverclient/server/endpoint/jobs_endpoint.py
```diff
@@ -32,21 +38,20 @@ def get(self, job_id=None, req_options=None):
         return jobs, pagination_item
 
     @api(version="3.1")
-    def cancel(self, job_id):
+    def cancel(self, job_id: str):
         id_ = getattr(job_id, "id", job_id)
```
vogelsgesang, line 42: thanks to the `getattr` trick, `cancel` currently accepts either a `str` or a `JobItem`...

Please,
* update the type hint accordingly
* replace the `getattr` trick by the explicit `isinstance` check similar to `wait_for_job`

tableauserverclient/server/endpoint/jobs_endpoint.py
```diff
@@ -8,14 +8,20 @@
...
 
 class Jobs(Endpoint):
     @property
     def baseurl(self):
         return "{0}/sites/{1}/jobs".format(self.parent_srv.baseurl, self.parent_srv.site_id)
 
     @api(version="2.6")
-    def get(self, job_id=None, req_options=None):
+    def get(
+        self, job_id=None, req_options: Optional[RequestOptionsBase] = None
```
vogelsgesang, line 23: type hints for `job_id` missing

vogelsgesang (changes requested):

test/test_job.py
jorwoods: Adding type annotations to all of the test cases like that enables mypy to inspect the body of the test. From [urllib3's notes on adding type hints](https://sethmlarson.dev/blog/2021-10-18/tests-arent-enough-case-study-after-adding-types-to-urllib3):

> Once we completed the addition of type hints to the source code of urllib3 we set our sights on the test suite. This may seem strange as our users are very unlikely to directly benefit from us adding type hints there, however there’s one big benefit: we’re able to find issues with types from a user perspective!

vogelsgesang: Nice! Didn't know about this, yet. Sounds good to me

vogelsgesang (approved):

PR #946, opened 2021-11-02
Fix slack once and for all

t8y8:
The red X keeps coming back so I'd like to mark this as allowably fail-able -- that way the innards of Slack/OAuth/Tableau credentials don't keep polluting test run reports :)

### DIFF
--- a/.github/workflows/slack.yml
+++ b/.github/workflows/slack.yml
@@ -4,6 +4,7 @@ on: [push, pull_request, issues]
 
 jobs:
   slack-notifications:
+    continue-on-error: true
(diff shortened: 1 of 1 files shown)


t8y8:
I signed the old Tableau CLA when I was an employee -- does that not transfer?

bcantoni (approved):

bcantoni:
@t8y8 this looks great - I was just creating another PR to remove it (#947), but continue on error will be great.

We'll also need to merge this (probably cherry pick?) into the master branch because that's where the actions are run from.

PR #947, opened 2021-11-02
Temporarily disable (remove) Slack workflow again because it sometime…

bcantoni:
…s fails


bcantoni:
#946 by Tyler will be a better solution.

vogelsgesang:
closing this PR, since I don't think we will ever want to merge it given we already have the other solution in place

PR #950, opened 2021-11-02
Jorwoods/type hint flow runs

jorwoods:
Add type hints to flow runs

### DIFF
--- a/.github/workflows/slack.yml
+++ b/.github/workflows/slack.yml
@@ -9,6 +9,7 @@ jobs:
     name: Sends a message to Slack when a push, a pull request or an issue is made
     steps:
       - name: Send message to Slack API
+        continue-on-error: true
         uses: archive/github-actions-slack@v2.2.2
         id: notify
         with:
--- a/.github/workflows/slack.yml
+++ b/.github/workflows/slack.yml
@@ -4,6 +4,7 @@ on: [push, pull_request, issues]
 
 jobs:
   slack-notifications:
+    continue-on-error: true
     runs-on: ubuntu-20.04
     name: Sends a message to Slack when a push, a pull request or an issue is made
     steps:
--- a/tableauserverclient/models/flow_run_item.py
+++ b/tableauserverclient/models/flow_run_item.py
@@ -2,50 +2,54 @@ import xml.etree.ElementTree as ET
 from ..datetime_helpers import parse_datetime
 import itertools
 
+from typing import Dict, List, Optional, Type, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from datetime import datetime
 
 class FlowRunItem(object):
     def __init__(self) -> None:
-        self._id=None
-        self._flow_id=None
-        self._status=None
-        self._started_at=None
-        self._completed_at=None
-        self._progress=None
-        self._background_job_id=None
+        self._id: str = ''
+        self._flow_id: Optional[str] = None
+        self._status: Optional[str] = None
+        self._started_at: Optional["datetime"] = None
+        self._completed_at: Optional["datetime"] = None
+        self._progress: Optional[str] = None
+        self._background_job_id: Optional[str] = None
 
     
     @property
-    def id(self):
+    def id(self) -> str:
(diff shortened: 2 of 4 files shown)


.github/workflows/slack.yml
```diff
@@ -4,10 +4,12 @@ on: [push, pull_request, issues]
 
 jobs:
   slack-notifications:
+    continue-on-error: true
     runs-on: ubuntu-20.04
     name: Sends a message to Slack when a push, a pull request or an issue is made
     steps:
       - name: Send message to Slack API
+        continue-on-error: true
```
jorwoods, line 12: @t8y8 @bcantoni The `continue-on-error` only appears to have an impact if its on the step, and not at the job level.

t8y8: Aww that's right, bad copy/paste job for me then.

Thanks!

bcantoni:
I haven't used continue on fully understand, but looking at the [docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) it seems it can be used either at the job or step level.

One thing definitely not clear for me: from which branch is the workflows definition file used for a PR like this? For most things it says the version in the default repo branch will be used (`master`), but we also have `development` and then a copy here in this PR. It's possible that just having it here is giving you the result you want.

Regardless, if it works here then that's great. Once merged to development I can do another cherry-pick to master so we all match.

tableauserverclient/models/flow_run_item.py
```diff
@@ -2,50 +2,54 @@
...
 class FlowRunItem(object):
     def __init__(self) -> None:
-        self._id=None
-        self._flow_id=None
-        self._status=None
-        self._started_at=None
-        self._completed_at=None
-        self._progress=None
-        self._background_job_id=None
+        self._id: str = ''
```
t8y8, line 12: Why change this from `None` to `str`? It might be ok but I'm a little nervous about the ripple impact in places where we might be checking this against is `None`

tableauserverclient/server/endpoint/flow_runs_endpoint.py
```diff
@@ -5,20 +5,26 @@
...
 
     @property
-    def baseurl(self):
+    def baseurl(self) -> str:
         return "{0}/sites/{1}/flows/runs".format(self.parent_srv.baseurl, self.parent_srv.site_id)
 
     # Get all flows
     @api(version="3.10")
-    def get(self, req_options=None):
+    def get(self, req_options: Optional["RequestOptions"]=None) -> Tuple[List[FlowRunItem], PaginationItem]:
```
t8y8, line 27: We use this format throughout our endpoints, I was thinking we might want to refactor this into a type with a typevar like `TSResponse[T] = Tuple[List[T], PaginationItem] -> TSResponse[FlowRunItem]` but I haven't dug into TypeVars deeply. No need to do it for this PR

t8y8 (approved):

tableauserverclient/models/flow_run_item.py
jorwoods: I'm on the fence on this one. A flow run is a bit different theoretically from some of the other item types. A user of the TSC library would not find themselves creating a `FlowRunItem` to then interact with the server, nor should the server produce `FlowRunItem`s without an id. Leaving it as None makes the most sense from an API uniformity standpoint, but making the id non-optional in a sense reinforces the idea that an end user wouldn't create a flow run by starting with the python object.

I can change it back and update the type hints accordingly if you disagree with the idea behind hard setting it to string.

tableauserverclient/server/endpoint/flow_runs_endpoint.py
t8y8: I will play around with that in a new PR, we can do it after all of these are in.

tableauserverclient/models/flow_run_item.py
t8y8: That makes sense, I think it's fine as you have it then

PR #956, opened 2021-11-10
Jorwoods/type hint revisions

jorwoods:
Adding type hints, tests, and other corrections to Revisions

### DIFF
--- a/tableauserverclient/models/revision_item.py
+++ b/tableauserverclient/models/revision_item.py
@@ -56,7 +56,7 @@ class RevisionItem(object):
 
     @classmethod
     def from_response(
-        cls: Type["RevisionItem"],
+        cls,
         resp: bytes,
         ns,
         resource_item
--- a/tableauserverclient/models/revision_item.py
+++ b/tableauserverclient/models/revision_item.py
@@ -1,6 +1,6 @@
 import xml.etree.ElementTree as ET
 from ..datetime_helpers import parse_datetime
-from typing import List, Mapping, Optional, TYPE_CHECKING, Type, Union
+from typing import List, Optional, TYPE_CHECKING, Type
 
 if TYPE_CHECKING:
     from datetime import datetime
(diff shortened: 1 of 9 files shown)


jorwoods:
@t8y8 @jacalata Type hints and tests added for revisions now. Ready for review.

tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -495,3 +502,12 @@ def download_revision(
             )
         )
         return os.path.abspath(download_path)
+
+    @api(version="2.0")
```
jacalata, line 506: This one should also be 2.3

jacalata (changes requested):
like magic, thank you :)

tableauserverclient/server/endpoint/workbooks_endpoint.py
jorwoods: Updated to 2.3.

tableauserverclient/models/revision_item.py
```diff
@@ -1,46 +1,66 @@
...
         return (
             "<RevisionItem# revisionNumber={_revision_number} "
-            "current={_current} deleted={_deleted}>".format(**self.__dict__)
+            "current={_current} deleted={_deleted} user={_user_id}>".format(**self.__dict__)
         )
 
     @classmethod
-    def from_response(cls, resp, ns, resource_item):
+    def from_response(
+        cls: Type["RevisionItem"],
```
t8y8, line 59: nit: I don't think we need to annotate the `cls` here, since it is passed in by the language itself and not the user

t8y8 (approved):
Minor comment

tableauserverclient/models/revision_item.py
jorwoods: Removed the type annotation.

PR #957, opened 2021-11-10
Bump default Server version

jorwoods:
As of 2021-11-11, Tableau Server versions below 2020.4.x are out of support. Bumping the default server version in accordance with that.

### DIFF
--- a/tableauserverclient/server/server.py
+++ b/tableauserverclient/server/server.py
@@ -59,7 +59,7 @@ class Server(object):
         self._session = requests.Session()
         self._http_options = dict()
 
-        self.version = "2.3"
+        self.version = "3.10"
         self.auth = Auth(self)
         self.views = Views(self)
         self.users = Users(self)
--- a/test/test_group.py
+++ b/test/test_group.py
@@ -124,8 +124,7 @@ class GroupTests(unittest.TestCase):
             add_user_response = f.read().decode('utf-8')
         with requests_mock.mock() as m:
             m.get(self.baseurl, text=get_xml_response)
-            m.post('http://test/api/2.3/sites/dad65087-b08b-4603-af4e-2887b8aafc67/groups/ef8b19c0-43b6-11e6-af50'
-                   '-63f5805dbe3c/users', text=add_user_response)
+            m.post('{0}/ef8b19c0-43b6-11e6-af50-63f5805dbe3c/users'.format(self.baseurl), text=add_user_response)
             all_groups, pagination_item = self.server.groups.get()
             single_group = all_groups[0]
             self.server.groups.add_user(single_group, '5de011f8-5aa9-4d5b-b991-f462c8dd6bb7')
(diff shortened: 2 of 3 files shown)


test/test_group.py
```diff
@@ -152,8 +151,7 @@ def test_remove_user_before_populating(self):
             response_xml = f.read().decode('utf-8')
         with requests_mock.mock() as m:
             m.get(self.baseurl, text=response_xml)
-            m.delete('http://test/api/2.3/sites/dad65087-b08b-4603-af4e-2887b8aafc67/groups/ef8b19c0-43b6-11e6-af50'
-                     '-63f5805dbe3c/users/5de011f8-5aa9-4d5b-b991-f462c8dd6bb7',
+            m.delete(self.baseurl + '/ef8b19c0-43b6-11e6-af50-63f5805dbe3c/users/5de011f8-5aa9-4d5b-b991-f462c8dd6bb7',
```
bcantoni, line 154: For consistency with the above, could this also use the "string".format() way here rather than concatenation?

bcantoni:
We should weigh the pros and cons here.

What benefit should this bring?

I can think of one possible con: customers are still using unsupported (old) Tableau versions. This could be a surprise to them if they upgrade to a new TSC release and their old scripts stop working.

tableauserverclient/server/server.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, server_address, use_server_version=False):
         self._session = requests.Session()
         self._http_options = dict()
 
-        self.version = "2.3"
```
jacalata, line 62: might make sense to move this down next to the use_server_version() check

jacalata (approved):
I think bumping the server version is a good idea. To make it definitely painless, what we should really be doing is checking the server version in the sign_in call at https://github.com/tableau/server-client-python/blob/b3ca20e6765c7cff2d5b095e880dc2b2a811d825/tableauserverclient/server/endpoint/auth_endpoint.py#L27

I can add some error handling in that before our next release.

t8y8:
The base was intentionally set low because the "server-info" endpoint didn't exist before "2.4", and trying to determine the server version would fail if it were a 2.3v server. Setting the default to the lowest lets us always upgrade.

I've argued to make the "upgrade to latest sever version automatically on sign in" the default forever, and would be happy to promote that... but even if you did that, you need the **default** version to be lowish. If you upgrade default in lockstep with the latest, an upgrade to TSC would break on sign-in against an older server, even if that server is a supported version.

jorwoods:
@t8y8 That makes sense. You kinda hope that no one is on a server that old, but it would definitely happen. I intentionally didn't update to latest, but instead only to versions still in support. That said, a less disruptive way would be to default the check server info to `True`.

It looks to me like the internals of fetching the server version will already gracefully handle not being able to find the `get_server_info` endpoint. I will close this PR. I see there is already an open issue on setting the default of using the latest server version. I will move the conversation there.

PR #968, opened 2021-12-09
Jorwoods/download excel files from views

jorwoods:
Built on top of #949 . Adds support for downloading excel crosstabs from views.


cdonahue-chartis:
Thanks for taking this issue on! I found myself needing to do the same recently, and was exploring whether to just directly use the REST API given the limitation within the Python client. Any guidance on when this might make its way into master? 

Really appreciate the work - this will be a big help to many.

jorwoods:
@cdonahue-chartis Strange. I see the functions in master, but when I install 0.18 in a venv, I also do not see the excel functions.

cdonahue-chartis:
@jorwoods Interesting. Glad I wasn't missing something. Please let me know if/when you have any updates on release timing. In the meantime, I'm able to use the dev branch and the Excel export works wonderfully. Thanks again.

PR #971, opened 2022-01-05
Update api-ref.md

julianste:
fix typo


bcantoni:
It was just a typo fix, so signing the CLA wasn't needed. For future larger changes it would be.

PR #978, opened 2022-01-26
Add section about handling SSL certificates for Tableau Server

bcantoni:
Also including some minor grammer edits


t8y8 (approved):
This has been an open question for many moons! I like the compromise, thanks for doing the legwork to show how to pass a self-signed cert to trust and at the end, with warning, mentioning you can verify False if you really really want to.

🚀

PR #982, opened 2022-01-28
Fix typings for clean run in mypy

jacalata:
probably introduced these failures in merge mistakes when I merged all those type-hints at once.


jorwoods:
Awesome. Once this gets merged, I'll rebase my other open PRs.

t8y8 (approved):

PR #989, opened 2022-02-13
Jorwoods/default to server version

jorwoods:
Default to using the API version of the specified Tableau Server.


jorwoods:
Could we just wrap that call in the init with a try/except, and log the error?

jacalata:
I think the nicest fix will be to let the user pass in the verify=False option to the constructor.  Probably cleaner to add that separately to this change, so I'll go ahead and check this in.

Issue #988, opened 2022-02-11
Use defusedxml library for prevention of xml attacks


jacalata:
See https://github.com/tableau/tabcmd/issues/15

https://pypi.org/project/defusedxml/

PR #1002, opened 2022-03-11
Use defusedxml library to reduce xml attack surface

jacalata:
https://github.com/tableau/server-client-python/issues/988

from https://pypi.org/project/defusedxml/, we need to make sure we call defusedxml.ElementTree instead of xml.etree.ElementTree for these methods: parse(), iterparse(), fromstring(), XMLParser
the only one of those we call is fromstring()

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ setup(
     test_suite='test',
     setup_requires=pytest_runner,
     install_requires=[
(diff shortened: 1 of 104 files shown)


ovinis (approved):
Thanks for updating the library. I think this also addresses Tabcmd issue that was logged for TSC https://github.com/tableau/tabcmd/issues/15

PR #1004, opened 2022-03-15
Add comments to clarify queryset iteration

jorwoods:
@jacalata Hope that is clearer.

### DIFF
--- a/tableauserverclient/server/query.py
+++ b/tableauserverclient/server/query.py
@@ -16,11 +16,15 @@ class QuerySet:
         self._pagination_item = None
 
     def __iter__(self):
+        # Not built to be re-entrant. Starts back at page 1, and empties
+        # the result cache.
         self.request_options.pagenumber = 1
         self._result_cache = None
         total = self.total_available
         size = self.page_size
         yield from self._result_cache
+
+        # Loop through the subsequent pages.
         for page in range(1, math.ceil(total / size)):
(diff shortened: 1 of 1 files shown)


tableauserverclient/server/query.py
```diff
@@ -31,12 +35,16 @@ def __getitem__(self, k):
         page = self.page_number
         size = self.page_size
 
+        # Create a range object for quick checking if k is in the cached result.
         page_range = range((page - 1) * size, page * size)
 
         if isinstance(k, slice):
```
jacalata, line 41: is slice ever set here?

jorwoods: `k` can either be an int or a `slice` object.

It could be a slice if it was called in the following manner:

```python
server.workbooks.all()[10:40:3]
```

jorwoods: In that example, if the client code calls

```python
server.workbooks.all()[10:40:3]
```

Then `__getitem__` receives `k=slice(start=10, stop=40, step=3)`. Any of the typical slicing syntax will work. For `server.workbooks.all()[:50]` then `k=slice(start=None, stop=50, step=None)`, then the `None`'s get replaced with sensible defaults so that it acts the way that slicing would on a list.

If start is `None`, it becomes `0`
If stop is `None`, it becomes the total number available
If step is `None`, it becomes 1

PR #1010, opened 2022-03-19
Add examples for simplified queryset to samples

jorwoods:


jacalata:
I need to come up with a better name than "Django style filtering", I've never used that so the docs were pretty hard to follow at first.

PR #1015, opened 2022-03-28
Jorwoods/quota tiers allow null

jorwoods:
Allow setting `site.user_quota` equal to None if license levels are set.

### DIFF
--- a/tableauserverclient/models/site_item.py
+++ b/tableauserverclient/models/site_item.py
@@ -1,3 +1,4 @@
+from multiprocessing.sharedctypes import Value
 import warnings
 import xml.etree.ElementTree as ET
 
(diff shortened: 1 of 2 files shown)


tableauserverclient/models/site_item.py
```diff
@@ -226,7 +227,9 @@ def user_quota(self) -> Optional[int]:
 
     @user_quota.setter
     def user_quota(self, value: Optional[int]) -> None:
-        if any((self.tier_creator_capacity, self.tier_explorer_capacity, self.tier_viewer_capacity)):
+        if value is not None and any(
```
jacalata, line 230: seems kind of odd that the only way to do this is to go through a step where both user quota and site capacity are None.  Maybe a helper method like replace_capacity_with_quota(user_quota=) and vice versa would make this cleaner to do?

jacalata (changes requested):
Here's a test that passes without your change: can you add it and update?
    def test_null_site_quota(self) -> None:
        test_site = TSC.SiteItem('testname', 'testcontenturl', tier_explorer_capacity=1, user_quota=None)
        assert test_site.tier_explorer_capacity == 1
        with self.assertRaises(ValueError):
            test_site.user_quota = 1
        test_site.tier_explorer_capacity = None
        test_site.user_quota = 1

jorwoods:
I also modified the existing `test_update` for sites to fail appropriately if the fix is not in place. Added a new test for the new override method. Added the test proposed by @jacalata to exhibit failures when setting user_quota directly to some non-null value, but work when `None`.

jacalata (approved):

jacalata:
thanks!

PR #1023, opened 2022-04-07
Add Status and StartedAt filters for jobs endpoint

jorwoods:
Adds Status and StartedAt filters for jobs endpoint. Removes deprecated argument to jobs get endpoint.

### DIFF
--- a/tableauserverclient/server/endpoint/jobs_endpoint.py
+++ b/tableauserverclient/server/endpoint/jobs_endpoint.py
@@ -17,9 +17,7 @@ class Jobs(QuerysetEndpoint):
         return "{0}/sites/{1}/jobs".format(self.parent_srv.baseurl, self.parent_srv.site_id)
 
     @api(version="2.6")
-    def get(
-        self, req_options: Optional[RequestOptionsBase] = None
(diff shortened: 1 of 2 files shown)


tableauserverclient/server/endpoint/jobs_endpoint.py
```diff
@@ -17,18 +17,7 @@ def baseurl(self):
         return "{0}/sites/{1}/jobs".format(self.parent_srv.baseurl, self.parent_srv.site_id)
 
     @api(version="2.6")
-    def get(
-        self, job_id: Optional[str] = None, req_options: Optional[RequestOptionsBase] = None
-    ) -> Tuple[List[BackgroundJobItem], PaginationItem]:
-        # Backwards Compatibility fix until we rev the major version
-        if job_id is not None and isinstance(job_id, str):
```
jacalata, line 24: why removing this now?

jorwoods: I saw that it was deprecated and removed it. Though, you are right it is outside of the scope for this PR. Reverting that change.

Issue #974, opened 2022-01-18
ValueError: project_id must be defined


hanishche:
Hi,

I am running into error:  ValueError: project_id must be defined.

Below is the code:

```
tsConn = TableauServerConnection('server_url','Token_name','token','site')
server = TSC.Server(tsConn.server_url, use_server_version=True)
tableau_auth = TSC.PersonalAccessTokenAuth(token_name=tsConn.token_name
                                           ,personal_access_token=tsConn.token_secret
                                           ,site_id=tsConn.site)
with server.auth.sign_in_with_personal_access_token(tableau_auth):
    print('[Logged in successfully to {}]'.format(tsConn.server_url))
    print('[Loading workbooks...]')        
    all_workbooks, pagination_item = server.workbooks.get()
    tableauWB = []
    for workbook in TSC.Pager(server.workbooks.get):
        tableauWB.append(
        (workbook.id
         ,workbook.name
         ,workbook.owner_id
         ,workbook.content_url
          ,'None' if len(workbook.project_id) ==0 else workbook.project_id
         ,workbook.project_name
         ,workbook.size
         ,'None' if len(workbook.tags) ==0 else workbook.tags
         ,pd.to_datetime(str(workbook.created_at)) 
         ,pd.to_datetime(str(workbook.updated_at)) 
         ,workbook.webpage_url       ))

print('[Tableau {} workbooks loaded]'.format(len(tableauWB)))

```

![image](https://user-images.githubusercontent.com/32518059/149996230-440e3b29-eccf-47e2-bc27-56bae861ca92.png)

bcantoni:
May be similar to #930

jpsued:
Hi together,
I'm getting this error after we upgraded from Tableau Server version 2020.2.4 to 2021.3.6.
All data sources / workbooks I'm getting the error for are located in the new introduced 'Personal Space' 
Maybe this helps to track down/handle the issue.

Planswalker:
Hi everyone,

I'm also running into this exact issue when trying to query all workbooks. I can also confirm that this is related to the "Personal Space" feature, which would make sense since if I'm understanding the wording in https://help.tableau.com/current/pro/desktop/en-us/personal_space.htm correctly, Workbooks created in a Personal Space aren't associated with a project by design. 

Is there a reason that the project_id field is required for creating a workbook? Or can we make that an optional parameter without issue?

bcmyguest1:
[...]

PR #1028, opened 2022-04-20
Workbook item nullable project

bcmyguest1:
Remove check for null project_id which is now possible as a result of the introduction of Personal Space in new tableau versions. Related issue found here: https://github.com/tableau/server-client-python/issues/974


jacalata:
Could you update the existing test for invalid_project_id?

bcmyguest1:
@jacalata yes will do

PR #1031, opened 2022-05-04
Add JWT support as an option for sign in

andybigbear:
Since we start to support sign in Via JWT. need to change TSC as well for adopt this feature.
this CL:
1. changed the TSC for JWT sign in support
2. update the sample/login as an example
3. create another example of invoking API using JWT

### DIFF
--- a/samples/login.py
+++ b/samples/login.py
@@ -5,11 +5,14 @@
 ####
 
 import argparse
+from contextlib import nullcontext
 import getpass
 import logging
+import enum
 
 import tableauserverclient as TSC
 
+# from tableauserverclient.models import json_web_token_auth
 
 # If a sample has additional arguments, then it should copy this code and insert them after the call to
 # sample_define_common_options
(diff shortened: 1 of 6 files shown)


samples/login.py
```diff
@@ -49,35 +54,69 @@ def sample_define_common_options(parser):
...
     # Make sure we use an updated version of the rest apis, and pass in our cert handling choice
-    server = TSC.Server(args.server, use_server_version=True, http_options={"verify": check_ssl_certificate})
+    server = TSC.Server(args.server, use_server_version=True)
     server.auth.sign_in(tableau_auth)
     print("Logged in successfully")
 
     return server
 
 
+def getSingleCredentialType(args):
```
jacalata, line 92: all this is unnecessary if we add jwt to the argument group

tableauserverclient/models/json_web_token_auth.py
```diff
@@ -0,0 +1,13 @@
+class JsonWebTokenAuth(object):
+    def __init__(self, json_web_token, site_id=None):
+        self.json_web_token = json_web_token
```
jacalata, line 3: why pass in a site id if it never gets used?

jacalata (commented):
I have an existing implementation of this that adds JWT as a sub-type of Credentials, which I think is a cleaner design that won't lead to a proliferation of sign_in_with_x methods.  How do you feel about using that?

andybigbear:
> 

Hi Jac, if you already have a design for Jwt sign in, I'm up for it as well. Could I review your design as well? If all is good, we can close this CL.

jacalata:
> > 
> 
> Hi Jac, if you already have a design for Jwt sign in, I'm up for it as well. Could I review your design as well? If all is good, we can close this CL.

It's based on this: https://github.com/tableau/server-client-python/pull/1032

Issue #1024, opened 2022-04-12
When we publish workbooks or datasources and use debug mode, user and password are displayed in plain text


DieKry:
[extract_from_log_publishing_using_debug_mode.txt](https://github.com/tableau/server-client-python/files/8472988/extract_from_log_publishing_using_debug_mode.txt)
Our environment:
Tableau ServerVersion: 2021.4.2 (20214.21.1217.2252) 64-bit LINUX
TableauServerClient: 0.17.0
Python: 3.8

The excerpts from the log show this fact. It is not relevant whether we publish workbooks with embedded datasources or with published datasouces. Can the output of user and password in the LOG be suppressed or encrypted when we use debug mode? We actually expect this to be the case.
In the excerpts user, password, database server and port were replaced. 
I added the excerpts from the log in the attached file.

@jacalata. Sorry for the delay, but I had a few days vacation. In the meantime I was able to install version 0.19.0 in a test environment and verify that no sensitive data is visible in plain text. Thanks a lot, works as expected.

jacalata:
@DieKry this fix is now available in the latest release, 0.19.0. Please let us know if you continue to see any sensitive values logged after updating!

PR #1032, opened 2022-05-04
create single Credentials class with redacted debug views

jacalata:
Creates a framework for future credential types to be added to
Addresses #1024 for the current types


tableauserverclient/models/tableau_auth.py
```diff
@@ -1,41 +1,70 @@
...
+
+    warnings.warn(
+        "TableauAuth(..., site=...) is deprecated, " "please use TableauAuth(..., site_id=...) instead.",
+        DeprecationWarning,
+    )
+
+
+# The traditional auth type: username/password
+class TableauAuth(Credentials):
     def __init__(self, username, password, site=None, site_id=None, user_id_to_impersonate=None):
```
t8y8, line 26: To make "site" even harder to use but be backwards compatible, we can make it kw-only

`def __init__(self, username, password, site_id=None, user_id_to_impersonate=None, *, site=None):`

Since site and site_id are the same thing and are the 3rd positional in both cases, it will continue to work for legacy code, or code that is using site as a kw-arg.

tableauserverclient/models/tableau_auth.py
```diff
@@ -1,41 +1,70 @@
...
+    @property
+    def credentials(self):
+        credentials = "Credentials can be username/password, Personal Access Token, or JWT"
+        +"This method returns values to set as an attribute on the credentials element of the request"
+
+    def __repr__(self):
+        display = "All Credentials types must have a debug display that does not print secrets"
+
+
+def deprecate_site_attribute():
```
t8y8, line 15: At this point, maybe we should just pick a release to kill it as well, it's been many years.

t8y8 (approved):
🚀 with minor notes

andybigbear:
close this PR and will use Jac's PR for adding the JWT: https://github.com/tableau/server-client-python/pull/1032

andybigbear:
Hi Jac, Could you please directly add the JsonWebTokenAuth(Credentials) in this PR?

jacalata:
> Hi Jac, Could you please directly add the JsonWebTokenAuth(Credentials) in this PR?

That will be a separate PR for 22.2

andybigbear (approved):
LGTM

tableauserverclient/models/tableau_auth.py
```diff
@@ -1,41 +1,70 @@
...
+    import warnings
+
+    warnings.warn(
+        "TableauAuth(..., site=...) is deprecated, " "please use TableauAuth(..., site_id=...) instead.",
+        DeprecationWarning,
+    )
+
+
+# The traditional auth type: username/password
+class TableauAuth(Credentials):
```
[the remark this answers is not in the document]
jacalata: We don't want to update it because it's part of our public API: existing code using this library will break if we change this class name.

tableauserverclient/models/tableau_auth.py
jacalata: Let's call it 0.20, which should come about October.

tableauserverclient/models/tableau_auth.py
jacalata: will skip this and just delete it after creating 0.19

Issue #512, opened 2019-10-18
'filter' object has no attribute 'pop'


mrreyes512:
In the [export.py line 46](https://github.com/tableau/server-client-python/blob/b260ebcafec6e6ebe89597f13e1e18ed9ca55406/samples/export.py#L46)

I'm getting the below error... 

```bash
Traceback (most recent call last):
  File "export.py", line 84, in <module>
    main()
  File "export.py", line 56, in main
    view = views.pop()
AttributeError: 'filter' object has no attribute 'pop'
```

I'm calling it as illustrated below:

```bash
python export.py --csv --file export1.csv `report_ID`
```

Do i need to specify any further information?

andrewbcooper:
I had the same issue and found this on Stackoverflow.  Are you using Python 3.x?

https://stackoverflow.com/questions/34820657/filter-object-has-no-attribute-pop

JoeyKwok:
Mine returns 'IndexError: pop from empty list', using python 3.6 corrected with list.
I am using view.id as report_ID. Am I using the correct one?

bcantoni:
Confirmed this is still a bug in the sample. There are 2 issues here:

* Making the sample work correctly in Python 3.x (as @andrewbcooper points out)
* Better handling for the case where the resource_id is not found (as @JoeyKwok points out)

jacalata:
The updated samples should cover this nicely.

PR #1034, opened 2022-05-06
Jac/export sample fix

jacalata:
delete sample script that manually built a pdf of a whole workbook, replaced with direct call to do that
delete sample script that duplicated the 'export image' feature
replaced the need for filter-> list by called get_by_id


jacalata:
Will fix https://github.com/tableau/server-client-python/issues/512

PR #1035, opened 2022-05-08
Jac/redact embedded creds

### LINKED ISSUE #1024: When we publish workbooks or datasources and use debug mode, user and password are displayed in plain text

jacalata:
call a redaction method before logging workbook/ds file contents to overwrite embedded credentials with junk  (fixes #1024  for real)
does not work in 3.6


tableauserverclient/helpers/strings.py
```diff
@@ -0,0 +1,66 @@
...
+            search_start = replacement_begin + 8
+        except ValueError:
+            search_start = -1
+    content = content.replace(target, replacement)
+    return content
+
+
+@singledispatch
+def redact(content):
+    # softly failing to redact in python 3.6, which is not technically supported
```
vogelsgesang, line 51: we should rather fail hard.
Security is not optional, and customers should know immediately if they are accidentally logging secrets

tableauserverclient/helpers/strings.py
```diff
@@ -0,0 +1,66 @@
...
+    search_start: int = 0
+    while search_start >= 0:
+        try:
+            replacement_begin: int = content.index(target, search_start) + 8
+            content = _replace(content, replacement_begin, replacement)
+            search_start = replacement_begin + 8
+        except ValueError:
+            search_start = -1
+    content = content.replace(target, replacement)
+    return content
```
vogelsgesang, line 46: how is this supposed to work?
I fail to see how this reliably removes passwords from the string

vogelsgesang (changes requested):

tableauserverclient/helpers/strings.py
jacalata: The current problem is that we are logging the xml of a datasource or workbook, and including the password for data connections if it is embedded. This shows up in the format password=blahblah, so we are redacting both the word 'password' and the text immediately *after* the word. 
I haven't written it to be more specific or do any xml handling because I want it to be able to apply to whatever string is being logged, without us having to identify all the attributes that we use for passwords.

tableauserverclient/helpers/strings.py
jacalata: yea, it looks like we already half-dropped support for 3.6 so I'll drop it altogether.

tableauserverclient/helpers/strings.py
vogelsgesang: Thanks for the high-level explanation! I agree with the general approach.
The implementation, however, does not do what you are describing...

It only partially replaces the text after the word.

If I call
```
_redact_typeful("password=my_super_secret_passphrase_which_nobody_should_ever_see", "password", "redacted")
```
I get
> redactedredactedr_secret_passphrase_which_nobody_should_ever_see

Note how a significant part of my password is still visible

jacalata: True. I've added the ability to keep going until the first quote or xml end-bracket character. This will still leave part of some passwords visible, but seems like a decent trade-off?

t8y8:
The approach breaks my brain a little -- given that workbooks/datasources have been XML for ~15 years and probably will remain so, I would advocate an approach where we just strip any element or attribute with a tag name or attribute name of 'password', 'secret' etc.

It would protect you from parsing xml manually and be easier to troubleshoot in the future I think.

Something like 
`root.findall(".//[@password]"` should do it

tableauserverclient/helpers/strings.py
jacalata: removed this logic altogether

tableauserverclient/helpers/strings.py
```diff
@@ -0,0 +1,65 @@
...
+T = TypeVar("T", str, bytes)
+
+
+# TODO: ideally this would be in the logging config
+def safe_to_log(server_response: requests.Response) -> str:
+    """Checks if the server_response content is not xml (eg binary image or zip)
+    and replaces it with a constant"""
+    ALLOWED_CONTENT_TYPES = ("application/xml", "application/xml;charset=utf-8")
+    if server_response.headers.get("Content-Type", None) not in ALLOWED_CONTENT_TYPES:
+        return "[Truncated File Contents]"
```
mwhagedorn, line 20: Is "Redacted" more correct here?

mwhagedorn: just a thought

test/request_factory/test_workbook_requests.py
```diff
@@ -12,3 +15,41 @@ def test_embedded_extract_req(self):
     def test_generate_xml(self):
         workbook_item: TSC.WorkbookItem = TSC.WorkbookItem("name", "project_id")
         TSC_RF.RequestFactory.Workbook._generate_xml(workbook_item)
+
+    def test_generate_xml_invalid_connection(self):
```
mwhagedorn, line 19: next couple tests seem to not be about sensitive data, is this just fixing code coverage misses from the past?

mwhagedorn (commented):
looks good, couple minor questions, no blockers

test/request_factory/test_workbook_requests.py
jacalata: yep, these cover a couple of error checks I added in request_factory while trying to write the actual test I wanted.

tableauserverclient/helpers/strings.py
jacalata: Hmm, I think better not to say "redacted" because it isn't using our "redaction" method, but yea, truncated sounds inaccurate. Not sure what would be a good alternative.

Issue #1020, opened 2022-03-31
[support a REST API] [Adding option mapAssetsTo to Remove User from Site method]


SawyerMoscow:
## Request Type
If you know, say which of these types your request is in the title, and follow the suggestions for that type when writing your description.

****Type 1: support a REST API:**** 
DELETE /api/api-version/sites/site-id/users/user-id?mapAssetsTo=new-user-id
https://help.tableau.com/v2020.3/api/rest_api/en-us/REST/rest_api_ref.htm#remove_user_from_site

## Description
In REST API version 3.9 appeared option mapAssetsTo for Remove User from Site method allows transferring to another user the contents owned by the user being removed. I didn't find this option implementation in TSC Python library.
I think I can sign a CLA and prepare PR by myself if you approved this issue.

jacalata:
sounds good to me - happy for you to build it. Do you have a specific design in mind?

SawyerMoscow:
I thought just to add parameter new_user_id with default None to function remove here:
https://github.com/tableau/server-client-python/blob/master/tableauserverclient/server/endpoint/users_endpoint.py#L66
Add decorator parameter_added_in on this function.
And then extend url with this code:
`url += "?mapAssetsTo={0}".format(new_user_id)`
if new_user_id is given.

PR #1036, opened 2022-05-11
Reassign content on user removal

jorwoods:
Add support for `mapAssetsTo` endpoint option.


jorwoods:
Adds functionality requested in #1020

PR #1039, opened 2022-05-16
Do not eagerly fetch content in case a stream was requested

### LINKED ISSUE #1037: Datasource download held in memory
**Describe the bug**
When downloading very large datasources the entire file is held in system memory rather than streamed to disk resulting in excessive memory usage (30GB extract used all available resources on a machine with 64GB RAM).  

**Versions**
 - Tableau Server version: 2021.3.8 (20213.22.0213.1115) 64-bit Windows
 - Python version: 3.9
 - TSC library version: 0.17.0

**To Reproduce**
server = TSC.Server(tab_server_url)
server.use_server_version()
tab_auth = TSC.TableauAuth(username=user,
                                              password=pw,
                                              site_id=tab_site
                                              )
with server.auth.sign_in(tab_auth):
    req_options = TSC.RequestOptions()
    req_options.filter.add(TSC.Filter(TSC.RequestOptions.Field.ProjectName,
                                      tab_project
                                      )
                           )
    req_options.filter.add(TSC.Filter(TSC.RequestOptions.Field.Name,
                                      extract_name
                                      )
                           )
    ds_datasource, ds_pag_item = server.datasources.get(req_options)
    server.datasources.download(datasource_id=ds_datasource[0].id,
                                filepath=str(save_file_path)
                                     )

**Results**
This download did complete, however system resources were at 100% usage and no file exists within directory until completion.

jorwoods:
This should fix #1037. It appears to me that `Endpoint._make_request` was eagerly fetching content to be able to evaluate and log it, when it needs to allow the content to be iterated over if it is an octet-stream.

### DIFF
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -73,15 +73,18 @@ class Endpoint(object):
 
         server_response = method(url, **parameters)
 
+        # Check if response is xml, and if so, validate namespace. Checking the
+        # content type header prevents eager evaluation of streaming requests.
+        if server_response.headers.get("Content-Type") == "application/xml":
+            self.parent_srv._namespace.detect(server_response.content)
+        self._check_status(server_response)
+
         # Response.content is a property. Calling it will load the entire response into memory. Checking if the
         # content-type is an octet-stream accomplishes the same goal without eagerly loading content.
         stream = parameters.get("stream", False)
         stream = stream or (server_response.headers.get("Content-Type") == "application/octet-stream")
         if stream:
             return server_response
-
-        self.parent_srv._namespace.detect(server_response.content)
-        self._check_status(server_response)
 
         # This check is to determine if the response is a text response (xml or otherwise)
         # so that we do not attempt to log bytes and other binary data.
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -75,12 +75,13 @@ class Endpoint(object):
 
         # Response.content is a property. Calling it will load the entire response into memory. Checking if the
         # content-type is an octet-stream accomplishes the same goal without eagerly loading content.
-        if server_response.headers.get("Content-Type") != "application/octet-stream":
-            self.parent_srv._namespace.detect(server_response.content)
-        self._check_status(server_response)
-
-        if server_response.headers.get("Content-Type") == "application/octet-stream":
+        stream = parameters.get('stream', False)
+        stream = stream or (server_response.headers.get("Content-Type") == "application/octet-stream")
+        if stream:
             return server_response
+
+        self.parent_srv._namespace.detect(server_response.content)
+        self._check_status(server_response)
 
         # This check is to determine if the response is a text response (xml or otherwise)
         # so that we do not attempt to log bytes and other binary data.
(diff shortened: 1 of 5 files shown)


t8y8 (approved):
🚀 

Could we do a test with bytesio or similar? (Did I make a test when I added the streaming?)

Also, I've always been skeptical of the namespace detection.... I think it might be safe to rely on the namespace as it has been for years, and maybe just document that you can override it if you're using Tableau Server 8.3 or whatever :D

t8y8:
The comments about namespace detection are more musings for a future PR, not needed here.

jorwoods:
@t8y8 Added a test. In so doing, discovered that the way the `Views` endpoint fetches for excel and csv only worked because it fetched the content eagerly. Changed them into generators so that content can be streamed as appropriate.

tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -64,9 +69,17 @@ def _make_request(
 
         logger.debug("request {}, url: {}".format(method.__name__, url))
         if content:
-            logger.debug("request content: {}".format(content[:1000]))
+            logger.debug("request content: %r", content[:1000])
```
t8y8, line 72: You should be able to to do `{!r}` for a repr format. Down with the old string substitution!

t8y8 (approved):
🚀 small nit

tableauserverclient/server/endpoint/endpoint.py
jorwoods: @t8y8  I normally agree with you, f-strings wherever possible! However, in this case, it is done this way to ensure that line never actually runs unless the log level is debug, whereas an f-string or format would eagerly evaluate to only then be discarded. 

https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles

t8y8: Gross, but ok :)

t8y8 (approved):
🚀

tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -64,9 +69,17 @@ def _make_request(
...
 
         server_response = method(url, **parameters)
+
+        # Response.content is a property. Calling it will load the entire response into memory. Checking if the
+        # content-type is an octet-stream accomplishes the same goal without eagerly loading content.
+        stream = parameters.get("stream", False)
+        stream = stream or (server_response.headers.get("Content-Type") == "application/octet-stream")
+        if stream:
+            return server_response
```
jacalata, line 81: don't we still want to check the status of the response? the _check_status method as is does refer to response.content, but that could be changed.

jacalata:
Do these changes affect the samples at all? Mostly wondering about the change in the csv/etc methods

jorwoods:
> Do these changes affect the samples at all? Mostly wondering about the change in the csv/etc methods

I only saw one sample that used csv, and this would not affect that sample. What would be affected though is that the open connection to the server is not completed until the generators for `View.csv` and `View.excel` have actually been consumed. I think that keeping the connection open is a fair compromise to not eagerly fetch all of the data and load into memory. Other than that caveat, all internal uses of how `_make_request` fetches data remains the same.

tableauserverclient/server/endpoint/endpoint.py
jorwoods: Reordered things so the response status code is properly checked before returning the stream.

PR #1047, opened 2022-05-28
Fix QuerySet slicing logic

jorwoods:
The old logic for handling slices on QuerySet objects worked for toy examples when the result set was small, but would crash if having to work through too many items. This logic will work for  all sizes of results.

### DIFF
--- a/tableauserverclient/server/query.py
+++ b/tableauserverclient/server/query.py
@@ -60,24 +60,7 @@ class QuerySet:
             k_range = range(start, stop, step)
             if all(i in page_range for i in k_range):
                 return self._result_cache[k]
-            else:
-                # To collect all results for the requested slice, we have to
-                # find what page range query. If the step size is one, pull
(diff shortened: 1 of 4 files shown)


jacalata:
looks good, I know I've made it a little ugly with the rebasing. 

Is it possible to write a test that shows this fix? What kind of sizes were you seeing the crash at?

jacalata:
(I will fix the conflicts/merge it, don't worry about those)

jorwoods:
Turns out the real issue was it trying to span pages. I have a better fix incoming. Working on writing the test for it.

jorwoods:
The real issue was requesting an index that was a multiple of the page size, which would happen when trying to span pages when slicing. It would get stuck in an infinite loop whenever the requested item number is a multiple of the page size. Fixed this by incrementing the `k` number by 1 whenever the requested index is not in the current cache's range. This makes it so that the function will increment to the next page as appropriate when the slice requires it.

PR #1049, opened 2022-05-31
Prepare release 0.19

jacalata:
[Fix createextracts bug]
[Add Status and StartedAt filters for jobs endpoint]
[Workbook item nullable project]
[create single Credentials class ]
[Reassign content on user removal]
[samples fix]
[Add black and mypy configurations to pyproject.toml ]

[Extract refreshable item IDs from job XML response ]
[Do not eagerly fetch content when a stream was requested]


jorwoods:
#1047 should probably be included too so that an earlier mistake I made in slicing on querysets does not make it into something pip install

jacalata:
@jorwoods I restored it again, should be good now?

PR #1044, opened 2022-05-26
Solution for issue number #1033

amar-kumar-yadav:
Tableau 2021.4 has introduced new set of permissions which are as follows:
Workbooks
  • Run Explain Data
  • Create/Refresh Metrics
Data Sources
  • Save As
  • Move
Ask Data Lenses (all new)
  • View
  • Overwrite
  • Move
  • Delete
  • Set Permissions

These permissions are missing in the current wrapper class.
I have added these permissions in the Model and also created a new populate permission method for Lense permissions.

PR #1053, opened 2022-06-02
Jac/amar kumar yadav 1044

jacalata:
not sure why the black/mypy jobs were failing for PR #1044 so pushing a new one in my name to check if that fixes it 🤷

### DIFF
--- a/tableauserverclient/models/permissions_item.py
+++ b/tableauserverclient/models/permissions_item.py
@@ -48,7 +48,7 @@ class Permission:
         Table = "table"
         Database = "database"
         View = "view"
-        Lense = 'lense'
+        Lens = 'lense'
 
 
 class PermissionsRule(object):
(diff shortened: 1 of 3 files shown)


amar-kumar-yadav:
Hi @jacalata , i see that you have removed "e" from permission_item.py line 51. Wouldn't it fail because when you do the API call for populating the permissions you add "s" after every item for eg. workbook's', flow's'. And for Lens it would add upto Lens's'  but the actual API call has Lens'es'. This is why i kept Lens = "lense" in permission_item.py line 51 rather than Lens = "Lens". Please let me know if my understanding is incorrect.

jacalata:
I missed that, you are correct. I've added sample code to validate the methods and updated that logic in a new pull request, since I accidentally merged this one to master anyway https://github.com/tableau/server-client-python/pull/1054

amar-kumar-yadav:
when can we expect the update in the master branch? @jacalata

PR #1060, opened 2022-06-24
Remove vf support from populate_excel

### LINKED ISSUE #1059: Remove CSVRequestOptions from populate_excel
https://github.com/tableau/server-client-python/blob/1eeaca8709f548b73d7306a1251322c784e656c8/tableauserverclient/server/endpoint/views_endpoint.py#L129

I can remove the CSVRequestOptions from being accepted by populate_excel. A better user experience would be to add the server side functionality.

jorwoods:
Closes #1059. Ideally server side would support view filters, but since it does not, removing that from TSC.


jacalata:
Thanks - I opened a bug internally to try and get this properly supported.

PR #1068, opened 2022-07-13
Add version awareness to site create/update methods

jacalata:
Added a sample handling sites, fixed sites operations so they work against different API versions.


tableauserverclient/server/endpoint/exceptions.py
```diff
@@ -23,40 +27,41 @@ def from_response(cls, resp, ns):
...
         self.content = server_response.content
+        self.url = request_url or "server"
 
     def __str__(self):
-        return "\n\nError status code: {0}\n{1}".format(self.code, self.content)
+        return "\n\nInternal error {0} at {1}\n{2}".format(self.code, self.url, self.content)
 
 
-class MissingRequiredFieldError(Exception):
+class MissingRequiredFieldError(TableauError):
```
mmuttreja-tableau, line 40: What's the purpose of this class (and the ones below) ?

mmuttreja-tableau (approved):

tableauserverclient/server/endpoint/exceptions.py
jacalata: To be able to return more informative errors

PR #1070, opened 2022-07-16
Log RequestOptions params

jorwoods:
Write RequestOptions params to debug log. This is helpful when troubleshooting requests to see how filters get applied.

### DIFF
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -1,5 +1,7 @@
 from ..models.property_decorators import property_is_int
+import logging
 
+logger = logging.getLogger("tableau.request_options")
 
 class RequestOptionsBase(object):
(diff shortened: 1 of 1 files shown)


bcantoni:
@jorwoods this looks like a nice addition. Any concerns about logging sensitive information here? I believe at the DEBUG level we're already logging the full request & response, so adding this doesn't log anything new or worse.

jorwoods:
@bcantoni I thought about this and do not think it will contain sensitive information. All of the sensitive information that gets passed is in the body of requests, not in what gets passed as part of the query string. (Unless there is a case I am forgetting about)

bcantoni (approved):
I agree - also it's only under debug level which is logging a bunch from all the other requests. Nice addition!

PR #1075, opened 2022-07-23
Jac/client version header

jacalata:
Add a user-agent-header with details about the current version of tsc being used.


tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -26,18 +27,29 @@
...
+
+            _version_header = __TSC_VERSION__
+
         headers = {}
         if auth_token is not None:
             headers["x-tableau-auth"] = auth_token
         if content_type is not None:
             headers["content-type"] = content_type
-
+        headers["User-Agent"] = "Tableau Server Client {}".format(_version_header)
```
anyoung-tableau, line 52: User-Agent headers are generally of the form `<product>/<version>`. Thoughts on using `Tableau Server Client/{}`?

https://www.rfc-editor.org/rfc/rfc9110.html#name-user-agent

jacalata: good point!

PR #1078, opened 2022-07-23
Update & Extend documentation

bossenti:
While working with the tableau server-client-python library I stumbled about some bugs in the documentation which I fixed within this PR.
Furthermore, I've created documentation for the method `add_to_schedule()` of the `schedules`-endpoint as this was missing so far.

Hopefully this PR fits so far, otherwise feel free to let me know.

Thanks for this great library, I really like the API of the client. It's fun to work with it :)

### DIFF
--- a/docs/api-ref.md
+++ b/docs/api-ref.md
@@ -560,7 +560,7 @@ None. A list of `ConnectionItem` objects are added to the data source (`datasour
 #### datasources.publish
 
 ```py
-datasources.publish(datasource_item, file_path, mode, connection_credentials=None)
+datasources.publish(datasource_item, file, mode, connection_credentials=None, as_job=False)
 ```
 
 Publishes a data source to a server, or appends data to an existing data source.
@@ -2682,13 +2682,13 @@ The schedule methods are based upon the endpoints for schedules in the REST API 
 
 Source files: server/endpoint/schedules_endpoint.py
 
-#### schedule.add_to_schedule
+#### schedules.add_to_schedule
 
 ```py
 schedules.add_to_schedule(schedule_id)
 ```
 
-Adds a `DataSourceItem` or a `WorkbookItem` to a schedule.
+Adds a `DatasourceItem`, `FlowItem`, or `WorkbookItem` to a schedule.
 
 
 REST API: [Add Data Source to Schedule](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_jobs_tasks_and_schedules.htm#add_data_source_to_schedule)
@@ -2701,7 +2701,7 @@ Name  |  Description
 :--- | :---
 `schedule_id` | The Tableau-internal identifier of the schedule to add to. Can be retrieved from a `ScheduleItem`, e.g., by performing `schedules.get()`
 `workbook` | (Optional) A `WorkbookItem` that should be added to the schedule. Only provide either `workbook` or `datasource`, passing multiple items will be deprecated.
-`datasource` | (Optional) A `DataSourceItem` that should be added to the schedule. Only provide either `workbook` or `datasource`, passing multiple items will be deprecated.
+`datasource` | (Optional) A `DatasourceItem` that should be added to the schedule. Only provide either `workbook` or `datasource`, passing multiple items will be deprecated.
 `flow` | (Optional) A `FlowItem` that should be added to the schedule. Only applicable if `workbook = None` and `datasource = None`.
 `task_type` | (Optional) Determines the type of task. One of `TaskItem.Type`: `ExtractRefresh`, `DataAcceleration`, `RunFlow`.
 
@@ -2731,7 +2731,7 @@ server.schedules.add_to_schedule(
 <br>
 
 
-#### schedule.create
+#### schedules.create
 
 ```py
 schedules.create(schedule_item)
@@ -2779,7 +2779,7 @@ import tableauserverclient as TSC
 <br>
 
 
-#### schedule.delete
+#### schedules.delete
 
 ```py
 schedules.delete(schedule_id)
@@ -2810,7 +2810,7 @@ Error  |  Description
 `Schedule ID undefined`  |  The identifier is not a valid identifier for a schedule on the server.
 
 
-#### schedule.get
+#### schedules.get
 
 ```py
 schedules.get([req_options=None])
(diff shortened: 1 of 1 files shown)


docs/api-ref.md
```diff
@@ -576,9 +574,10 @@ REST API: [Publish Datasource](https://help.tableau.com/current/api/rest_api/en-
 Name | Description
 :--- | :---
 `datasource_item`  |  The `datasource_item` specifies the new data source you are adding, or the data source you are appending to. If you are adding a new data source, you need to create a new `datasource_item` with a `project_id` of an existing project. The name of the data source will be the name of the file, unless you also specify a name for the new data source when you create the instance. See [DatasourceItem](#datasourceitem-class).
-`file_path`  |  The path and name of the data source to publish.
+`file`  |  The file path or file object of the data source to publish.
```
bcantoni, line 577: Could you make these same changes (rename file_path->file and add as_job) to the call above so they match:
`datasources.publish(datasource_item, file_path, mode, connection_credentials=None)`

docs/api-ref.md
```diff
@@ -2686,7 +2685,7 @@ Source files: server/endpoint/schedules_endpoint.py
 #### schedule.create
 
 ```py
-schedule.create(schedule_item)
+schedules.create(schedule_item)
```
bcantoni, line 2737: For all 3 of these schedule-->schedules changes, could you also make them in the preceding heading? (Ex: `#### schedule.create`)

docs/api-ref.md
```diff
@@ -2682,6 +2682,55 @@ The schedule methods are based upon the endpoints for schedules in the REST API
...
+
+
+
+**Parameters**
+
+Name  |  Description
+:--- | :---
+`schedule_id` | The Tableau-internal identifier of the schedule to add to. Can be retrieved from a `ScheduleItem`, e.g., by performing `schedules.get()`
+`workbook` | (Optional) A `WorkbookItem` that should be added to the schedule. Only provide either `workbook` or `datasource`, passing multiple items will be deprecated.
+`datasource` | (Optional) A `DataSourceItem` that should be added to the schedule. Only provide either `workbook` or `datasource`, passing multiple items will be deprecated.
```
bcantoni, line 2704: Same: should be DatasourceItem

docs/api-ref.md
```diff
@@ -2683,10 +2682,59 @@ The schedule methods are based upon the endpoints for schedules in the REST API
 
 Source files: server/endpoint/schedules_endpoint.py
 
+#### schedule.add_to_schedule
+
+```py
+schedules.add_to_schedule(schedule_id)
+```
+
+Adds a `DataSourceItem` or a `WorkbookItem` to a schedule.
```
bcantoni, line 2691: This should be DatasourceItem (lower case s).

bcantoni (changes requested):
@bossenti thanks for the contribution here! I've added a few corrections to make but otherwise it's looking good.

docs/api-ref.md
```diff
@@ -2683,10 +2682,59 @@ The schedule methods are based upon the endpoints for schedules in the REST API
...
+
+REST API: [Add Data Source to Schedule](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_jobs_tasks_and_schedules.htm#add_data_source_to_schedule)
+
+
+
+**Parameters**
+
+Name  |  Description
+:--- | :---
+`schedule_id` | The Tableau-internal identifier of the schedule to add to. Can be retrieved from a `ScheduleItem`, e.g., by performing `schedules.get()`
```
kramantab, line 2702: Slight edit suggested:  The identifier (`luid`) of the schedule to add the data source or workbook to.  
@bcantoni let me know if "luid" here is incorrect. Also, in the description we say "Adds a `DatasourceItem` or a `WorkbookItem` to a schedule.", but we looks like you can also add a flow?

docs/api-ref.md
kramantab: @bcantoni - more a question for you.  it says here that "passing multiple items will be deprecated."  So does this mean you can do that today and this functionality will be deprecated in the future?

bossenti: As I could get I from the source code, you can currently pass a workbook and a datasource item at the same time with one method call, which leads to PendingDeprecationWarning. A FlowItem can only be passed individually.

docs/api-ref.md
bossenti: I`ve added the `FlowItem` to the method description.
If using "luid" is appropriate, I'll adpat the parameter description as well.

docs/api-ref.md
bossenti: Done

docs/api-ref.md
bossenti: Done

docs/api-ref.md
bossenti: Done

bossenti:
Thanks @kramantab and @bcantoni for reviewing.
I've updated the PR according to four comments.

docs/api-ref.md
bcantoni: Good catch adding FlowItem as well.

@kramantab I think to be consistent with the REST API documentation we can keep the more generic "id" usage here rather than "luid". That might be a future task if the REST API docs are moving that direction: updating TSC id vs luid usage.

bossenti: I've updated the description correspondingly.

docs/api-ref.md
bcantoni: @kramantab it does sound that way (meaning: might be deprecated in the future). I'm happy with the wording @bossenti put in here. It mentions the possible deprecation without being to specific about it.

bcantoni (approved):

bcantoni:
All looks good to me. Passing it over to @kramantab for any additional feedback or changes.

@bossenti after @kramantab approves, you can merge. Please choose the "Squash and merge" option to combine all your commits into one.

kramantab (approved):
Looking good Thanks!

bossenti:
@bcantoni, @kramantab, unfortunately I can't merge the pull request because I lack the appropriate permissions:  `Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.`

bcantoni:
Done - thanks @bossenti for the contribution!

PR #1081, opened 2022-07-29
change initialization defaults

jacalata:
use_server_version = False so that we don't try and contact the server before people finish setting certs


bcantoni (approved):
Looks good!

PR #1086, opened 2022-08-05
Jac/user import

jacalata:
Debugging some of the user logic in tabcmd when I realized it really belongs in here, to be re-usable

This implements bulk importing users in a csv file: note that it still doesn't have the efficiency of a real bulk import, because the users are added one by one, but it at least makes it possible to use the file import process.


tableauserverclient/models/user_item.py
```diff
@@ -260,4 +256,157 @@ def _parse_element(user_xml, ns):
...
+                    values[UserItem.CSVImportFileItem.Column.DISPLAY_NAME],
+                    values[UserItem.CSVImportFileItem.Column.EMAIL],
+                    values[UserItem.CSVImportFileItem.Column.AUTH],
+                    None,
+                )
+            return user
+
+        @staticmethod
+        def validate_file_for_import(csv_file: io.TextIOWrapper, logger) -> List[int]:
+            num_errors = 0
```
mmuttreja-tableau, line 323: I know the code and function names are descriptive but it would be nice to have comments still about what the functions are doing.

mmuttreja-tableau (approved):
Does dev branch need a version bump? I got a error installing the development branch 
`tabcmd 2.0.6.dev13 requires tableauserverclient>=0.19, but you have tableauserverclient 0.18.0.post0.dev47 which is incompatible`

tableauserverclient/models/user_item.py
```diff
@@ -260,4 +256,157 @@ def _parse_element(user_xml, ns):
...
+                return None
+            line = line.strip().lower()
+            values: List[str] = list(map(str.strip, line.split(",")))
+            user = UserItem(values[UserItem.CSVImportFileItem.Column.USERNAME])
+            if len(values) > 1:
+                if len(values) > UserItem.CSVImportFileItem.Column.MAX:
+                    raise ValueError("Too many attributes for user import")
+                while len(values) <= UserItem.CSVImportFileItem.Column.MAX:
+                    values.append("")
+                site_role = UserItem.CSVImportFileItem.evaluate_site_role(
```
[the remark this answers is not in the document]
jacalata: Probably, but python references to nested classes are awkward and I didn't think it was worth it

bcantoni:
@mmuttreja-tableau no it won't need a version bump. It looks like you have an older install-from-source which pip can't always upgrade. In that environment try `pip uninstall -y tableauserverclient` and then pip install your tabcmd again. It should pick up and install v0.19 cleanly.

jacalata:
Versioneer labels it based on tags - I added an 0.19.1 tag to the dev branch so that may have fixed it

tableauserverclient/models/user_item.py
jacalata: Added comments

Issue #1062, opened 2022-06-28
Embedding a password (for scheduling an extract) does not work


jog-ioki:
**Describe the bug**
We are currently trying to automate the creation of workbooks on the server via Tableau APIs. Version is:
 
```
print(server.server_info.get())
ServerInfoItem: [product version: 2022.1.1, build no.:20221.22.0415.1144, REST API version:3.15]
```
 
All our workbooks have an embedded connection string to a PostgreSQL database (no published datasource!) All workbooks contain their own and a custom SQL is “contained” in each of datasource per workbook)
Uploading with a live connection works
We now try to get some workbooks and create an Extract for them. Here is example code (note that this is unrelated to TSC)
 
```
from tableau_api_lib import TableauServerConnection
srv = TableauServerConnection(tableau_server_config, env='PROD')
srv.sign_in()
 
response = srv.query_sites()
site_list = [(site['name'], site['id']) for site in response.json()['sites']['site']]
 
workbooks_df = get_workbooks_dataframe(srv)[['name', 'id']]
what_to_search = ['xyz']
 
workbooks_df = workbooks_df[workbooks_df['name'].isin(what_to_search)]
response = srv.create_extracts_for_workbook(workbook_id=workbooks_df.id.iloc[0])
```
 
Running this will give us an error like: 
```
    Affected data source
    Embedded in workbook
    Error message
    Connectionless Failure (status code = 10000, Missing password. Invalid username or password.)
    Requested at
    June 21, 2022, 1:25 PM
    Requested by
    xx@xx.com
    Failed at
    June 21, 2022, 1:25 PM
    Suggested action
    **Check the Data Connection page for necessary updates to an access token or embedded credentials.**
```
The action seem to be clear – we need to embed the password.
 
Problem: We are not able to embed the password to start the extract.
Here is the code - we tried using TSC for it:
 
```
token = TSC.PersonalAccessTokenAuth(token_name='xxx', personal_access_token=pat, site_id='')
server = TSC.Server("xxx", use_server_version=True)
with server.auth.sign_in(token):
   
    workbook = server.workbooks.get_by_id('xyz')
    server.workbooks.populate_connections(workbook)
    connection1 = workbook.connections[0]
 
    datasource1 = server.datasources.get_by_id(connection1.datasource_id)
    server.datasources.populate_connections(datasource1)
 
    connection1.embed_password=True
 
    server.workbooks.update_connection(workbook, connection1)
```

**Versions**
Details of your environment, including:
[...]

PR #1099, opened 2022-09-01
Use explicit None identity check for datasource updates

jorwoods:
Makes it so that update requests for connections do an explicit identity check for if they are None instead of gating falsey values. Resolves issues like #1062.


jacalata:
Thanks!

PR #1111, opened 2022-09-14
Allow injection of sessions

MrwanBaghdad:
This PR allows injection of `requests.Session` objects.
This covers use cases when an end user may want to have more control over the session in case of proxies, custom authentication flows enforced by proxies in front of the Tableau server, and custom headers need to be injected to allow access to the Tableau server.


MrwanBaghdad:
@jacalata Can you please take a look?

jacalata:
Very nice! We do have another call where it sets _session = requests.Session() in the _set_auth() method - does that need to be overridden or updated somehow as well? 

I've also rebased this for you onto our new dev branch.

MrwanBaghdad:
@jacalata  That's a good point. I've updated the PR to allow injection of a callable `session_factory`, which returns a `request.Session` object when called, that'd allow recreation of `request.Session` object by the end-user when the `_clear_auth` is called

PR #1113, opened 2022-09-19
Add build package to pip install

jorwoods:
Development branch missing, so targeting master.

This PR adds the build package to the pip install step explicitly so that build can run and create the artifacts to publish to PyPI.


jacalata (approved):
Thanks! I forgot that we have this repo set to delete branches after they merge to master, will re-create development.

PR #1115, opened 2022-09-20
Jac/smoke tests

jacalata:
- cleaned up the client version 
- made it available to the calling program and *actually* in the headers
-  Added a smoke test action for monitoring what's on PyPi, and a sample smoke test to help people debug


.github/workflows/pypi-smoke-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+        python-version: ['3.x']
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - name: Set up Python ${{ matrix.python-version }} on ${{ matrix.os }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: pip install Tabcmd
```
nikhillakshman, line 29: I think you meant tsc

.github/workflows/pypi-smoke-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+      run: |
+        pip uninstall tableauserverclient
+        pip install tableauserverclient
+    - name: Launch app
+      run: |
+        python -c "import tableauserverclient as TSC
+        server = TSC.Server('Fake-Server-Url', use_server_version=False)"
```
nikhillakshman, line 36: For my understanding, how does this look when it fails?

nikhillakshman (approved):

.github/workflows/pypi-smoke-tests.yml
jacalata: I'm not certain, because sometimes the github environment does something differently to what I expect. But it should check the exit code of the call to python: if any exceptions are thrown that will be something different to 0, and will be called a failed step.

Issue #1063, opened 2022-06-29
Feature : Add permission control for Data Roles, Ask Data Lenses, and Metrics


RumblingHacked:
## Summary
Need to add support for modifying permissions for Data Roles, Ask Data Lenses, and Metrics.

## Request Type
If you know, say which of these types your request is in the title, and follow the suggestions for that type when writing your description.

****Type 1: support a REST API:**** 
If it is functionality that already exists in the [REST API](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm), example API calls are the clearest way to explain your request.
Need the following calls supported in TSC and the permission list updated.

Data roles:
DELETE /api/api-version/sites/site-luid/projects/project-luid/default-permissions/dataroles/groups/group-luid/capability-name/capability-mode

Lenses:
DELETE /api/api-version/sites/site-luid/projects/project-luid/default-permissions/lenses/groups/group-luid/capability-name/capability-mode

Metrics:
DELETE /api/api-version/sites/site-luid/projects/project-luid/default-permissions/metrics/groups/group-luid/capability-name/capability-mode

Data roles:
PUT /api/api-version/sites/site-luid/projects/project-luid/default-permissions/dataroles

Lenses:
PUT /api/api-version/sites/site-luid/projects/project-luid/default-permissions/lenses

Metrics:
PUT /api/api-version/sites/site-luid/projects/project-luid/default-permissions/metrics

****Type 2: add a REST API and support it in tsc.****
If it is functionality that can be achieved somehow on Tableau Server but not through the REST API, describe the current way to do it. (e.g: functionality that is available in the Web UI, or by using the Hyper API). For UI, screenshots can be helpful.

****Type 3: new functionality****
Requests for totally new functionality will generally be passed to the relevant dev team, but we probably can't give any useful estimate of how or when it might be implemented. If it is a feature that is 'about' the API or programmable access, here might be the best place to suggest it, but generally feature requests will be more visible in the [Tableau Community Ideas](https://community.tableau.com/s/ideas) forum and should go there instead.

## Description
A clear and concise description of what the feature request is. If you think that the value of this feature might not be obvious, include information like how often it is needed, amount of work saved, etc. If your feature request is related to a file or server in a specific state, describe the starting state when the feature can be used, and the end state after using it. If it involves modifying files, an example file may be helpful. 

This feature is needed to automate permission assignments for everything that is available in Tableau.

[...]

PR #1120, opened 2022-09-23
Add permission control for Data Roles and Metrics (Issue #1063)

TrimPeachu:
* Add functions to control permissions

* Add resources

### DIFF
--- a/tableauserverclient/models/tableau_types.py
+++ b/tableauserverclient/models/tableau_types.py
@@ -29,9 +29,5 @@ TableauItem = Union[DatasourceItem, FlowItem, ProjectItem, ViewItem, WorkbookIte
 def plural_type(content_type: Resource) -> str:
     if content_type == Resource.Lens:
         return "lenses"
-    elif content_type == Resource.Datarole:
-        return "dataroles"
-    elif content_type == Resource.Metric:
-        return "metrics"
     else:
         return "{}s".format(content_type)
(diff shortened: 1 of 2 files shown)


tableauserverclient/models/tableau_types.py
```diff
@@ -27,5 +29,9 @@ class Resource:
 def plural_type(content_type: Resource) -> str:
     if content_type == Resource.Lens:
         return "lenses"
+    elif content_type == Resource.Datarole:
```
jacalata, line 32: this isn't necessary: lenses is a special case because it can't just add "s" on the end, but datarole and metric can use the default behavior.

jacalata (changes requested):

tableauserverclient/models/tableau_types.py
TrimPeachu: Understood. So should I delete this part and create new PR? Not sure why this PR was closed. 
Thanks

jacalata:
oops, this got automatically closed when I removed the dev branch. Re-opening.

tableauserverclient/models/tableau_types.py
jacalata: No, that was an accident, sorry. You can just make the update in this PR.

TrimPeachu: Ok, np, updated

t8y8 (approved):

jaybythebay:
I think this will close https://github.com/tableau/server-client-python/issues/1063.  When will this get released?

jacalata (approved):

PR #1125, opened 2022-09-27
fix behavior when url has no protocol

jacalata:
Defaulting to http if no protocol is entered, since that is the behavior we expect in tabcmd currently.
Manually handle a redirect on login, so we don't send a GET request that fails.

### DIFF
--- a/tableauserverclient/server/endpoint/auth_endpoint.py
+++ b/tableauserverclient/server/endpoint/auth_endpoint.py
@@ -28,7 +28,18 @@ class Auth(Endpoint):
     def sign_in(self, auth_req):
         url = "{0}/{1}".format(self.baseurl, "signin")
         signin_req = RequestFactory.Auth.signin_req(auth_req)
-        server_response = self.parent_srv.session.post(url, data=signin_req, **self.parent_srv.http_options)
+        server_response = self.parent_srv.session.post(
+            url, data=signin_req, **self.parent_srv.http_options, allow_redirects=False
+        )
+        # manually handle a redirect so that we send the correct POST request instead of GET
+        # this will make e.g http://online.tableau.com work to redirect to http://east.online.tableau.com
+        if server_response.status_code == 301:
+            server_response = self.parent_srv.session.post(
+                server_response.headers["Location"],
+                data=signin_req,
+                **self.parent_srv.http_options,
+                allow_redirects=False,
+            )
         self.parent_srv._namespace.detect(server_response.content)
         self._check_status(server_response, url)
         parsed_response = fromstring(server_response.content)
--- a/tableauserverclient/server/server.py
+++ b/tableauserverclient/server/server.py
@@ -61,7 +61,7 @@ class Server(object):
         self._site_id = None
         self._user_id = None
 
-        self._server_address = server_address
+        self._server_address: str = server_address
         self._session_factory = session_factory or requests.session
 
         self.auth = Auth(self)
@@ -103,6 +103,8 @@ class Server(object):
 
     def validate_server_connection(self):
         try:
+            if not self._server_address.startswith("http"):
+                self._server_address = "http://" + self._server_address
             self._session.prepare_request(requests.Request("GET", url=self._server_address, params=self._http_options))
         except Exception as req_ex:
             warnings.warn("Invalid server initialization\n  {}".format(req_ex.__str__()), UserWarning)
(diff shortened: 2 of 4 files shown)


tableauserverclient/server/endpoint/auth_endpoint.py
```diff
@@ -28,7 +28,18 @@ def baseurl(self):
...
         signin_req = RequestFactory.Auth.signin_req(auth_req)
-        server_response = self.parent_srv.session.post(url, data=signin_req, **self.parent_srv.http_options)
+        server_response = self.parent_srv.session.post(
+            url, data=signin_req, **self.parent_srv.http_options, allow_redirects=False
+        )
+        # manually handle a redirect so that we send the correct POST request instead of GET
+        # this will make e.g http://online.tableau.com work to redirect to http://east.online.tableau.com
+        if server_response.status_code == 301:
+            server_response = self.parent_srv.session.post(
+                server_response.headers["Location"],
```
anyoung-tableau, line 38: Any concerns with case-sensitivity of the header name?

bhuvneshdev (approved):
LGTM

tableauserverclient/server/endpoint/auth_endpoint.py
```diff
@@ -28,7 +28,18 @@ def baseurl(self):
...
+                data=signin_req,
+                **self.parent_srv.http_options,
+                allow_redirects=False,
```
anyoung-tableau, line 41: Any chance *this* `post` could redirect too?

tableauserverclient/server/server.py
```diff
@@ -103,6 +103,8 @@ def __init__(self, server_address, use_server_version=False, http_options=None,
 
     def validate_server_connection(self):
         try:
+            if not self._server_address.startswith("http"):
+                self._server_address = "http://" + self._server_address
```
anyoung-tableau, line 107: Seems weird to default to `http` in 2022.

anyoung-tableau (approved):

tableauserverclient/server/server.py
jacalata: yep - maintaining backward compatibility for tabcmd

tableauserverclient/server/endpoint/auth_endpoint.py
jacalata: python requests library uses a case-insensitive lookup for headers, so it doesn't matter what casing we use here.

tableauserverclient/server/endpoint/auth_endpoint.py
jacalata: I think it could, but I don't know how many is reasonable and I don't want to set up an infinite loop.  Thoughts?

anyoung-tableau: As a datapoint, .NET defaults to 50. https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.maxautomaticredirections?view=net-6.0

jacalata: That seems like a lot! This question gave a reasonable scenario of four redirections, and it looks like browsers allow 10-20, which is closer to what I thought. I think 20 seems generous but not too much of a time-waster.

tableauserverclient/models/server_info_item.py
```diff
@@ -32,7 +35,11 @@ def rest_api_version(self):
 
     @classmethod
     def from_response(cls, resp, ns):
-        parsed_response = fromstring(resp)
+        try:
+            parsed_response = fromstring(resp)
+        except xml.etree.ElementTree.ParseError as error:
+            warnings.warn("Unexpected response for ServerInfo: {}".format(resp))
```
t8y8, line 41: this isn't the general intended use for warnings -- they should be things thrown in tests or lints or ci/cd.

We should re-raise the exception, or if it isn't important just log and move on, but not a warnings.wan

t8y8 (changes requested):

t8y8 (approved):

tableauserverclient/server/endpoint/auth_endpoint.py
jacalata: logged for fixing in another change: https://github.com/tableau/server-client-python/issues/1127

PR #1136, opened 2022-10-28
Jac/site by name

jacalata:
I don't think we've ever supported this query. 

Added a convenience method for users to get the current site info.


t8y8:
Query Site by Name was a holdover from the 1.0 internal API, but it's been publicly documented forever, and still is (https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#query_site)

I can't speak to whether or not it works now, but it used to!

jacalata:
Ohhh, I didn't focus enough to think that it's not using our normal filtering, but the sites/my-site?key=[name/content-url]. So I know for sure it's not working now (bad request: key name not allowed) but for the past I only confirmed that we never supported it through regular filtering. I did think I remembered using it in the past so phew!

jacalata:
too stale, easier to restart.

PR #1142, opened 2022-11-15
Update publish-pypi.yml

jacalata:


bcantoni (changes requested):
@jacalata the right side of this looks wrong - we don't have a `main` branch on this project (it's still `master`)

PR #1159, opened 2022-12-22
Fix github reference for publishing action

bcantoni:
Trying again using an example I see here:
https://docs.github.com/en/actions/learn-github-actions/contexts


jacalata (approved):
¯\_(ツ)_/¯

PR #1171, opened 2023-01-12
Fix issues with connections publishing workbooks

nosnilmot:
Allow publishing using connection credentials on ConnectionItem class without ConnectionCredentials instance, as documentated

Accept empty string for username or password in connection credentials

Avoid Tableau Server internal server error when publishing with empty connection list by setting connections to None


jacalata (approved):

jacalata:
Looks good, thanks for making these fixes!

PR #1172, opened 2023-01-12
Allow download to file-like objects

nosnilmot:
Also remove code duplication in download()/download_revision() for datasources and workbooks


jacalata (changes requested):
looks nice, thanks! If you could address the mypy failures, should be good to go.

nosnilmot:
> looks nice, thanks! If you could address the mypy failures, should be good to go.

mypy failures should be addressed now

jacalata (approved):

Issue #1175, opened 2023-01-13
Some revision properties not populating


jharris126:
**Describe the bug**
Key revision properties (`current` and `created_at`) of the `RevisionItem` are not populating or populating with default values.

**Versions**
Details of your environment, including:
 - Tableau Server version (or note if using Tableau Online): **Tableau Server 2022.3**
 - Python version: **3.9**
 - TSC library version: **0.23.4**

**To Reproduce**
```python
wb = server.workbooks.get_by_id('90c3056a-eee6-4120-b1cd-194d8ba7dcb2')

server.workbooks.populate_revisions(wb)
wb_last_revision_tsc = wb.revisions[0]
print(
    wb_last_revision_tsc.revision_number,
    wb_last_revision_tsc.current,
    wb_last_revision_tsc.created_at
)

url = f'{server.baseurl}/sites/{server.site_id}/workbooks/{wb.id}/revisions'
revisions = requests.get(url, headers=request_headers).json()['revisions']

wb_last_revision = [rev for rev in revisions['revision'] if rev['current']][0]
print(
    wb_last_revision['revisionNumber'],
    wb_last_revision['current'],
    wb_last_revision['publishedAt']
)
```

**Results**
```
44 False None
44 True 2021-10-28T16:11:34Z
```
`revision_number` and `current` properties of the `RevisionItem` object are populating with default values instead of correct values.

jharris126:
https://github.com/tableau/server-client-python/blob/master/tableauserverclient/models/revision_item.py#L70-L72
It looks like the issue is the XML response using different item names than TSC is expecting.
`isCurrent` = `current`
`isDeleted` = `deleted`
`createdAt` = `publishedAt`

Not sure if this should be fixed in TSC or in the Tableau REST API itself because the expected values appear to match the Tableau REST API naming conventions more closely.

CC @jorwoods  since we are chatting about this is Slack right now

jorwoods:
Thanks @jharris126! The official documentation does not match what is actually returned by the server endpoint. https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_revisions.htm#get_workbook_revisions

jharris126:
https://github.com/tableau/server-client-python/blob/master/tableauserverclient/models/revision_item.py#L73
also need to be changed to `publisher` instead of `user`

alexepperly:
I'm having this same issue.  Got a ton of "Nones" when trying to populate the same fields you're talking about.

[...]

PR #1176, opened 2023-01-13
fix revision references where xml returned does not match docs

jharris126:
Fixes Issue #1175

XML returned by Tableau REST API differs from what TSC code is expecting as well as the [sample response in the docs site](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_revisions.htm#get_workbook_revisions). This change fixes the 4 `RevisionItem` properties with references mismatch between the xml response and the docs.


jharris126:
Hmm... just signed the CLA, how do I rerun the build?

jorwoods:
@jharris126 the test probably needs to be updated as well. https://github.com/tableau/server-client-python/blob/97747b7b253d659de7871e0b3f20c0a1702b4c2a/test/assets/workbook_revision.xml#L5

jacalata:
@alexepperly fair comment, there is always other stuff going on but a little polite impatience can definitely push things along :)

alexepperly:
> @alexepperly fair comment, there is always other stuff going on but a little polite impatience can definitely push things along :) 

image

Issue #140, opened 2017-02-08
All "Item" classes should have a repr implemented


graysonarts:

Stgocf:
can i take this?

jacalata:
Absolutely!

bcantoni:
Closing due to age and inactivity.

PR #1191, opened 2023-02-14
Jac/repr for models

jacalata:
Add __repr__ implementations for a bunch of models
Also add test to make it easy to track not-yet-implemented models
This begins addressing https://github.com/tableau/server-client-python/issues/140


tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -78,16 +78,16 @@ def _make_request(
             self.parent_srv.http_options, auth_token, content, content_type, parameters
         )
 
-        logger.debug("request {}, url: {}".format(method, url))
+        logger.debug("request method {}, url: {}".format(method.__name__, url))
         if content:
             redacted = helpers.strings.redact_xml(content[:1000])
-            logger.debug("request content: {}".format(redacted))
+            # logger.debug("request content: {}".format(redacted))
```
nikhillakshman, line 84: are the comments intentional?

jacalata: yes - they're too noisy to need most of the time, but worth adding specifically when debugging issues with server responses. The other option would be to implement a trace level for logging, but that was too much work.

nikhillakshman (approved):

bhuvneshdev (approved):

test/models/test_repr.py
```diff
@@ -0,0 +1,40 @@
...
+    PaginationItem wrapper_descriptor
+    Permission wrapper_descriptor
+    ServerInfoItem wrapper_descriptor
+    SiteItem wrapper_descriptor
+    TableItem wrapper_descriptor
+    Resource wrapper_descriptor
+    """
+
+    # not all models have __repr__ yet: see above list
+    @pytest.mark.xfail()
```
anyoung-tableau, line 34: Is there a work item tracking adding `repr` to the remaining models? Could we enable this test and remove these models from `m` at runtime?

anyoung-tableau (approved):

test/models/test_repr.py
jacalata: https://github.com/tableau/server-client-python/issues/140 but I haven't added much info there. I like the runtime idea, will add.

PR #1195, opened 2023-02-18
Implement custom view objects

jacalata:
They are not quite the same as normal views so they got their own class. 

You can
- Get a list of custom views on a site
- Get the details of a custom view
- Export an image of a custom view.
- Delete a custom view 
- Update the name or owner of a custom view.

Note that you can **only** export as image(png) - no csv or pdf export is available. 
This will enable the feature to be added to tabcmd - yay!

### DIFF
--- a/samples/explore_workbook.py
+++ b/samples/explore_workbook.py
@@ -74,6 +74,10 @@ def main():
         if all_workbooks:
             # Pick one workbook from the list
             sample_workbook = all_workbooks[0]
+            sample_workbook.name = "Name me something cooler"
+            sample_workbook.description = "That doesn't work"
+            updated: TSC.WorkbookItem = server.workbooks.update(sample_workbook)
+            print(updated.name, updated.description)
 
             # Populate views
             server.workbooks.populate_views(sample_workbook)
@@ -126,6 +130,32 @@ def main():
                 with open(args.preview_image, "wb") as f:
                     f.write(sample_workbook.preview_image)
                 print("\nDownloaded preview image of workbook to {}".format(os.path.abspath(args.preview_image)))
+
+
+            # get custom views
+            cvs, _ = server.custom_views.get()
+            for c in cvs:
+                print(c)
+
+            # for the last custom view in the list
+
+            # update the name
+            # note that this will fail if the name is already changed to this value
+            changed: TSC.CustomViewItem(id=c.id, name="I was updated by tsc")
+            verified_change = server.custom_views.update(changed)
+            print(verified_change)
+
+            # export as image. Filters etc could be added here as usual
+            server.custom_views.populate_image(c)
+            filename = c.id + "-image-export.png"
+            with open(filename, "wb") as f:
+                f.write(c.image)
+            print("saved to " + filename)
+
+            if args.delete:
+                print("deleting {}".format(c.id))
+                unlucky = TSC.CustomViewItem(c.id)
+                server.custom_views.delete(unlucky.id)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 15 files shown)


samples/explore_workbook.py
```diff
@@ -74,6 +74,10 @@ def main():
         if all_workbooks:
             # Pick one workbook from the list
             sample_workbook = all_workbooks[0]
+            sample_workbook.name = "Name me something cooler"
+            sample_workbook.description = "That doesn't work"
```
anyoung-tableau, line 78: To confirm, updating the description doesn't work?

tableauserverclient/server/endpoint/custom_views_endpoint.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Not yet implemented: pdf or csv exports
+    """
+
+    @api(version="3.18")
+    def update(self, view_item: CustomViewItem) -> Optional[CustomViewItem]:
+        if not view_item.id:
+            error = "Custom view item missing ID."
+            raise MissingRequiredFieldError(error)
+        if not (view_item.owner or view_item.name or view_item.shared):
```
anyoung-tableau, line 85: I don't see anything about updating the `shared` property in the REST API docs. Is the check on `view_item.shared` necessary?

tableauserverclient/models/custom_view_item.py
```diff
@@ -0,0 +1,156 @@
...
+                parsed_owners = UserItem.from_response_as_owner(tostring(custom_view_xml), ns)
+                if parsed_owners and len(parsed_owners) > 0:
+                    cv_item._owner = parsed_owners[0]
+
+            if view_elem is not None:
+                parsed_views = ViewItem.from_response(tostring(custom_view_xml), ns)
+                if parsed_views and len(parsed_views) > 0:
+                    cv_item._view = parsed_views[0]
+
+            if workbook_id:
```
anyoung-tableau, line 148: For my knowledge, what is the reasoning behind favoring this `workbook_id` override instead of always parsing the XML response?

anyoung-tableau (approved):

samples/explore_workbook.py
jacalata: Yes - it's an open bug, hopefully fixed soon.

tableauserverclient/server/endpoint/custom_views_endpoint.py
jacalata: It's aspirational. Actually sending a value now wouldn't work, but this means that it will work in tsc as soon as the REST API supports it.

tableauserverclient/models/custom_view_item.py
jacalata: This is matching the behavior in view_item.py, where I believe the reasoning was       
a) our Fields implementation means a View response doesn't have to contain a workbook element, but the user quite likely has the workbook id available     
 and b) if both the parameter and the xml element are available, we want a consistent priority order and chose this order as least likely to surprise a user: if they pass in a value, they will know where it came from when it shows up.

tableauserverclient/server/endpoint/custom_views_endpoint.py
anyoung-tableau: Gotcha. But might this condition prevent someone from changing a shared CV to be not-shared?

jacalata: No - this condition is just saying that if the user hasn't set any values on the view_item, don't bother sending an update request. Currently we'll send a request if they change the shared attribute, but that will fail.

anyoung-tableau: Hmm, if a user did not specify a value for either `owner` or `name` but set `shared` to `False`, this condition would still be true?

PR #1196, opened 2023-02-19
⬆️ Allow more recent version of packaging

r-richmond:
## Why make change
The current specification conflicts with new versions of black
https://github.com/psf/black/blob/d9b8a6407e2f46304a8d36b18e4a73d8e0613519/pyproject.toml#L68

## What do the new versions of packaging do
https://github.com/pypa/packaging/releases/tag/22.0
of note
1. Removes support for python 3.6 (which this package doesn't support already)
1. Adds support for python 3.11 (Which this isn't marked as supporting yet but probably should at some point)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 
 dependencies = [
     'defusedxml>=0.7.1',
-    'packaging~=21.3',
+    'packaging>=22.0',  # bumping to minimum version required by black
     'requests>=2.28',
     'urllib3~=1.26.8',
 ]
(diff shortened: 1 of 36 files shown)


r-richmond:
CLA signed previously as noted [here](https://github.com/tableau/document-api-python/pull/123#issuecomment-269398240)

jacalata:
hm, looks like that brings in some changes from black as well. Can you run black to get the code back into a valid state? It looks like the vast majority of changes are this one:
_Remove trailing newlines after code block open (https://github.com/psf/black/pull/3035) (22.6.0)_

Also, could you bump the named version to 22.0, for clarity?

r-richmond:
Addressed both black run & bumping minimum version as requested (also added comment for more clarity)

jacalata (approved):

r-richmond:
@jacalata do you have a rough estimate for when the next version will go out?

PR #1202, opened 2023-03-06
add query-tagging attribute to connection

jacalata:
need to figure out how to test this before merging

### DIFF
--- a/tableauserverclient/models/connection_item.py
+++ b/tableauserverclient/models/connection_item.py
@@ -17,6 +17,7 @@ class ConnectionItem(object):
         self.server_port: Optional[str] = None
         self.username: Optional[str] = None
         self.connection_credentials: Optional[ConnectionCredentials] = None
+        self.query_tagging: bool = None
 
     @property
     def datasource_id(self) -> Optional[str]:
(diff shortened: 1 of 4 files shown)


jacalata:
Validated this locally using one hyper datasource (cannot change the value for queryTagging) and one postgres connection (attribute successfullyl updated)

tableauserverclient/models/connection_item.py
```diff
@@ -52,6 +69,7 @@ def from_response(cls, resp, ns) -> List["ConnectionItem"]:
             connection_item.server_address = connection_xml.get("serverAddress", None)
             connection_item.server_port = connection_xml.get("serverPort", None)
             connection_item.username = connection_xml.get("userName", None)
+            connection_item._query_tagging = connection_xml.get("queryTaggingEnabled", None)
```
anyoung-tableau, line 72: Just double checking we don't need to use `string_to_bool()` here like we did on `connection_item.embed_password`

anyoung-tableau (approved):

tableauserverclient/models/connection_item.py
jacalata: good catch. Added that and tests to cover it.

PR #1204, opened 2023-03-06
catch exceptions from ServerInfo

jacalata:
Catch and log all exceptions so we can give more info on why the initial connection failed.

### DIFF
--- a/tableauserverclient/server/endpoint/server_info_endpoint.py
+++ b/tableauserverclient/server/endpoint/server_info_endpoint.py
@@ -41,5 +41,8 @@ class ServerInfo(Endpoint):
                 raise EndpointUnavailableError(e)
             raise e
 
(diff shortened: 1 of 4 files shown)


tableauserverclient/server/endpoint/server_info_endpoint.py
```diff
@@ -41,5 +41,8 @@ def get(self):
                 raise EndpointUnavailableError(e)
             raise e
 
-        self._info = ServerInfoItem.from_response(server_response.content, self.parent_srv.namespace)
+        try:
+            self._info = ServerInfoItem.from_response(server_response.content, self.parent_srv.namespace)
+        except Exception as e:
+            logging.getLogger(self.__class__.__name__).debug(server_response.content)
```
anyoung-tableau, line 47: Any value is logging out `e` here too?

anyoung-tableau (approved):

PR #1212, opened 2023-03-30
run long requests on second thread

jacalata:
added a thread for the request, so we can continue giving progress updates on the UI while downloading.
added tests - could do with more

Was not able to repro the original issue (ran into other problems with a large file) so this may not help, but it should at least capture more logs of the failure.

### DIFF
--- a/tableauserverclient/server/endpoint/datasources_endpoint.py
+++ b/tableauserverclient/server/endpoint/datasources_endpoint.py
@@ -9,8 +9,6 @@ from contextlib import closing
 from pathlib import Path
 from typing import List, Mapping, Optional, Sequence, Tuple, TYPE_CHECKING, Union
 
-import tableauserverclient
-
 if TYPE_CHECKING:
     from tableauserverclient.server import Server
     from tableauserverclient.models import PermissionsRule
--- a/tableauserverclient/server/endpoint/datasources_endpoint.py
+++ b/tableauserverclient/server/endpoint/datasources_endpoint.py
@@ -1,15 +1,12 @@
 import cgi
 import copy
 import json
-import logging
 import io
 import os
 
 from contextlib import closing
 from pathlib import Path
 from typing import List, Mapping, Optional, Sequence, Tuple, TYPE_CHECKING, Union
-
-import tableauserverclient.config
 
 if TYPE_CHECKING:
     from tableauserverclient.server import Server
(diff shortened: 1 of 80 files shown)


jacalata:
Note for review: this does contain 78 files, but 
a) ignore all the samples files, it's just an argument change. Exception: to see the new behavior, run publish_datasource.py
b) of the files under /models or /server/endpoints, almost all of them have just changed how they import the logger. You mostly just need to look at datasources_endpoint, endpoint, and fileuploads_endpoint

tableauserverclient/server/endpoint/datasources_endpoint.py
```diff
@@ -262,7 +271,11 @@ def publish(
 
         # Determine if chunking is required (64MB is the limit for single upload method)
         if file_size >= FILESIZE_LIMIT:
-            logger.info("Publishing {0} to server with chunking method (datasource over 64MB)".format(filename))
+            logger.info(
+                "Publishing {} to server with chunking method (datasource over {}MB, chunk size {}MB)".format(
```
anyoung-tableau, line 275: Are `FILESIZE_LIMIT` and `CHUNK_SIZE` in bytes, not MB?

tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -65,6 +76,54 @@ def set_user_agent(parameters):
...
+            if seconds >= 60:
+                seconds = 0
+                minutes = minutes + 1
+        return self.async_response
+
+    def wait_for_response(self, minutes, seconds, url):
+        logger.debug("{} Waiting....".format(datetime.timestamp()))
+        if minutes % 5 == 0 and seconds >= 60:
+            logger.info("[{}] Waiting ({} minutes so far) for request to {}".format(datetime.timestamp(), minutes, url))
+        elif minutes % 1 == 0 and seconds >= 60:
```
anyoung-tableau, line 123: Was this supposed to be `minutes == 1`?

tableauserverclient/server/endpoint/datasources_endpoint.py
jacalata: oops, yea

tableauserverclient/server/endpoint/endpoint.py
jacalata: no, originally I was just logging every minute and then I added the longer message every 5 minutes and borked the checks

anyoung-tableau (approved):

PR #1219, opened 2023-04-13
feat: add JWTAuth

jorwoods:
Add `JWTAuth` class, as well as allowing Token auth to do impersonation.


jacalata (changes requested):
Could you just add a __repr__ function for the JWTAuth class? something like 
        return "<jwt token={} site={}>".format(self.jwt[:5]+ "...", self.site_id)

jorwoods:
Done. How do you feel about explicitly marking the `Credentials.__repr__` and `Credentials.credentials` as abstractmethods?

Additionally, I feel like `sign_in_with_personal_access_token` should be deprecated since it is just a redundant passthrough.

jorwoods:
Was there something missing from this PR that I should have implemented?

jacalata:
oops, I think this got auto-closed when I merged the dev branch

ngohuuduc:
Hello, 
Is there any update regarding to this matter ?

PR #1253, opened 2023-07-10
846 fix filter in operator spaces bug

LarsBreddemann:
This fix corrects the filter condition formatting in `server.filter.py` enabling searching for projects with spaces in the project name.


jacalata:
I'm not sure what it wants for the CLA - can you try clicking on that link "we need to confirm this"?

Also: can you write a test that covers this fix?

LarsBreddemann:
I've added test cases, fixed the formatting, and will submit a new PR.

PR #1265, opened 2023-08-01
pin black and mypy versions, update many dependencies

jacalata:
so that local runs always match CI runs


bcantoni (approved):

bcantoni:
One thing I noticed on this branch, running pytest locally I got one error:

    test/models/test_repr.py x

However, the Checks all passed so I'm not sure what this was.

bcantoni:
Ah that test that failed was an "xfail" and it's expected.

PR #1267, opened 2023-08-01
Update actions to newer versions to get supported Node versions

bcantoni:


bcantoni:
GitHub is automatically running on newer versions of Node, but it's a good idea to upgrade to the more recent Actions anyways:
https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

@jacalata if this passes all checks, let's get merged in to development with the others.

PR #1273, opened 2023-08-23
add support for custom schedules in TOL

a-torres-2:
Adds additional endpoint for creating an extract refresh task in TOL.

Also updates subscription item to contain schedule item if present (recent addition to TOL)

### DIFF
--- a/samples/create_extract_task.py
+++ b/samples/create_extract_task.py
@@ -1,0 +1,80 @@
+####
+# This script demonstrates how to create schedules using the Tableau
+# Server Client.
+#
+# To run the script, you must have installed Python 3.7 or later.
+####
+
+
+import argparse
+import logging
+
+from datetime import time
+
+import tableauserverclient as TSC
+
+
+def main():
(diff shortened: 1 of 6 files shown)


samples/create_extract_task.py
```diff
@@ -0,0 +1,80 @@
...
+        monthly_schedule = TSC.ScheduleItem(
+            None,
+            None,
+            None,
+            None,
+            monthly_interval,
+        )
+
+        target_item = TSC.Target(
+            "6582fdd0-1df8-43f5-9f18-dacc5238ec0a", # the id of the workbook or datasource
```
jacalata, line 57: I think generally the samples use workbooks.get(0) so they just work on any server. Is it possible to fetch the set of workbooks that can be extracted?

jacalata (changes requested):
Looks good once checks are fixed.

samples/create_extract_task.py
a-torres-2: sounds good 👍, updated accordingly

jacalata (approved):

PR #1289, opened 2023-10-04
Add docs for JWT and impersonation

jacalata:


bcantoni (approved):
LGTM with one comment: for the JWT section should we say something like "added in TSC 0.28"? Let's also get Lee and Dan's feedback.

akromish:
thanks for putting this together. just bumping as I'm sure other people have seen tableau jwt but assumed server-client has not yet implemented support even though you guys have.

Issue #1290, opened 2023-10-10
Problem when instantiating Tasks in the tableauserver when they have no Schedule assigned


danignraku:
**Describe the bug**
In our Tableau Site we don't use schedules for tasks, and when trying to fetch the issues from the endpoint, it fails.

**Versions**
Details of your environment, including:
 - Tableau Server Online
 - Python 3.9
 - TSC library version v0.28

**To Reproduce**
In a server without Schedules, try to get tasks with, for example:
`
req_option = tsc.RequestOptions(pagesize=1000)
all_extracts, pagination_item = server.tasks.get(req_option)
`

**Results**
It fails because there is no name for schedules.

**NOTE:** Be careful not to post user names, passwords, auth tokens or any other private or sensitive information.

The solution is to make the schedule optional for the tasks, as those can be created without schedules.

danignraku:
The solution is pretty easy. In task_item:
![image](https://github.com/tableau/server-client-python/assets/122803216/0c23c2cb-f3dd-4d4e-9840-4c9d2a7f3cef)
and 
![image](https://github.com/tableau/server-client-python/assets/122803216/bb7a609b-ae12-4d9b-ac50-5219d9ea33ed)

PR #1297, opened 2023-10-15
Jorwoods/tasks no schedule

jorwoods:
My thoughts on resolving #1290 and adding type annotations on tasks. I noticed at the moment, running a task returns the raw `bytes` from the server response. A change to the TSC API would be consuming that response and returning a `JobItem`.

### DIFF
--- a/tableauserverclient/models/task_item.py
+++ b/tableauserverclient/models/task_item.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from defusedxml.ElementTree import fromstring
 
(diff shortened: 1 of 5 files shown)


jorwoods:
Seems that the token for the test coverage has some missing permissions.
https://github.com/tableau/server-client-python/settings/actions

bcantoni:
@jacalata could you take a look at the issue here with generating test coverage: https://github.com/tableau/server-client-python/actions/runs/6527723147/job/17869783004?pr=1297

jacalata (approved):

Issue #1318, opened 2023-11-13
Wrong type annotation in workbook.refresh


pes-magic:
**Describe the bug**
`workbooks.refresh` is implemented to receive `workbook_id` as a string type argument, but it can actually receive `WorkbookItem` as well. The documentation also states that it receives `workbook_item`.

Implementation: https://github.com/tableau/server-client-python/blob/master/tableauserverclient/server/endpoint/workbooks_endpoint.py#L91
Documentation: https://tableau.github.io/server-client-python/docs/api-ref#workbooksrefresh

**Versions**
 - TSC 0.28

**To Reproduce**
Nothing to do

**Results**
Static code analysis displays false warnings

jorwoods:
@pes-magic Can you mark this as closed by #1319?

pes-magic:
Sure, thank you for reminding me

PR #1319, opened 2023-11-13
Fixed type annotation for workbook.refresh

pes-magic:
`workbook.refresh` is implemented to accept both `WorkbookItem` and `str` as arguments, but the type annotation describes it as receiving `str`, which can cause false warnings in static analysis.

Since the documentation states that it receives `workbook_item`, the name of the argument is also changed from `workbook_id` to `workbook_item`.

Issue: https://github.com/tableau/server-client-python/issues/1318


pes-magic:
I've signed the CLA

jacalata (approved):

PR #1328, opened 2023-12-13
support for multiple intervals for hourly, daily, and monthly schedules

a-torres-2:
Allow multiple intervals to be supported for schedules REST API.

Associated to this work item - https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001eHgiFYAS/view


jacalata (approved):
I think it looks good but there's a lot of logic.

PR #1343, opened 2024-01-17
Add Data Acceleration and Data Freshness Policy support

ltiffanydev:
Added support for Data Acceleration and Data Freshness Policy REST API. 
Mainly adding functionalities to workbook.update & workbook.get_by_id


bcantoni:
@ltiffanydev I closed & reopened this PR to force the CLA bot to run again and it looks happy now.

One change you'll need to make on this PR is target it for the `development` branch instead of `master`.

tableauserverclient/models/data_freshness_policy_item.py
```diff
@@ -0,0 +1,210 @@
...
+
+    # No intervals expected for Day frequency
+    if frequency == DataFreshnessPolicyItem.FreshAt.Frequency.Day:
+        return None
+
+    if frequency == DataFreshnessPolicyItem.FreshAt.Frequency.Week:
+        interval_values = [(i[1]).title() for i in interval]
+        return parse_week_intervals(interval_values)
+
+    if frequency == DataFreshnessPolicyItem.FreshAt.Frequency.Month:
```
jacalata, line 176: Can't this re-use the existing schedule parsing code?

tableauserverclient/models/view_item.py
```diff
@@ -186,4 +198,25 @@ def from_xml(cls, view_xml, ns, workbook_id="") -> "ViewItem":
             tags = TagItem.from_xml_element(tags_elem, ns)
             view_item.tags = tags
             view_item._initial_tags = copy.copy(tags)
+        if data_acceleration_config_elem is not None:
```
jacalata, line 201: should this check the API version before adding these elements? I assume earlier versions won't understand them.

jacalata (dismissed):

tableauserverclient/models/data_freshness_policy_item.py
ltiffanydev: I was considering that previously, but turns out there's some differences in how it's implemented - hence the simplest way for us is to have our own interval parser. 
E.g. in schedule, the parser returns MonthlyInterval/WeeklyInterval type +  has different member var, different structure than what we have.

samples/update_workbook_data_freshness_policy.py
```diff
@@ -0,0 +1,165 @@
...
+            sample_workbook.data_freshness_policy = TSC.DataFreshnessPolicyItem(
+                TSC.DataFreshnessPolicyItem.Option.SiteDefault)
+            updated: TSC.WorkbookItem = server.workbooks.update(sample_workbook)
+            print("Workbook " + updated.name + " updated data freshness policy option to: " +
+                  updated.data_freshness_policy.option)
+
+            # Update Workbook Data Freshness Policy to "FreshEvery" schedule.
+            # Set the schedule to be fresh every 10 hours
+            # Once the data_freshness_policy is already populated (e.g. due to previous calls),
+            # it is possible to directly change the option & other parameters directly like below
```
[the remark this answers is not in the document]
ltiffanydev: The default for any workbooks for data freshness policy is "SiteDefault", which will still refresh every 2 hours I believe.
The feature itself doesn't have no refreshing option.

samples/update_workbook_data_freshness_policy.py
```diff
@@ -0,0 +1,165 @@
...
+        print([workbook.name for workbook in all_workbooks])
+
+        if all_workbooks:
+            # Pick 1 workbook that has live datasource connection.
+            # Assuming 1st workbook met the criteria for sample purposes
+            # Data Freshness Policy is not available on extract & file-based datasource.
+            sample_workbook = all_workbooks[2]
+
+            # Get more info from the workbook selected
+            # Troubleshoot: if sample_workbook_extended.data_freshness_policy.option returns with AttributeError
```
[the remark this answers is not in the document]
ltiffanydev: hi @jacalata, do you mean to add try-catch in the getter for data_freshness_policy.option (data_freshness_policy_item.py ln 121)?
Or just in the sample file?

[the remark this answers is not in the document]
jacalata: just in the sample file, and put the comment as a printed out error message so the user definitely sees it.

tableauserverclient/models/view_item.py
ltiffanydev: what's the best way to check the API version inside models object?
Do I need to pass in Server object and call check_at_least_version?

ltiffanydev: hi @jacalata, pinging for advice how to do version check

ltiffanydev: After evaluating the changes, I think adding versioning check in here is unnecessary since this function from_xml is only for parsing the response of the API call.
If using older api versions, calling get view will not return any data_acceleration_elem on the response & the code will check if the element is not None.

jacalata (approved):

bcantoni:
@ltiffanydev reminder when you merge this to use the "squash and merge" option to bring all these commits into one. I think that might be the default anyways.

Also, you can ignore the failing License Compliance check. It's happening to all PRs on TSC at the moment.

samples/update_workbook_data_acceleration.py
```diff
@@ -0,0 +1,109 @@
...
+            sample_workbook = all_workbooks[2]
+
+            # Enable acceleration for all the views in the workbook
+            enable_config = dict()
+            enable_config["acceleration_enabled"] = True
+            enable_config["accelerate_now"] = True
+
+            sample_workbook.data_acceleration_config = enable_config
+            updated: TSC.WorkbookItem = server.workbooks.update(sample_workbook)
+            # Since we did not set
```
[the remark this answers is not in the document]
ltiffanydev: added the complete comment

samples/update_workbook_data_acceleration.py
```diff
@@ -0,0 +1,109 @@
...
+            print("Enable acceleration for all the views in the workbook " + updated.name + ".")
+
+            # Disable acceleration on one of the view in the workbook
+            # You have to populate_views first, then set the views of the workbook
+            # to the ones you want to update. Note that workbook.views expect List[ViewItem]
```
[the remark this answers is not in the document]
ltiffanydev: added type annotations for _views : Optional[List[ViewItem]] in workbook_item

PR #1346, opened 2024-01-19
CGI Removal

markm-io:
Removed cgi due to deprecation and updated to use [email.message.Message](https://docs.python.org/3/library/email.compat32-message.html#email.message.Message) instead.

### DIFF
--- a/tableauserverclient/server/endpoint/datasources_endpoint.py
+++ b/tableauserverclient/server/endpoint/datasources_endpoint.py
@@ -439,7 +439,7 @@ class Datasources(QuerysetEndpoint):
         with closing(self.get_request(url, parameters={"stream": True})) as server_response:
             m = Message()
             m["Content-Disposition"] = server_response.headers["Content-Disposition"]
-            params = m.get_filename()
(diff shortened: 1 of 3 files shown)


bcantoni:
@markm-io thanks for putting this together. There are 2 other places the CGI module is used. Would you mind updating `flows_endpoint.py` and `workbooks_endpoint.py` here as well?

Also, you'll need to fix those `black` and `mypy` errors before this PR will go green. You can see the steps explained more in the [dev guide](https://tableau.github.io/server-client-python/docs/dev-guide#code-and-commit).

markm-io:
> @markm-io thanks for putting this together. There are 2 other places the CGI module is used. Would you mind updating `flows_endpoint.py` and `workbooks_endpoint.py` here as well?
> 
> Also, you'll need to fix those `black` and `mypy` errors before this PR will go green. You can see the steps explained more in the [dev guide](https://tableau.github.io/server-client-python/docs/dev-guide#code-and-commit).

All files have been updates and both black and mypy commands have been run with corrections made.

jacalata (approved):

Issue #1324, opened 2023-11-23
Resolution parameters for PDF generation


adbumi:
## Summary
The Tableau REST API supports the resolution parameters **vizHeight** and **vizWidth** for PDF generation of views and workbooks, but the tableauserclient does not provide this capability.

## Suggested implementation
The implementation in tableauserclient seems to be quite simple. It is just a matter of adding these additional parameters to the PDFRequestOptions class and extending the get_query_params() method for the parameters.

I have inherited a child class from PDFRequestOptions and added the functionality. However, I would appreciate it if this functionality could be added to the tsc module as updates to the base class could otherwise break it. 

This is my workaround implementation: 

    `class PdfOptions(PDFRequestOptions):

        def __init__(self, page_type, orientation, viz_height, viz_width):
            super(PdfOptions, self).__init__(page_type, orientation)
            self.viz_height = viz_height
            self.viz_width = viz_width
  
       def get_query_params(self):
            params = super().get_query_params()
    
            if self.viz_height:
                params["vizHeight"] = self.viz_height
            if self.viz_width:
                params["vizWidth"] = self.viz_width
    
            self._append_view_filters(params)
            return params
`

bcantoni:
This should be available now that v0.31 was released today.

PR #1348, opened 2024-01-24
feat: allow viz height and width parameters

jorwoods:
Add support to PDFRequestOptions for viz height and width.

### DIFF
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -1,3 +1,5 @@
+import sys
+
 from tableauserverclient.models.property_decorators import property_is_int
 import logging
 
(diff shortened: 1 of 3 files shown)


tableauserverclient/server/request_options.py
```diff
@@ -287,6 +309,15 @@ def get_query_params(self):
         if self.max_age != -1:
             params["maxAge"] = self.max_age
 
+        if (self.viz_height is None) ^ (self.viz_width is None):
```
jacalata, line 312: I think this operator is unusual enough that a comment would be helpful

jacalata (approved):
Thanks!

tableauserverclient/server/request_options.py
jorwoods: Added comment.

jorwoods:
Resolves #1324

PR #1355, opened 2024-02-14
Add support for daily (24 hour) intervals for tasks

bcantoni:
Fixes #1354 by adding "24" as a valid interval. The model still doesn't do anything with the interval returned, but with this change you won't get an exception by just fetching tasks.

PR #1356, opened 2024-02-14
Update Actions to include Python 3.12 and reduce duplicate test runs

bcantoni:


bcantoni:
The Actions changes seem to have worked - in my last [PR](https://github.com/tableau/server-client-python/pull/1355/checks) I had 50 checks run; this time it was just 24.

jacalata (approved):
So the action now says "on every pull request, or on a push to main/dev"?

bcantoni:
@jacalata yes that's the intent - previously it was running on create/update pull request (correct) and also any commits pushed to a branch that's part of a pull request (not horrible, but ran everything twice). This syntax should do it for any PR as well as any pushes to master/development branches, so we benefit from the double-check there once a PR lands.

PR #1390, opened 2024-06-06
chore: add typing to Pager

jorwoods:
Add typing to the `Pager` object. The typing addition makes it clearer for language servers to detect what type of objects are yielded from the iterable, making TSC easier to use.

Additionally, this simplifies the way items are returned from the intermediate list object. Different from before, if the server count changes while the Pager is working, this change will continue to function. If desired, it can be reset to exiting gracefully should the server count change.

### DIFF
--- a/tableauserverclient/server/pager.py
+++ b/tableauserverclient/server/pager.py
@@ -1,9 +1,28 @@
+import copy
 from functools import partial
+from typing import Generic, Iterator, List, Optional, Protocol, Tuple, TypeVar, Union, runtime_checkable
 
-from . import RequestOptions
+from tableauserverclient.models.pagination_item import PaginationItem
+from tableauserverclient.server.request_options import RequestOptions
 
 
-class Pager(object):
+T = TypeVar("T")
+ReturnType = Tuple[List[T], PaginationItem]
+
+
+@runtime_checkable
+class Endpoint(Protocol):
+    def get(self, req_options: Optional[RequestOptions], **kwargs) -> ReturnType:
+        ...
+
+
+@runtime_checkable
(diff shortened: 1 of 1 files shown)


tableauserverclient/server/pager.py
```diff
@@ -12,60 +31,42 @@ class Pager(object):
...
-
-            except IndexError:
-                # The total count on Server changed while fetching exit gracefully
+    def __iter__(self) -> Iterator[T]:
+        options = copy.deepcopy(self._options)
+        while True:
+            # Fetch the first page
+            current_item_list, pagination_item = self._endpoint(options)
+
+            if pagination_item.total_available is None:
```
jacalata, line 60: Much cleaner! Is this just something you did while adding types because you're more familiar with iterators now?

jacalata (approved):

tableauserverclient/server/pager.py
jorwoods: Exactly. The only thing I'm unsure of with this PR is if the behavior of throwing an exception while iterating if the total count changes should be preserved. Theoretically there's no way of knowing that the item that was added/removed was already yielded by the iterator or hadn't been reached yet. My uncertainty comes from not know if that behavior is what we're trying to alert the library user of, or if it was an implementation detail about how state was being tracked previously.

jacalata: I'm pretty sure it was an implementation detail - on a busy site it's possible for e.g. the available workbooks or projects to be changing in the background very frequently, and the user will almost never care.

jacalata (approved):

PR #1391, opened 2024-06-06
Add workbook and datasource name attributes to JobItem Model

ivanB1975:
Added the workbook_name to the attributes of the jobItem Model. Added also the test.

### DIFF
--- a/tableauserverclient/models/job_item.py
+++ b/tableauserverclient/models/job_item.py
@@ -33,6 +33,7 @@ class JobItem(object):
         datasource_id: Optional[str] = None,
         flow_run: Optional[FlowRunItem] = None,
         updated_at: Optional[datetime.datetime] = None,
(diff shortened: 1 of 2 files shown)


ivanB1975:
Cla is signed

jorwoods:
@ivanB1975 I think if you're going to pull out workbook name, then you should pull out the datasource name as well.

ivanB1975:
> @ivanB1975 I think if you're going to pull out workbook name, then you should pull out the datasource name as well.

Thanks for having a look at it. Ok, I will need to check this

ivanB1975:
@jorwoods I added the datasource name as well

bcantoni:
I'll try close and reopen the PR to refresh the CLA bot status.

jacalata (approved):

PR #1394, opened 2024-06-09
chore: no implicit reexport

jorwoods:
Mypy has a setting to check for implicit reexports. In separate testing I have found that explicit exports, like what is used in this commit, make it easier for language servers to detect what is actually exported by the library which makes use by end users easier.

PEP8 specifies that "relative imports for intra-package imports are highly discouraged. Always use the absolute package path for all imports." This commit also makes imports absolute to comply with PEP8.


jacalata (approved):
love it

PR #1413, opened 2024-06-29
feat: group bulk edit

### LINKED ISSUE #1406: [TYPE 1] Add bulk removal for users from groups
## Summary
Please add the operation that is already supported by the API to bulk remove users from a group.

## Description
The API endpoint `PUT /api/api-version/sites/site-id/groups/group-id/users/remove` (see https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_users_and_groups.htm#remove_user_to_group ) allows to specify a list of users in the request body and all of them are removed from the group with only one API call. It would be nice to have this functionality in the Python library as well so we don't need to loop over the users one by one to remove them from a group.

jorwoods:
Close #1406 

Enables bulk adding and removing users to/from groups.

### DIFF
--- a/tableauserverclient/server/endpoint/groups_endpoint.py
+++ b/tableauserverclient/server/endpoint/groups_endpoint.py
@@ -1,17 +1,17 @@
 import logging
 
-from .endpoint import QuerysetEndpoint, api
-from .exceptions import MissingRequiredFieldError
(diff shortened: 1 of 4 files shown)


dtaskai:
Hi @jorwoods! Do you have an ETA on the merging of this PR? (and a new version release)

jacalata (approved):

PR #1422, opened 2024-07-22
feat: tag everything

### LINKED ISSUE #1421: [TYPE 1]Add add/delete tags on workbooks as a direct API call for permission reasons
Summary
Add add and delete tags on a workbook as a direct API call instead of indirectly updating a workbook because permissions may allow users to change tags but not update the workbook.

The current REST API calls of adding/deleting a tag on a workbook

Add
https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_workbooks_and_views.htm#add_tags_to_workbook

Delete
https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_workbooks_and_views.htm#delete_tag_from_workbook

Description
To update tags in a workbook, the current method is update the tags on a workbook then update the workbook: workbooks.update(workbook_item).  The main problem is the user must have update permissions on the workbook but changing the tags is a different permission. Currently users may have ability to change tags but not update the workbook.

The proposal is to either add new function calls like workbooks.add_tag(workbook_id, tag) and worbooks.delete_tag(workbook_id, tag) or handle it using a combination call but separate from update() like workbooks.update_tags(workbook_id, tags[]).

jorwoods:
Closes #1421


jorwoods:
Since workbooks, views, and data sources all follow the same pattern here, theoretically we could solve this with a Mixin class and add the functionality to all three with reduced duplication.

jorwoods:
Virtual connections also follow this pattern and could have the functionality added via a mixin class if that's the route we want to go.

jacalata:
yes, I think that's a great idea. I'd love to get all the shared functionality actually extracted from individual objects in this code.

jorwoods:
After #1429 gets merged, I can rebase again and add tagging support for virtual connections as well.

jacalata (approved):

PR #1435, opened 2024-08-06
docs: add docstrings detailing the filter options

jorwoods:
Add docstrings to `QuerySetEndpoint` subclasses so that details on what filter options are available can be displayed within the editor rather than navigating to the docs separately.


jacalata (approved):
Looks good. We should figure out a way to just publish the list of filters with each version.

jorwoods:
Yeah, the way I compiled this list was by scraping the [filtering and sorting docs](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_filtering_and_sorting.htm). May be difficult to update it with every server version.

PR #1444, opened 2024-08-21
v0.33 - Add GroupSets

jacalata:
Features:
- add name, datasource-name to Job item
- enable bulk add and remove users
- Linked Tasks: get, get by ID, run Now
- implement Tags: create new, add/delete for workbooks, flows, datasources
- get page and chunk size from env vars
- add some __repr__ implementations
- implement virtual connections

Bugfix: 
#1447 
#1449


jorwoods:
Please consider #1447 and #1449 for inclusion in this release. They fix minor errors in typing that end up affecting the IDE/LSP's ability to properly detect type information.

jacalata:
I should have been squashing all these merges  - I went back and re-did the virtual connections, I'll do the tagging change as well, and leave the others as is.

jorwoods:
And one more, #1454. Without it, if a user queries permissions on an object that has a groupSet element, the call will error.

bcantoni:
@jacalata @jorwoods let's coordinate on 0.33. I'm working on one fix for a customer issue.

bcantoni:
Okay I think we have all the PRs that @jorwoods and I were waiting for included now.

bcantoni (approved):

PR #1447, opened 2024-08-24
chore: typing decorators

jorwoods:
Decorators like `api`, `parameter_added_in`, and `_tsrequest_wrapped` altered the typing signatures and obfuscated mypy from inspecting any of the functions wrapped by these decorators. This PR aims to type these decorators, and fix any underlying typing issues revealed by fixing this.


jacalata (approved):
looks great

PR #1449, opened 2024-08-27
fix: Pager typing

jorwoods:
Pager Protocols were missing the generic flags. Added those in so the Pager correctly passes through the typing information. Also removes the kwargs from the function signature of the Endpoint.get protocol to make the Workbook endpoint match.

Adding a return annotation of `None` to the tests is very important because it is what enables static type checkers, like mypy, to inspect those functions. With these annotations now in place, users of TSC should more transparently be able to carry through typing information when using the Pager.

### DIFF
--- a/tableauserverclient/server/pager.py
+++ b/tableauserverclient/server/pager.py
@@ -1,24 +1,24 @@
+from collections.abc import Iterable, Iterator
 import copy
 from functools import partial
-from typing import Generic, Iterable, Iterator, List, Optional, Protocol, Tuple, TypeVar, Union, runtime_checkable
+from typing import List, Optional, Protocol, Tuple, TypeVar, Union, runtime_checkable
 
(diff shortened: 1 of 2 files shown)


jacalata:
Not sure I understand your comment about adding the return type to the tests. How does that make it easier for users to get typing information in their code?

jorwoods:
Making the protocols generic is what allows the pager to retrieve the typing information from the endpoint provided to Pager and surface that to users. 

Adding a return type on tests is what gives mypy the ability to inspect the test functions. Since test functions somewhat mirrors how people would use the package, contributors to TSC should annotate test functions returning None to verify that typing information is available to the package consumers.

jorwoods:
In this case, when I added typing to the Pager object in #1390, I missed annotating the functions as returning None. Since those test functions were not annotated, mypy falsely reported that there were no errors in typing.

jacalata (approved):

PR #1452, opened 2024-08-30
docs: clarify django style filters

jorwoods:
Add notes about supported endpoints and being able to use multiple filters at a time.


bcantoni (approved):
@jorwoods looks good - any. reason to hold on this? Looks like we could merge anytime.

jorwoods:
Should be good to go. Nothing in this is 0.33 specific.

PR #1453, opened 2024-08-30
docs: environment variables supported by v0.33

jorwoods:
Add a page to the documentation for environment variables added in #1439


bcantoni:
Looks good! We'll merge after 0.33 goes out.

PR #1454, opened 2024-09-03
feat: add as_reference method to groupset

jorwoods:
Missed adding this method when creating the rest of the GroupSetItem model.


jorwoods:
Being able to handle `groupSet` elements in permissions responses from Server was also missing. Added that as well.

jacalata (approved):

Issue #1456, opened 2024-09-04
FlowRuns endpoint does not return Pagination XML


jorwoods:
**Describe the bug**
A call to [Get Flow Runs](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_flow.htm#get_flow_runs) does not return pagination XML.

**Versions**
Details of your environment, including:
 - Tableau Server version (or note if using Tableau Online): 2022.3.1
 - Python version: 3.12.4
 - TSC library version: 0.32

**To Reproduce**
Steps to reproduce the behavior. Please include a code snippet where possible.
```python
import tableauserverclient as TSC
server = TSC.Server("https://url", True)
# auth
flow = server.flows.filter(name="some_flow")[0]
runs = server.flows.filter(flow_id=flow.id, page_size=10)[:]
```

**Results**
```
File "\.venv\Lib\site-packages\tableauserverclient\server\query.py", line 70, in __getitem__
    page_range = range((page - 1) * size, page * size)
                        ~~~~~^~~
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```
**NOTE:** Be careful not to post user names, passwords, auth tokens or any other private or sensitive information.

Dug into it more and server side, the "GET /api/api-version/sites/site-id/flows/runs" does not return a pagination element in the XML. I think being able to iterate over run history would be useful. Unclear if FlowRuns lacking pagination is intentional. If so, I can update TSC to reflect that.

jacalata:
I've pointed this out to the team that owns the API, but if they fix it it won't be available immediately so updating tsc is worth doing anyway.

jorwoods:
#1460 tweaks the QuerySet class in such a way that it should continue to function even if the server doesn't return pagination information. Adding it in server side would still be ideal to improve clarity and return information about what is truly available server-side.

jacalata:
closed this bug with release 
 https://github.com/tableau/server-client-python/releases/tag/v0.35
We will update the core REST API docs to mention the limitation on the API results.

PR #1460, opened 2024-09-12
fix: queryset support for flowruns

jorwoods:
FlowRun's get endpoint does not return a PaginationItem. This PR provides a tweak to QuerySet to enable a workaround so all items matching whatever filters are supplied.

It also corrects the return types of `flowruns.get` and fixes the XML test asset to reflect what is really returned by the server.

### DIFF
--- a/tableauserverclient/server/query.py
+++ b/tableauserverclient/server/query.py
@@ -44,7 +44,7 @@ class QuerySet(Iterable[T], Sized):
     QuerySet is an iterable, and can be used in for loops, list comprehensions,
     and other places where iterables are expected.
 
-    QuerySet is also a Sized, and can be used in places where the length of the
+    QuerySet is also Sized, and can be used in places where the length of the
     QuerySet is needed. The length of the QuerySet is the total number of items
     available in the QuerySet, not just the number of items that have been
     fetched. If the endpoint does not return a total count of items, the length
@@ -57,8 +57,8 @@ class QuerySet(Iterable[T], Sized):
     should create a new QuerySet for each place you need to use it, convert it
     to a list, or create a deep copy of the QuerySet.
 
-    QuerySet's are also indexable, and can be sliced. If you try to access an
-    item that has not been fetched, the QuerySet will fetch the page that
+    QuerySets are also indexable, and can be sliced. If you try to access an
+    index that has not been fetched, the QuerySet will fetch the page that
     contains the item you are looking for.
     """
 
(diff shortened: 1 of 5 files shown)


jorwoods:
A stop-gap for #1456 . Or a permanent fix if flowruns was never intended to return pagination XML.

jorwoods:
For endpoints like flowruns that do not return a `PaginationItem`, what should the len be? I set it to `sys.maxsize` in this PR because users may length check a QuerySet as part of a normal workflow. A len of 0 would be misleading, indicating to the user that there are no matches for the endpoint and/or filters they supplied. __len__ must return a non-negative int. Sentinel values such as -1 or None do not work. This only leaves maxsize as the possible flag.

Could theoretically be changed to raise an exception when length checking a QuerySet that doesn't actually have a `total_available` from the pagination XML.

tableauserverclient/server/query.py
```diff
@@ -35,6 +36,32 @@ def to_camel_case(word: str) -> str:
...
+    of the QuerySet will be sys.maxsize. If there is no total count, the
+    QuerySet will continue to fetch items until there are no more items to
+    fetch.
+
+    QuerySet is not re-entrant. It is not designed to be used in multiple places
+    at the same time. If you need to use a QuerySet in multiple places, you
+    should create a new QuerySet for each place you need to use it, convert it
+    to a list, or create a deep copy of the QuerySet.
+
+    QuerySet's are also indexable, and can be sliced. If you try to access an
```
jacalata, line 60: nit: no apostrophe

tableauserverclient/server/query.py
```diff
@@ -114,10 +151,15 @@ def _fetch_all(self: Self) -> None:
...
+            response = self.model.get(self.request_options)
+            if isinstance(response, tuple):
+                self._result_cache, self._pagination_item = response
+            else:
+                self._result_cache = response
+                self._pagination_item = PaginationItem()
 
     def __len__(self: Self) -> int:
-        return self.total_available
+        return self.total_available or sys.maxsize
```
jacalata, line 162: why do this instead of counting the existing page?

test/test_flowruns.py
```diff
@@ -13,6 +15,19 @@
 GET_BY_ID_INPROGRESS_XML = "flow_runs_get_by_id_inprogress.xml"
 
 
+def server_response_error_factory(code: str, summary: str, detail: str) -> str:
```
jacalata, line 18: This is nice, can you put it in a helper file so other tests can use it?

jacalata (approved):

tableauserverclient/server/query.py
jorwoods: Fixed.

test/test_flowruns.py
jorwoods: Moved it to _utils

PR #1461, opened 2024-09-13
chore(versions): Upgrade minimum python version

jorwoods:
As of October, 2024, Python 3.8 is out of support. Upgrading syntax to target Python 3.9. Adds builds for Python 3.13.


bcantoni:
@jorwoods @jacalata my initial thoughts:

1. I support the idea of moving the Python version forward like this.
2. I'm a little concerned about the volume of changes, especially with all those format strings. Was that a scripted change you made? (I.e. how confident are we nothing broke?)
3. By upgrading the syntax to target 3.9, will the library still work for 3.8 or will this be a breaking change?

jorwoods:
> I'm a little concerned about the volume of changes, especially with all those format strings. Was that a scripted change you made? (I.e. how confident are we nothing broke?)

This was a scripted change. In a year when the minimum version is bumped to 3.10, the volume of change should be lower. It is a lot of change, but I have noticed that some areas of the codebase use f-strings, some use `.format`, and some use the older python2 style `%s` string substitution. So while it is a lot of change, hopefully it brings more consistency in these small details across the codebase. Personally, I find the f-strings easier to read. f-strings are also faster than `.format` or `%s`, but not in a way that will be meaningful with the main purpose being around API calls.

> By upgrading the syntax to target 3.9, will the library still work for 3.8 or will this be a breaking change?

This would be a breaking change, but not because of the f-strings. The main syntax change I see that will cause it to not work for 3.8 would be the change from `Dict` to `dict`, `List` to `list`, `Set` to `set`, and `Tuple` to `tuple`. 

My comment about 3.8 being out of support is that the Python Software Foundation will stop releasing security fixes for 3.8 in [October 2024, as it is considered "End of Life."](https://devguide.python.org/versions/) Obviously we could continue to support older versions of python past their end of life, but I think it would hold back adopting newer syntax, features, and quality of life changes.

bcantoni:
Thanks Jordan that makes sense. I do prefer the f-strings as well and I'm in agreement on officially dropping 3.8 soon. I'm just trying to be careful for people still running on old versions.

Issue #1369, opened 2024-04-17
ValueError: invalid literal for int() with base 10: 'Second'


apparao84:
**Describe the bug**
When calling the IncrementalExtract refresh bridge refresh script as mentioned below We are getting below error. This worked successfully until last week.

```
Error:- ValueError: invalid literal for int() with base 10: 'Second'
```

**Versions**
Details of your environment, including:
Tableau Online
Python version: 3.8
TSC library version 3.22

**To Reproduce**
Steps to reproduce the behavior. Please include a code snippet where possible.

```
# Name of the data source in tableau
datasource = sys.argv[1]
 
tableau_auth = tsc.PersonalAccessTokenAuth(
    token_name=personal_access_token_name,
    personal_access_token=personal_access_token_secret,
    site_id=site
)
server = tsc.Server(tableau_server, use_server_version=True)
server.auth.sign_in(tableau_auth)
 
# Get datasource id from datasource name
req_option = tsc.RequestOptions(pagesize=1000)
req_option.filter.update([tsc.Filter(tsc.RequestOptions.Field.Name, tsc.RequestOptions.Operator.Equals, datasource)])
all_datasources = list(tsc.Pager(server.datasources, req_option))
 
datasource_id = all_datasources[0].id
 
# Get 1000 tasks and then filter for only extract tasks under the datasource id of interest
# I don't believe you can filter tasks
all_extracts = list(tsc.Pager(server.tasks, tsc.RequestOptions(pagesize=1000)))
 
# The task type for incremental extracts is 'IncrementExtractTask' and full extracts is 'extractRefresh'
# The task type for incremental exrracts via Bridges is 'IncrementExtractViaBridgeTask'
extracts = [i for i in all_extracts if (i.target is not None and i.target.id == datasource_id) and ('IncrementExtractViaBridgeTask' == i.task_type)]
 
incremental_extract_obj = extracts[0]
 
# Trigger the incremental extract associated with the datasource
update_xml = server.tasks.run(incremental_extract_obj)
job_id = xmltodict.parse(update_xml)['tsResponse']['job']['@id']
print(f"Refresh extract job {job_id} created")
```

**Results**
What are the results or error messages received?

```
C:\>D:\Python\python.exe C:\TableauPythonScripts\TableauExtractIncRefTrigScript.py Datasource_name
==================
Traceback (most recent call last):
  File "D:\Python\lib\site-packages\tableauserverclient\models\interval_item.py", line 183, in interval
[...]

Issue #1358, opened 2024-02-15
Schedules cannot handle valid monthly schedules


jacalata:
The schedules code only allows the string "LastDay". (https://github.com/tableau/server-client-python/blob/b06a7a249f7b0123f8b99fc43bd45539ad5f9eec/tableauserverclient/models/interval_item.py#L249C3-L253C33)

It should also allow the strings "First", "Second"..."Fifth".
https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_extract_and_encryption.htm#create_cloud_extract_refresh_task

While fixing this code, it would be ideal to reduce the duplication of elements e.g the array of valid hourly values is defined twice in the file.

bcantoni:
Another example of this is reported in #1369 which I closed as a duplicate.

bcantoni:
We're investigating a fix here (internal issue 16715720).

bcantoni:
Here are steps to reproduce the issue on Tableau Server.

1. Login to Tableau Server and head to the Schedules settings page. (You'll need to be a server admin to edit these.)
2. Create a new schedule, type Extract Refresh, under Frequency select Monthly, then under On change it to any value other than the default Day, then save the schedule
<img width="519" alt="image" src="https://github.com/user-attachments/assets/49f5bdea-a6d0-4a92-b13a-5be2e11c02f9">

3. Run the script below which will loop over all schedules and call get_by_id() on each one. This should trigger the ValueError exception.

```py
import tableauserverclient as TSC


tableau_auth = TSC.PersonalAccessTokenAuth(
    'PAT-NAME',
    'PAT-VALUE',
    'SITE-NAME-OR-BLANK'
    )
server = TSC.Server('SERVER-URL', use_server_version=True)
server.auth.sign_in(tableau_auth)

all_sch, pagination = server.schedules.get()
for sch in all_sch:
    sc = server.schedules.get_by_id(sch.id)
    print(sc.id)
```

You should get the error with exceptions like the following:

```
Traceback (most recent call last):
  File "/Users/brian/.pyenv/versions/tools-3.12/lib/python3.12/site-packages/tableauserverclient/models/interval_item.py", line 259, in interval
    if not (1 <= int(interval_value) <= 31):
                 ^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'Monday'

ValueError: Invalid interval value for a monthly frequency: Monday.
```

bcantoni:
[...]

PR #1462, opened 2024-09-13
Add support for all types of monthly repeating schedules

bcantoni:
This should fix handling all of the "monthly" schedule types now; see details in #1358 and #1369.

### DIFF
--- a/tableauserverclient/models/interval_item.py
+++ b/tableauserverclient/models/interval_item.py
@@ -247,7 +247,7 @@ class MonthlyInterval(object):
     @interval.setter
     def interval(self, interval_values):
         # Valid monthly intervals strings can contain any of the following
-        #   day numbers (1-31)
+        #   day numbers (1-31) (integer or string)
         #   relative day within the month (First, Second, ... Last)
         #   week days (Sunday, Monday, ... LastDay)
         VALID_INTERVALS = [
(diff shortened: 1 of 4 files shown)


bcantoni:
@jacalata and/or @jorwoods if you could help review that would be great.

tableauserverclient/models/interval_item.py
```diff
@@ -246,21 +246,35 @@ def interval(self):
...
-            if interval_value != "LastDay":
-                try:
-                    if not (1 <= int(interval_value) <= 31):
-                        raise ValueError(error)
-                except ValueError:
-                    if interval_value != "LastDay":
-                        raise ValueError(error)
+            try:
+                if not (1 <= int(interval_value) <= 31):
+                    error = f"Invalid monthly numeric frequency interval: {interval_value}."
```
anyoung-tableau, line 272: This error message can never happen right? Its ValueError will get caught below and either swallowed or a new ValueError will be thrown. I haven't used Python much, is it common to use exceptions to control flow?

Maybe use the `isinstance(interval, (int, float))` construct like on line 80 to detect if it's a number first.

bcantoni: Well in theory I think it would fail right there if the numerical value was 32 or "32".

I agree using exceptions here is a little funky. I was trying to fit in with the methods used already.

jorwoods: Using exceptions for control flow isn't uncommon in python. The `except ValueError` on 274 would indeed catch and consume the one raised on 273. The type check proposed by @anyoung-tableau does seem like a good alternative to let the different errors surface.

bcantoni: I'll take another look at it. The challenge is that even numbers are strings by the time they land here.

jacalata: Easiest way to do it seems like explicitly listing the known allowed digits: 
for value in range(1, 32):
    VALID_INTERVALS.add(str(value))

jacalata (approved):

bcantoni:
@jacalata I modified the validation and simplified by including the 1..31 range as both int or string. Looks better now and avoids that "never reachable" error path I started with.

jacalata (approved):
Looks good

PR #1465, opened 2024-09-18
chore(versions): Upgrade minimum python version

jorwoods:
(Redo of #1461, which got messed up when I was trying to rebase development)

As of October, 2024, Python 3.8 is ~out of support~ **end of life**. Upgrading syntax to target Python 3.9. Adds builds for Python 3.13.


jorwoods:
@bcantoni Apologies, the other PR got closed when I was trying to fix my branches. This one has the same changes successfully rebased on development. I have noticed that the tests appear to run considerably faster with this branch and the syntax upgraded.

jorwoods:
Copying comment from the original PR here:

> I'm a little concerned about the volume of changes, especially with all those format strings. Was that a scripted change you made? (I.e. how confident are we nothing broke?)

This was a scripted change. In a year when the minimum version is bumped to 3.10, the volume of change should be lower. It is a lot of change, but I have noticed that some areas of the codebase use f-strings, some use `.format`, and some use the older python2 style `%s` string substitution. So while it is a lot of change, hopefully it brings more consistency in these small details across the codebase. Personally, I find the f-strings easier to read. f-strings are also faster than `.format` or `%s`, but not in a way that will be meaningful with the main purpose being around API calls.

> By upgrading the syntax to target 3.9, will the library still work for 3.8 or will this be a breaking change?

This would be a breaking change, but not because of the f-strings. The main syntax change I see that will cause it to not work for 3.8 would be the change from `Dict` to `dict`, `List` to `list`, `Set` to `set`, and `Tuple` to `tuple`. 

My comment about 3.8 being out of support is that the Python Software Foundation will stop releasing security fixes for 3.8 in [October 2024, as it is considered "End of Life."](https://devguide.python.org/versions/) Obviously we could continue to support older versions of python past their end of life, but I think it would hold back adopting newer syntax, features, and quality of life changes.

tableauserverclient/_version.py
```diff
@@ -84,7 +84,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
                 stderr=(subprocess.PIPE if hide_stderr else None),
             )
             break
-        except EnvironmentError:
+        except OSError:
```
jorwoods, line 87: Non f-string change, but according to the [Built-in Exceptions docs](https://docs.python.org/3/library/exceptions.html#OSError.filename2) `EnvironmentError` was merged into `OSError` in 3.3.

test/test_regression_tests.py
```diff
@@ -3,7 +3,7 @@
 try:
     from unittest import mock
 except ImportError:
-    import mock  # type: ignore[no-redef]
+    from unittest import mock
```
jorwoods, line 6: Clean up this whole block

versioneer.py
```diff
@@ -328,7 +327,7 @@ def get_root():
         me_dir = os.path.normcase(os.path.splitext(me)[0])
         vsr_dir = os.path.normcase(os.path.splitext(versioneer_py)[0])
         if me_dir != vsr_dir:
-            print("Warning: build in %s is using versioneer.py from %s" % (os.path.dirname(me), versioneer_py))
+            print("Warning: build in {} is using versioneer.py from {}".format(os.path.dirname(me), versioneer_py))
```
jorwoods, line 330: f-string this

versioneer.py
```diff
@@ -398,7 +397,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
                 [c] + args, cwd=cwd, env=env, stdout=subprocess.PIPE, stderr=(subprocess.PIPE if hide_stderr else None)
             )
             break
-        except EnvironmentError:
+        except OSError:
```
jorwoods, line 400: Change of error from EnvironmentError to OSError. According to the [Built-in Exceptions docs](https://docs.python.org/3/library/exceptions.html#OSError.filename2), EnvironmentError was merged into OSError in 3.3

versioneer.py
```diff
@@ -408,7 +407,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
             return None, None
     else:
         if verbose:
-            print("unable to find command, tried %s" % (commands,))
+            print("unable to find command, tried {}".format(commands))
```
jorwoods, line 410: f-string this

versioneer.py
```diff
@@ -970,7 +969,7 @@ def git_get_keywords(versionfile_abs):
                 if mo:
                     keywords["date"] = mo.group(1)
         f.close()
-    except EnvironmentError:
+    except OSError:
```
jorwoods, line 972: Change of error from EnvironmentError to OSError. According to the [Built-in Exceptions docs](https://docs.python.org/3/library/exceptions.html#OSError.filename2), EnvironmentError was merged into OSError in 3.3

versioneer.py
```diff
@@ -1100,7 +1099,7 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, run_command=run_command):
             if verbose:
                 fmt = "tag '%s' doesn't start with prefix '%s'"
                 print(fmt % (full_tag, tag_prefix))
-            pieces["error"] = "tag '%s' doesn't start with prefix '%s'" % (full_tag, tag_prefix)
+            pieces["error"] = "tag '{}' doesn't start with prefix '{}'".format(full_tag, tag_prefix)
```
jorwoods, line 1102: f-string this

versioneer.py
```diff
@@ -1145,13 +1144,13 @@ def do_vcs_install(manifest_in, versionfile_source, ipy):
...
     try:
-        f = open(".gitattributes", "r")
+        f = open(".gitattributes")
         for line in f.readlines():
             if line.strip().startswith(versionfile_source):
                 if "export-subst" in line.strip().split()[1:]:
                     present = True
         f.close()
-    except EnvironmentError:
+    except OSError:
```
jorwoods, line 1153: Change of error from EnvironmentError to OSError. According to the [Built-in Exceptions docs](https://docs.python.org/3/library/exceptions.html#OSError.filename2), EnvironmentError was merged into OSError in 3.3

versioneer.py
```diff
@@ -1185,7 +1184,7 @@ def versions_from_parentdir(parentdir_prefix, root, verbose):
             root = os.path.dirname(root)  # up a level
 
     if verbose:
-        print("Tried directories %s but none started with prefix %s" % (str(rootdirs), parentdir_prefix))
+        print("Tried directories {} but none started with prefix {}".format(str(rootdirs), parentdir_prefix))
```
jorwoods, line 1187: f-string this

versioneer.py
```diff
@@ -1212,7 +1211,7 @@ def versions_from_file(filename):
     try:
         with open(filename) as f:
             contents = f.read()
-    except EnvironmentError:
+    except OSError:
```
jorwoods, line 1214: Change of error from EnvironmentError to OSError. According to the [Built-in Exceptions docs](https://docs.python.org/3/library/exceptions.html#OSError.filename2), EnvironmentError was merged into OSError in 3.3

versioneer.py
```diff
@@ -1229,7 +1228,7 @@ def write_to_version_file(filename, versions):
     with open(filename, "w") as f:
         f.write(SHORT_VERSION_PY % contents)
 
-    print("set %s to '%s'" % (filename, versions["version"]))
+    print("set {} to '{}'".format(filename, versions["version"]))
```
jorwoods, line 1231: f-string this

versioneer.py
```diff
@@ -1452,7 +1451,7 @@ def get_versions(verbose=False):
     try:
         ver = versions_from_file(versionfile_abs)
         if verbose:
-            print("got version from file %s %s" % (versionfile_abs, ver))
+            print("got version from file {} {}".format(versionfile_abs, ver))
```
jorwoods, line 1454: f-string this

versioneer.py
```diff
@@ -1723,7 +1722,7 @@ def do_setup():
     root = get_root()
     try:
         cfg = get_config_from_root(root)
-    except (EnvironmentError, configparser.NoSectionError, configparser.NoOptionError) as e:
+    except (OSError, configparser.NoSectionError, configparser.NoOptionError) as e:
```
jorwoods, line 1725: Change of error from EnvironmentError to OSError. According to the [Built-in Exceptions docs](https://docs.python.org/3/library/exceptions.html#OSError.filename2), EnvironmentError was merged into OSError in 3.3

versioneer.py
```diff
@@ -1748,9 +1747,9 @@ def do_setup():
     ipy = os.path.join(os.path.dirname(cfg.versionfile_source), "__init__.py")
     if os.path.exists(ipy):
         try:
-            with open(ipy, "r") as f:
+            with open(ipy) as f:
                 old = f.read()
-        except EnvironmentError:
+        except OSError:
```
jorwoods, line 1752: Change of error from EnvironmentError to OSError. According to the [Built-in Exceptions docs](https://docs.python.org/3/library/exceptions.html#OSError.filename2), EnvironmentError was merged into OSError in 3.3

versioneer.py
```diff
@@ -1769,12 +1768,12 @@ def do_setup():
...
     simple_includes = set()
     try:
-        with open(manifest_in, "r") as f:
+        with open(manifest_in) as f:
             for line in f:
                 if line.startswith("include "):
                     for include in line.split()[1:]:
                         simple_includes.add(include)
-    except EnvironmentError:
+    except OSError:
```
jorwoods, line 1776: Change of error from EnvironmentError to OSError. According to the [Built-in Exceptions docs](https://docs.python.org/3/library/exceptions.html#OSError.filename2), EnvironmentError was merged into OSError in 3.3

jorwoods (commented):
I reviewed all of the changes made by the script. The changes fall into a few main categories:

- Converting .format and %s to f-strings
- Removing unnecessary explicit inheritance from object
- Removing unnecessary explicit parameters being passed to `super`
- Switching to shorthand to define sets instead of creating a collection then passing to the `set` constructor
- Changing errors to OSError.

I tested the error handling locally, and it shouldn't have any impact to end users. The following reproducible example demonstrates these are interchangeable now.

```py
try:
   raise IOError
except OSError:
   print('caught IOError')
except Exception:
   print('whoops')
   
try:
   raise OSError
except IOError:
   print('caught OSError')
except Exception:
   print('whoops')

try:
   raise EnvironmentError
except OSError:
   print('caught EnvironmentError')
except Exception:
   print('whoops')
   
try:
   raise OSError
except EnvironmentError:
   print('caught OSError')
except Exception:
   print('whoops')
```

versioneer.py
jacalata: Are you reading for these, or do you have a tool finding them?

jorwoods: I read through all of the changes to find the ones the upgrade script missed.

jorwoods:
@jacalata I went through and fixed the f-strings the script missed. Should be good to go now.

Issue #1445, opened 2024-08-22
urllib3 specific version requirement causing dependencies conflict


jackexu:
**Describe the bug**

I was trying to upgrade the version from `0.25` to fix the `filename` issue per #1312 

However, I am not able to upgrade because the requirement of `urllib3==2.2.2` is causing dependencies conflict issue with other packages in my environment. Check the history and it seems starting 0.27 version the requirement became very specific.

Do we really need this specific version? Can we make it to `urllib3>=2` or exclude versions we don't want to use?  

# https://github.com/tableau/server-client-python/blob/257cf616f35cc36b24a73e9e102886a52ead1853/pyproject.toml#L18

**Versions**
Details of your environment, including:
 - Tableau Server version: 3.19
 - Python version: 3.9.13
 - TSC library version: 0.25 to 0.32

thesuperzapper:
@jacalata its extremely bad practice to pin a specific version of a one of the most common python libraries in the world.

The only reason why this doesn't make `tableauserverclient` completely un-installable is because no other package is crazy enough to specify a specific version.

Please use normal constraints like `>X.Y.Z,<3` where `X.Y.Z` is whatever the minimum required version is.

jackexu:
Agreed with @thesuperzapper 
If there are versions that have known issue then just exclude them - e.g. `>2, <3, !=X.Y.Z`

PR #1468, opened 2024-09-18
Draft: Make urllib3 dependency more flexible

bcantoni:
Per discussion in #1445


jacalata:
This was set by dependabot, we'll have to keep an eye on that to see what it tries to do next time.

jacalata (approved):

Issue #1450, opened 2024-08-28
Fix unhelpful error message when connecting to server fails: 'Invalid Version: Unknown'


genehynson:
**Describe the bug**

When initializing the Tableau Server client it's possible to encounter the error message `Invalid Version: Unknown` which is rather confusing and does not provide enough context to debug the real issue. This occurs whenever the Tableau server name provided does not resolve to a Tableau Server (e.g. perhaps because of an SSO wall). However, the error message incorrectly implies that there is some sort of issue during the negotiation of API versions, which is very confusing.

Our usage:
```
import tableauserverclient as tsc
...
self._server = tsc.Server(server_name)
self._server.add_http_options({"verify": verify_ssl})
self._server.use_server_version()
self._server.auth.sign_in(auth)
```

This error is raised because:
- `use_server_version()` invokes `_determine_highest_version()` which [invokes](https://github.com/tableau/server-client-python/blob/257cf616f35cc36b24a73e9e102886a52ead1853/tableauserverclient/server/server.py#L169) `self.server_info.get().rest_api_version`
- `ServerInfo.get()` makes an [unauthenticated request](https://github.com/tableau/server-client-python/blob/257cf616f35cc36b24a73e9e102886a52ead1853/tableauserverclient/server/endpoint/server_info_endpoint.py#L34) to the provided url to attempt to fetch the server version info. The response is received from the server but it is not the payload expected by the client (it's likely an html webpage) but no error is raised.
- The bad response payload is then [parsed by](https://github.com/tableau/server-client-python/blob/257cf616f35cc36b24a73e9e102886a52ead1853/tableauserverclient/models/server_info_item.py#L41) `ServerInfoItem.from_response()` where an error is raised.
- The error is caught and silenced, with [response](https://github.com/tableau/server-client-python/blob/257cf616f35cc36b24a73e9e102886a52ead1853/tableauserverclient/models/server_info_item.py#L45) `cls("Unknown", "Unknown", "Unknown")` returned.
- This results in `Server.version` to be set to "Unknown" [here](https://github.com/tableau/server-client-python/blob/257cf616f35cc36b24a73e9e102886a52ead1853/tableauserverclient/server/server.py#L183).
[...]

PR #1478, opened 2024-09-28
feat(exceptions): separate failed signin error

### LINKED ISSUE #1472: Can we convert this in to ServerResponseError.from_response exception instead of NonXMLResponseError
It would be helpful if ServerResponseError.from_response is implemented on line 173 instead NonXMLResponseError.

https://github.com/tableau/server-client-python/blob/4259316ef2e2656531b0c65c71d043708b37b4a9/tableauserverclient/server/endpoint/endpoint.py#L173

jorwoods:
Closes #1472

This makes sign in failures their own class of exceptions, while still inheriting from NotSignedInException to not break backwards compatability for any existing client code. This should allow users to get out more specific exceptions more easily on what failed with their authentication request.


bcantoni:
@jorwoods do you think we should/could incorporate a change here for the minor issue #1450? I don't know whether that other issue is big enough to fix, but since your change here is in a related area maybe we could do together.

jorwoods:
@bcantoni I made a tweak to how the ServerInfoItem does its error handling to raise the error immediately. It currently raises a `NonXMLResponseError`. Not sure if you would want something custom raised there to indicate that the problem is more extensive beyond it not being valid XML.

jorwoods:
I also changed the log level on those messages from info to exception so that they'll bubble up more clearly in logs.

bcantoni:
Did some quick testing and this change looks good.

Sign in with incorrect PAT, before:
```
tableauserverclient.server.endpoint.exceptions.NotSignedInError: (b'<?xml version=\'1.0\' encoding=\'UTF-8\'?><tsResponse xmlns="http://tableau.com/api" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://tableau.com/api https://help.tableau.com/samples/en-us/rest_api/ts-api_3_25.xsd"><error code="401001"><summary>Signin Error</summary><detail>The personal access token you provided is invalid.</detail></error></tsResponse>', 'https://MYSERVER/api/3.25/auth/signin')
```

Sign in with incorrect PAT, after:
```
tableauserverclient.server.endpoint.exceptions.FailedSignInError: Failed Sign In Error:

	401001: Signin Error
		The personal access token you provided is invalid.
```

bcantoni (approved):
On second thought I don't think we need to do anything specific for 1450 (no specific error for pointing to a non-Tableau server).

jorwoods:
Do you want it to raise the NonXMLResponseError? Or should I revert that change?

PR #1479, opened 2024-09-28
Update samples for Python 3.x compatibility

### LINKED ISSUE #1469: Modernize samples a bit to ensure they work with Python 3.x
* Run 2to3 to see what changes are proposed and whether they are needed/worthwhile
* Replace obsolete `env` package
* Update "samples list" in docs page to mention every sample that exists

bcantoni:
These are the only functional issues I've found so far. There are other style/idiom changes but those aren't critical.

This is a draft PR while I do some testing.

Fixes #1469


bcantoni:
@jacalata could you please review? Then I will squash and merge.

samples/login.py
```diff
@@ -80,7 +86,7 @@ def sample_connect_to_server(args):
     server = TSC.Server(args.server, use_server_version=True, http_options={"verify": check_ssl_certificate})
     server.auth.sign_in(tableau_auth)
     server.version = "2.6"
-    new_site: TSC.SiteItem = TSC.SiteItem("cdnear", content_url=env.site)
```
jacalata, line 83: is there a better hardcoded name we can use?

jacalata (approved):

samples/login.py
bcantoni: @jacalata good catch! Looking at it more closely what if we just remove this switchsite chunk to make the demo more clear?

```py
    server.version = "2.6"
    new_site: TSC.SiteItem = TSC.SiteItem("cdnear", content_url=args.site)
    server.auth.switch_site(new_site)
```

jacalata: yea we probably don't need that in an example called login

bcantoni: Done!

PR #1484, opened 2024-10-02
docs: add docstrings to auth objects and endpoints

jorwoods:


tableauserverclient/server/endpoint/auth_endpoint.py
```diff
@@ -70,14 +94,17 @@ def sign_in(self, auth_req: "Credentials") -> contextmgr:
     # The distinct methods are mostly useful for explicitly showing api version support for each auth type
     @api(version="3.6")
     def sign_in_with_personal_access_token(self, auth_req: "Credentials") -> contextmgr:
+        """Passthrough to sign_in method"""
         return self.sign_in(auth_req)
 
     @api(version="3.17")
     def sign_in_with_json_web_token(self, auth_req: "Credentials") -> contextmgr:
+        """Passthrough to sign_in method"""
         return self.sign_in(auth_req)
```
jorwoods, line 103: Since these are pass through methods, should they be deprecated?

jacalata: I think it's worth keeping them, because signing in is something people are working on when they don't necessarily understand the library or API. It just adds a bit of scaffolding for them.

jacalata (approved):

Issue #1451, opened 2024-08-29
[Discussion] Docs rewrite


jorwoods:
Currently, documentation is maintained on a separate branch, gh-pages, and written in Markdown. With all of the type hints that have been incorporated over the last few releases, IDE/LSP's will have better information about what function signatures are available and being used. Python's default way of making documentation available to users is through docstrings.

What I propose is that instead of separately maintaining the documentation, especially the user facing TSC API documentation, separately, that it be maintained through docstrings in the .py files. Then the markdown that is currently being used is converted to reStructuredText. This would open up the ability to use [sphinx](https://www.sphinx-doc.org/en/master/) and its [autodoc extension](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) to read those docstrings and produce the web navigable documentation. I believe the docs could still be published to gh-pages, so URLs would not have to change.

.rst is definitely not as simple of a format as .md, but the ability to fetch docstrings and function signatures I think offers a clear advantage over continuing with markdown.

jacalata:
@dzucker-tab, thoughts?

Our tooling would be constrained by having to work with our internal doc writing setup, which I'm not perfectly familiar with.

bcantoni:
In the past we've talked about this idea internally as well. One big benefit in my mind is we might have better traction keeping the docs up to date with each PR, whether it's automated docstrings or other doc content that can all come together in one PR.

jacalata:
#902

jorwoods:
I had missed the prior conversation about this. Concern over docs updates being tied to package updates is a pretty big one. I know that I have found myself missing adding changes to docs and then doing smaller PRs to fix it later. I guess the question is if having them available in docstrings and in code PRs means the docs revisions are likely to be less frequent?

jacalata:
Docs updates would have to go to /main, but we don't need to bump the package every time we update main. We'd hope not to have a stream of doc updates, but if we need to make some it won't be a real problem.
Dan says 👍 on the branch change and probably doc-strings.

jorwoods:
Ok. I'll progressively submit PRs and some conversions to the docs. I think it will take quite a while before the gh-pages branch can be really replaced.

jacalata:
Just linking this to the work in progress: PR #1832 sets up a Sphinx + ReadTheDocs pipeline that generates the API reference directly from docstrings, which is the approach this issue proposed. Once that lands and the gh-pages content is migrated over, this can close.

PR #1485, opened 2024-10-03
docs: docstrings for user item and endpoint

jorwoods:


jacalata (approved):
This is great. Are you doing it manually? We do have most of this in the REST docs.

jorwoods:
Its a mix of copying information in from https://tableau.github.io/server-client-python/docs/api-ref, https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm, and adding some extra detail where appropriate. The main thing I'm unsure of is detailing all of the ServerResponseError codes and details or if it should just point them to the server REST API docs for that detail.

jorwoods:
The reason for pulling this info into docstrings from the docs is if we go forward with the rewrite per #1451, then we could pull the docstrings in as the documentation. Making it available both in the web and in the IDE.

jacalata:
I think it's probably easier to point to the existing docs for error codes. Some of them have a pretty long list of possible return codes

PR #1486, opened 2024-10-04
Remove sample code showing group name encoding

jacalata:
This is no longer needed - ran the sample and verified that it works now.

However I did see a weird issue with the last few lines, which I've commented out for now, and it seems the code needs to be updated from sort to order_by? I want to spend more time looking at that in another change.


jorwoods:
I tested out the bug you identified here. The infinite loop happens when no matching results are found. I will get a fix created shortly.

renoyjohnm (approved):

PR #1488, opened 2024-10-07
docs: workbook docstrings

jorwoods:
Add detailed docstrings to workbook item and endpoint

### DIFF
--- a/tableauserverclient/server/endpoint/workbooks_endpoint.py
+++ b/tableauserverclient/server/endpoint/workbooks_endpoint.py
@@ -75,10 +75,11 @@ class Workbooks(QuerysetEndpoint[WorkbookItem], TaggingMixin[WorkbookItem]):
 
         Parameters
         ----------
-        (Optional) You can pass the method a request object that contains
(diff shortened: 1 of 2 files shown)


tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -141,6 +289,49 @@ def update(
...
+        Returns
+        -------
+        WorkbookItem
+            The updated workbook item.
+
+        Examples
+        --------
+
+        >>> import tableauserverclient as TSC
+        >>> tableau_auth = TSC.TableauAuth('username', 'password', site_id='site')
```
jacalata, line 319: I'm not sure about having all this sample code in the doc strings, it'll be easy to let it get out of date. I'm hoping to go the other direction and make it possible to run all the sample files regularly - perhaps significant code snippets could be put in a sample, and the comment could point there?

jorwoods: I can remove the examples.

jacalata (approved):

PR #1494, opened 2024-10-11
docs: docstrings for Server and ServerInfo

jorwoods:


jacalata (approved):
That enum looks out of place, but I can look at that another time

PR #1495, opened 2024-10-11
docs: docstrings for site item and endpoint

jorwoods:


tableauserverclient/server/endpoint/sites_endpoint.py
```diff
@@ -137,6 +362,25 @@ def encrypt_extracts(self, site_id: str) -> None:
 
     @api(version="3.5")
     def decrypt_extracts(self, site_id: str) -> None:
+        """
+        Decrypts all extracts on the site.
+
+        REST API: https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_extract_and_encryption.htm#decrypt_extracts
```
jacalata, line 368: These three encrypt methods are all Server Admin/Tableau Server only.  
I noticed that deleteSite/createSite each say this differently - I think we want this line to focus on Tableau Cloud v Server, not the role required, since we aren't giving any other role limits that I noticed, so I think they should all just say "Tableau Server only." Any thoughts?

jacalata (approved):

PR #1533, opened 2024-11-17
Merge gh-pages into development

jorwoods:
Required its own branch and merge due to unrelated git histories


jorwoods:
@jacalata @bcantoni 

Any thoughts as to what needs to happen with the CLAs here? All of these contributions were already in the gh-pages branch, so should have had a CLA in place, right?

jacalata:
We still should do this but at this point I think it needs to be restarted.

jacalata:
So the main blocker here is conflict between our doc writers standard workflow, and the desire to make it easier for people writing code to write docs alongside it. I think I've found a solution: we keep the gh-pages branch that uses Jekyll etc, for the tutorials, concept explanations, etc, which the docs team can edit without worrying about releases etc in the master branch, and we add a Sphinx action in the development/master branch that collects all the docstrings and serves them as html that can be embedded in the existing docs as the API reference section. 

I want to close this PR, but will check if there's any work in it that we want to save.

jacalata:
Closing in favour of Sphinx-based autodoc documentation approach. The Jekyll site will remain on gh-pages; API reference will be served via Read the Docs.

PR #1535, opened 2024-11-21
Fixing set default permissions for virtual connections

### LINKED ISSUE #1525: Bug - Virtual Connection default permissions
**Describe the bug**
In #1463 I've added support for virtual connection project permissions. I've now realized that they don't work.

Default virtualconnection permissions have been defined in project_item as follows:

```python
self._default_virtualconnection_permissions = None
......
@property
def default_virtualconnection_permissions(self):
    if self._default_virtualconnection_permissions is None:
        raise UnpopulatedPropertyError(self.ERROR_MSG)
    return self._default_virtualconnection_permissions()
```

While populating permissions, we use Resource.VirtualConection  referencing models/tableau_types 
```python
@api(version="3.23")
def populate_virtualconnection_default_permissions(self, item: ProjectItem) -> None:
    self._default_permissions.populate_default_permissions(item, Resource.VirtualConnection)
```
```python
class Resource:
    ...
    VirtualConnection = "virtualConnection"
```

Issue occurs while populating the permissions, as `_set_default_permissions` function tries to set the attribute `_default_virtualConnection_permissions` instead of `_default_virtualconnection_permissions` 

```python
def _set_default_permissions(self, permissions, content_type):
    attr = f"_default_{content_type}_permissions"
    setattr(
        self,
        attr,
        permissions,
    )
```

Possible solutions: 

1. Change `virtualConnection` definition in models/tableau_types.py to `virtualconnection` _- could this be a breaking change for sth?_
2. Change `default_virtualconnection_permissions` to `default_virtualConnection_permissions` _- snake case / camel case hybrid?  not a good look_
3. Change `attr = f"_default_{content_type}_permissions"` so it uses only lowercase

I've created the issue instead of directly raising a PR as I am not sure which one of these changes would be prefered

TrimPeachu:
Fixes #1525

### DIFF
--- a/tableauserverclient/models/project_item.py
+++ b/tableauserverclient/models/project_item.py
@@ -174,7 +174,7 @@ class ProjectItem:
         self._permissions = permissions
 
     def _set_default_permissions(self, permissions, content_type):
-        attr = f"_default_{content_type}_permissions"
+        attr = f"_default_{content_type}_permissions".lower()
         setattr(
             self,
             attr,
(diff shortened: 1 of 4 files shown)


TrimPeachu:
@jorwoods  tests added

jacalata (approved):
Looks good afaict, although I'm not actually familiar with the feature.

PR #1542, opened 2024-12-04
Adding support for thumbnail related options in workbook publish

renoyjohnm:
Adding support for thumbnail related options, `thumbnailsUserId` & `thumbnailsGroupId` in workbook publish

### DIFF
--- a/samples/publish_workbook.py
+++ b/samples/publish_workbook.py
@@ -36,9 +36,16 @@ def main():
         help="desired logging level (set to error by default)",
     )
     # Options specific to this sample
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument("--thumbnails-user-id", "-u", help="User ID to use for thumbnails")
+    group.add_argument("--thumbnails-group-id", "-g", help="Group ID to use for thumbnails")
+
+    parser.add_argument("--workbook-name", "-n", help="Name with which to publish the workbook")
     parser.add_argument("--file", "-f", help="local filepath of the workbook to publish")
     parser.add_argument("--as-job", "-a", help="Publishing asynchronously", action="store_true")
     parser.add_argument("--skip-connection-check", "-c", help="Skip live connection check", action="store_true")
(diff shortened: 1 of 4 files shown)


tableauserverclient/server/request_factory.py
```diff
@@ -958,6 +958,13 @@ def _generate_xml(
             views_element = ET.SubElement(workbook_element, "views")
             for view_name in workbook_item.hidden_views:
                 _add_hiddenview_element(views_element, view_name)
+
+        if workbook_item.thumbnails_user_id is not None:
+            workbook_element.attrib["thumbnailsUserId"] = workbook_item.thumbnails_user_id
+
+        if workbook_item.thumbnails_group_id is not None:
```
jacalata, line 965: only one of them should be set, so maybe make this an else if?

samples/publish_workbook.py
```diff
@@ -36,9 +36,16 @@ def main():
...
     # Options specific to this sample
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument("--thumbnails-user-id", "-u", help="User ID to use for thumbnails")
+    group.add_argument("--thumbnails-group-id", "-g", help="Group ID to use for thumbnails")
+
+    parser.add_argument("--workbook-name", "-n", help="Name with which to publish the workbook")
     parser.add_argument("--file", "-f", help="local filepath of the workbook to publish")
     parser.add_argument("--as-job", "-a", help="Publishing asynchronously", action="store_true")
     parser.add_argument("--skip-connection-check", "-c", help="Skip live connection check", action="store_true")
+    parser.add_argument("--project", help="Project within which to publish the datasource")
```
jacalata, line 47: nit: datasource -> workbook

jacalata (approved):

tableauserverclient/server/request_factory.py
renoyjohnm: sounds good, will change it to elif

jacalata (approved):

PR #1557, opened 2025-01-13
feat: add PageType.Unspecified to docs

gabriel-leal-de-almeida:
Hello! I was checking the class and there is the "Unspecified" options that is not mentioned in the docs and is just what I needed to use. I am proposing this change so it can also help others.

Reference in the codebase: https://github.com/tableau/server-client-python/blob/master/tableauserverclient/server/request_options.py#L421


jacalata (approved):
just need the CLA

bcantoni:
Merged - thanks @gabriel-leal-de-almeida for the change!

PR #1563, opened 2025-02-06
feat: Add fields:_all_ support

jorwoods:
Add support for all fields as specified in [Using fields in the REST API](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_fields.htm).


tableauserverclient/models/datasource_item.py
```diff
@@ -471,4 +571,35 @@ def _parse_element(datasource_xml: ET.Element, ns: dict) -> tuple:
...
+            favorites_total,
+            has_alert,
+            is_published,
+            server_name,
+            project,
+            owner,
         )
+
+
+@overload
```
jacalata, line 584: Do we not have these in a shared location?

tableauserverclient/models/project_item.py
```diff
@@ -87,6 +109,13 @@ def __init__(
         self._default_database_permissions = None
         self._default_table_permissions = None
 
+        self._project_count: Optional[int] = None
+        self._workbok_count: Optional[int] = None
```
jacalata, line 113: Typo

tableauserverclient/server/request_options.py
```diff
@@ -62,8 +62,9 @@ def __init__(self, pagenumber=1, pagesize=None):
         self.pagesize = pagesize or config.PAGE_SIZE
         self.sort = set()
         self.filter = set()
+        self.fields = set()
         # This is private until we expand all of our parsers to handle the extra fields
-        self._all_fields = False
```
jacalata, line 66: I would guess people have used this field even though it's private, no? Can we support both for a release?

jacalata (changes requested):

jacalata (approved):
👍

PR #1593, opened 2025-04-20
chore: type hint database and table objects

jorwoods:


tableauserverclient/server/endpoint/databases_endpoint.py
```diff
@@ -88,43 +169,45 @@ def _get_tables_for_database(self, database_item):
...
         return self._permissions.update(item, rules)
 
     @api(version="3.5")
-    def delete_permission(self, item, rules):
+    def delete_permission(self, item: DatabaseItem, rules: list[PermissionsRule]) -> None:
         self._permissions.delete(item, rules)
 
     @api(version="3.5")
-    def populate_table_default_permissions(self, item):
+    def populate_table_default_permissions(self, item: DatabaseItem):
```
jacalata, line 184: add -> None

jacalata: Actually I see there's still a couple methods needing more work, so this can be left for later.

jacalata (approved):

PR #1594, opened 2025-04-20
feat: enable retrieving only owned workbooks

jorwoods:


bcantoni (approved):
@jorwoods this looks good to me. That `ownedBy` parameter seems to have been there for a while; or do you think it's relatively new?

jorwoods:
I think it has been supported server side for a while, but missing from tsc. A conversation I had at TC25 made me realize that it was somehow missed.

PR #1595, opened 2025-04-20
ci: update slack action vm version

jorwoods:
Ubuntu 20.04 went out of support as of 2025-04-15. Updating to use ~24.04~ latest.


bcantoni:
@jorwoods what would you think about using `ubuntu-latest` here like we have in other actions? Assuming it works that would future proof us a bit :)

jorwoods:
@bcantoni good idea. Updated.

bcantoni (approved):
Great thank you - I'll get this in.

PR #1596, opened 2025-04-22
Add SSL option for connecting to Tableau Server with a weaker DH key length

### LINKED ISSUE #1582: Versions of Python greater than 3.10 require a larger Diffie-Hellman (DH) key than what Tableau Server uses
Hi, folks.

Here's the bug description:
Current versions of Python (3.12 and above) don't like Tableau Server's (e.g., version 2024.2) Diffie-Hellman key. This causes problems with Tableau Server Client Python, regardless of the TSC version. It's newer Python requiring a stronger DH key than what Tableau Server provides.

Here's more details and a suggestion for Tableau to increase the DH key in Tableau Server: https://ideas.salesforce.com/s/idea/a0BHp000016Klv0MAC/tableau-should-increase-the-size-of-its-diffiehellman-dh-key-exchange. 

Here's the environment information: 
Python Info:
        Python Version:  3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)]
        Tableau Server Client Version:  0.17.0
Tableau Info:
        Tableau Server Version: 2024.2.1
        Tableau Server Build: 20242.24.0719.1101
        REST API Version: 3.23

Here's how to reproduce this: 
1. Use Python version greater than 3.10. I don't get the error when I use Python 3.10.5. I do get the error when I use Python version 3.12.  Have Python 3.12 or greater authenticate into Tableau Server's REST API, and Python generates an error with the SSL handshake. 

Here is the error message:  "in do_handshake self._sslobj.do_handshake() ssl.SSLError: [SSL: DH_KEY_TOO_SMALL] dh key too small (_ssl.c:1000). Tableau Sever is not secure enough for the SSL connection with Python." 

Resolution
This foremost should be resolved in Tableau Server by increasing the size of the DH key there. 

In the meantime, I'm wondering if TSC has a preferred work-around to add to future versions of TSC. For example, lowering the default SSL security level in Python if an initial SSL handshake fails. I'm interested in feedback from Tableau and TSC developers on how concerned they are about this issue and if it's something that needs to be made more secure. 

Thank you,
Joe

bcantoni:
Fixes #1582

### DIFF
--- a/tableauserverclient/server/server.py
+++ b/tableauserverclient/server/server.py
@@ -2,6 +2,7 @@ from tableauserverclient.helpers.logging import logger
 
 import requests
 import urllib3
+import ssl
 
 from defusedxml.ElementTree import fromstring, ParseError
 from packaging.version import Version
@@ -91,6 +92,13 @@ class Server:
         and a later version of the REST API. For more information, see REST API
         Versions.
 
+    http_options : dict, optional
+        Additional options to pass to the requests library when making HTTP requests.
+
+    session_factory : callable, optional
+        A factory function that returns a requests.Session object. If not provided,
+        requests.session is used.
+
     Examples
     --------
     >>> import tableauserverclient as TSC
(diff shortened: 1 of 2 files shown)


jacalata (approved):
lgtm

PR #1601, opened 2025-04-27
Jorwoods/idp

jorwoods:
Add support for multiple IDP configurations, retrieving the IDPs from the Tableau Cloud, and specifying the IDP when adding or updating a user.


tableauserverclient/models/user_item.py
```diff
@@ -184,6 +186,18 @@ def groups(self) -> "Pager":
...
+        """
+        IDP configuration id for the user. This is only available on Tableau
+        Cloud, 3.24 or later
+        """
+        return self._idp_configuration_id
+
+    @idp_configuration_id.setter
+    def idp_configuration_id(self, value: str) -> None:
+        self._idp_configuration_id = value
+
```
jorwoods, line 200: While this is currently mutually exclusive with `TSC.UserItem.auth_setting`, @sf-stikone and I discussed leaving that out of TSC, instead leaving validation to occur server-side in case that should change in the future.

sf-stikone (approved):

tableauserverclient/models/user_item.py
```diff
@@ -238,6 +254,8 @@ def _set_values(
             self._auth_setting = auth_setting
         if domain_name:
             self._domain_name = domain_name
+        if idp_configuration_id:
```
jacalata, line 257: Do we have a doc comment mentioning that idp_configuration_id will override the domain name if both are present?

jacalata (approved):

tableauserverclient/models/user_item.py
jorwoods: @jacalata Good catch. That was a typo, and I removed it.

PR #1603, opened 2025-05-06
Add the VDS endpoint into the TSC

heshaojian:


bcantoni:
FYI this PR is on hold while we internally sort out where the VDS support will be exposed. I'll leave the PR up for now.

bcantoni:
[VizQL Data Service (VDS) just released its own Python library](https://github.com/tableau/VizQL-Data-Service/tree/main/python_sdk), so we can close this PR as not needed.

PR #1614, opened 2025-05-16
Centralise test file references

jacalata:
First I tried moving test assets into folders, then I realized that I had to update file references individually in all the test files. So I went back to do all the file path work in a utility and call that, for easier changing.


jorwoods:
I've thought about similar changes to the tests. I think switching to pathlib.Path objects for the test assets also simplifies things.

test/test_custom_view.py
```diff
@@ -12,17 +12,17 @@
...
-FILE_UPLOAD_APPEND = TEST_ASSET_DIR / "fileupload_append.xml"
+from test._utils import data_asset_path, xml_asset_path
+
+GET_XML = xml_asset_path("custom_view_get.xml")
+GET_XML_ID = xml_asset_path("custom_view_get_id.xml")
+POPULATE_PREVIEW_IMAGE = data_asset_path("Sample View Image.png")
+CUSTOM_VIEW_UPDATE_XML = xml_asset_path("custom_view_update.xml")
+CUSTOM_VIEW_POPULATE_PDF = data_asset_path("populate_pdf.pdf")
+CUSTOM_VIEW_POPULATE_CSV = data_asset_path("populate_csv.csv")
+CUSTOM_VIEW_DOWNLOAD = Path(xml_asset_path("custom_view_download.json"))
```
anyoung-tableau, line 23: I noticed some use `Path` from pathlib and some don't. Is that intentional?

bcantoni:
@jacalata this should target `development` instead, right?

jacalata:
This is too outdated to be worth merging, easier to do the work again.

PR #1617, opened 2025-05-16
Update permissions_item.py --added ExtractRefresh attribute

sparklingSky:


bcantoni:
@jacalata @jorwoods looks like this just went into `master`, so should we cherry-pick it back to `development` as well?

PR #1628, opened 2025-06-25
fix: virtual connections username

### LINKED ISSUE #1626: virtual connection's connection object does not have username field
**Describe the bug**
virtual connection's connection object does not have username field

**Versions**
 - Tableau Online
 - Python 3.11.9
 - TSC library version 0.37

**To Reproduce**

```python
def gen_connections(endpoint: Any) -> Iterator[ConnectionItem]:
    for obj in TSC.Pager(endpoint):
        endpoint.populate_connections(obj)

        obj_connections = obj.connections

        for conn in obj_connections:
            yield conn.username

print(list(gen_connections(server.virtual_connections)))
```

**Results**
a list of None objects

the xml response from the API call contains a field "username" that has the username, however this line gets from "userName", https://github.com/tableau/server-client-python/blob/28860189f440b180fed1488c9b93b51118b030ba/tableauserverclient/models/connection_item.py#L111C40-L111C76

jorwoods:
Closes #1626

VirtualConnections leverages the ConnectionItem object to parse the database connections server response. Most of other endpoints return "userName" and the VirtualConnections' "Get Database Connections" endpoint returns "username." Resolves the issue by allowing the ConnectionItem to read either. Update the test assets to reflect the actual returned value.

### DIFF
--- a/tableauserverclient/models/connection_item.py
+++ b/tableauserverclient/models/connection_item.py
@@ -120,7 +120,7 @@ class ConnectionItem:
             connection_item.embed_password = string_to_bool(connection_xml.get("embedPassword", ""))
             connection_item.server_address = connection_xml.get("serverAddress", connection_xml.get("server", None))
             connection_item.server_port = connection_xml.get("serverPort", connection_xml.get("port", None))
-            connection_item.username = connection_xml.get("userName", None)
+            connection_item.username = connection_xml.get("userName", connection_xml.get("username", None))
             connection_item._query_tagging = (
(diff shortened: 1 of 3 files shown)


jorwoods:
I also separately tested the [update connection](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_virtual_connections.htm#update_virtual_connection_database_connections) method, and it appears that request payload works with `userName`.

jacalata (approved):

PR #1631, opened 2025-07-05
feat: retrieve tableau server product name

### LINKED ISSUE #1592: [TYPE 3] [More detailed Server Info]
## Summary
The ServerInfo API call should return information about whether or not I am connecting to a Server or Cloud instance.

## Request Type
****Type 3: new functionality****
Requests for totally new functionality will generally be passed to the relevant dev team, but we probably can't give any useful estimate of how or when it might be implemented. If it is a feature that is 'about' the API or programmable access, here might be the best place to suggest it, but generally feature requests will be more visible in the [Tableau Community Ideas](https://community.tableau.com/s/ideas) forum and should go there instead.

## Description

Because some endpoints, payloads, and responses differ between Server and Cloud, the ServerInfo endpoint should also respond with information about which the session is connected to, so that methods can be appropriately enabled or disabled, and request payloads constructed that fit the system being connected to.

jorwoods:
Closes #1592

Calls a VizPortal API to retrieve the detailed product name, and attaches it to the TSC.Server object which will make determining what requests to build for things like subscriptions easier. Adds this functionality into the pre-existing `use_server_version` function because:

1. Users of the library will already know to call that method.
2. Users already aware that method makes API calls.

If the request errors or if the payload doesn't match the expected format, defaults to assuming TSC is talking with an on-prem Server instance.

Co-authored-by: emeric-dsj

### DIFF
--- a/tableauserverclient/server/endpoint/endpoint.py
+++ b/tableauserverclient/server/endpoint/endpoint.py
@@ -189,8 +189,8 @@ class Endpoint:
             loggable_response = helpers.strings.redact_xml(server_response.content.decode(server_response.encoding))
         return loggable_response
 
-    def get_unauthenticated_request(self, url):
-        return self._make_request(self.parent_srv.session.get, url)
+    def get_unauthenticated_request(self, url, parameters=None):
+        return self._make_request(self.parent_srv.session.get, url, parameters=parameters)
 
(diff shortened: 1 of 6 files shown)


jacalata:
Sorry, we can't include calls to these APIs because we have no commitment that they will remain stable or supported between releases.

PR #1637, opened 2025-07-15
Added new apis to workbook and datasource

vchavatapalli:


tableauserverclient/server/endpoint/datasources_endpoint.py
```diff
@@ -319,6 +319,71 @@ def update_connection(
         logger.info(f"Updated datasource item (ID: {datasource_item.id} & connection item {connection_item.id}")
         return connection
 
+    @api(version="3.26")
+    def update_connections(
+        self, datasource_item: DatasourceItem, connection_luids: list[str], authentication_type: str, username: Optional[str] = None, password: Optional[str] = None, embed_password: Optional[bool] = None
```
jorwoods, line 324: Should accept an `Iterable[ConnectionItem]` rather than taking a list of ids and then Connection attributes as keyword args.

tableauserverclient/server/endpoint/workbooks_endpoint.py
```diff
@@ -325,6 +325,72 @@ def update_connection(self, workbook_item: WorkbookItem, connection_item: Connec
         logger.info(f"Updated workbook item (ID: {workbook_item.id} & connection item {connection_item.id})")
         return connection
 
+    # Update workbook_connections
+    @api(version="3.26")
+    def update_connections(self, workbook_item: WorkbookItem, connection_luids: list[str], authentication_type: str, username: Optional[str] = None, password: Optional[str] = None, embed_password: Optional[bool] = None
```
jorwoods, line 330: Same as datasource method.

tableauserverclient/server/endpoint/datasources_endpoint.py
```diff
@@ -319,6 +319,71 @@ def update_connection(
...
+        Returns
+        -------
+        list of str
+            The connection LUIDs that were updated.
+        """
+        from xml.etree.ElementTree import Element, SubElement, tostring
+
+        url = f"{self.baseurl}/{datasource_item.id}/connections"
+
+        ts_request = Element("tsRequest")
```
jorwoods, line 358: Creating the XML payload belongs in the request_factory.

tableauserverclient/models/interval_item.py
```diff
@@ -318,4 +318,4 @@ def interval(self, interval_values):
         self._interval = interval_values
 
     def _interval_type_pairs(self):
-        return [(IntervalItem.Occurrence.MonthDay, self.interval)]
+        return [(IntervalItem.Occurrence.MonthDay, str(day)) for day in self.interval]
```
jorwoods, line 321: Unrelated change

jorwoods (changes requested):
1. Unit tests are missing for these endpoints. 
2. Can you point to the docs where you found these endpoints that support updating multiple connections at a time? I don't see it anywhere.

PR #1670, opened 2025-10-17
Add support for receiving "Customized Monthly" schedule intervals

### LINKED ISSUE #1365: Invalid interval value for a monthly frequency: Customized Monthly.
I have a schedule in Tableau Server which is set to run on a Monthly basis: Every 11ᵗʰ, 12ᵗʰ, 16ᵗʰ, 24ᵗʰ, 26ᵗʰ day of the month, at 5:05 PM.
When I use the below code to retrieve the specific schedule

`with server.auth.sign_in(tableau_auth):
    #print(id)
    sch_item = server.schedules.get_by_id('xxx')
    print(sch_item.name)`

It results in an error:
_ValueError: Invalid interval value for a monthly frequency: Customized Monthly._

I've used the Tableau REST API directly to see what the is output in the XML, where I see the interval monthDay is Customized Monthly. 

`<schedule id="f059d794-90dc-40d0-bfad-2ca78e437369" name="TEST MONTHLY SCHEDULE MAK" state="Active" priority="50" createdAt="2024-03-13T17:05:01Z" updatedAt="2024-03-16T16:05:14Z" type="Extract" frequency="Monthly" nextRunAt="2024-03-24T16:05:00Z" executionOrder="Parallel"><frequencyDetails start="17:05:00"><intervals><interval monthDay="Customized Monthly"/></intervals></frequencyDetails></schedule></tsResponse>`

I'm assuming there isn't a way to display the actual dates for the schedule, but does the interval item code need to be updated to handle the value of "Customized Monthly"?

**Versions**
 - Tableau Server version 2022.1.20
 - Python version 3.11.4
 - TSC library version 0.30

bcantoni:
Add handling to recognize "Customized Monthly" as a possible interval type that can be returned for schedules.

Also add a test with a simulated xml response to confirm.

Fixes #1365

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -48,7 +48,7 @@ jobs:
     - name: Test with pytest
       if: always()
       run: |
-        pytest test
+        pytest test -n auto
 
     - name: Test build
       if: always()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ dependencies = [
     'defusedxml>=0.7.1',  # latest as at 7/31/23
     'packaging>=23.1',   # latest as at 7/31/23
     'requests>=2.32',  # latest as at 7/31/23
-    'urllib3>=2.2.2,<3',
+    'urllib3>=2.6.0,<3',
     'typing_extensions>=4.0',
 ]
 requires-python = ">=3.9"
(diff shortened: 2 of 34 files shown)


jorwoods (approved):
I am not able to recreate a server responding with "Customized Monthly", but I see no issues with adding it.

jacalata (approved):
I think this string may have been removed since this was reported, but this will make the library work for people running against 2022 or other builds that return it.

PR #1797, opened 2026-06-25
fix: correct wheel packaging and add release safety checks

### LINKED ISSUE #1751: Pypi smoke tests should run at release

jacalata:
## Summary

- Fixes broken wheel structure caused by `include=[\"tableauserverclient*\"]` in `[tool.setuptools.packages.find]`, which on setuptools 80.10.2 doubles all subpackages as bare top-level entries in the wheel zip (e.g. `models/`, `server/`, `helpers/` alongside `tableauserverclient/`). Switched to an `exclude`-based approach which is robust across setuptools versions.
- Adds `test/test_packaging.py` with a wheel structure test that checks the zip root contains only `tableauserverclient/` — catches both this bug and the v0.39 regression where the package itself was missing entirely.
- Updates `publish-pypi.yml` to split into four sequential jobs: build + check wheel → publish to TestPyPI → smoke test from TestPyPI (Linux/macOS/Windows) → publish to real PyPI. Real PyPI is now unreachable unless the TestPyPI smoke test passes.
- Updates `pypi-smoke-tests.yml` to also trigger on `release: published`, not just the daily schedule.

Closes #1751.

## Test plan

- [ ] Trigger `publish-pypi.yml` manually via `workflow_dispatch` on this branch to verify build, wheel check, TestPyPI publish, and smoke test jobs run correctly (real PyPI publish will be skipped on non-master)
- [ ] Confirm `pytest test/test_packaging.py -m packaging` passes after `python -m build`
- [ ] Confirm `pytest test/test_packaging.py -m packaging` fails against the broken wheel built from `development` without this fix

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,9 +1,7 @@
 name: Publish to PyPi
 
-# This will build a package with a version set by versioneer from the most recent tag matching v____
-# It will publish to TestPyPi, and to real Pypi *if* run on master where head has a release tag
-# For a live run, this should only need to be triggered by a newly published repo release. 
-# This can also be run manually for testing
+# Builds the package, checks wheel structure, publishes to TestPyPI, runs smoke
+# tests against TestPyPI, then publishes to real PyPI only if all prior steps pass.
 on:
   release:
     types: [published]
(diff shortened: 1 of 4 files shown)


jacalata:
Closing in favour of PR from upstream branch.

PR #1806, opened 2026-06-26
feat: Add webhook update method with isEnabled/statusChangeReason

### LINKED ISSUE #1135: [Type 1] Add support for API method Update a Webhook
## Description
Tableau REST API provides a method to modify the properties of an existing webhook which is not currently exposed through TSC:
https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#update_webhook

This method is useful for easily re-enabling webhooks when they get disabled after four failed re-tries. 

Consider surfacing additional webhook attributes like **isEnabled** and **statusChangeReason** as a part of **webhook_item** model to better support webhook management.

As per note in Tableau REST API documentation review TSC code to use the event attribute of the webhook to specify the triggering API event for the webhook as the webhook-source element is being deprecated in future versions of Tableau webhooks.

jacalata:
## Summary
- Adds `Webhooks.update(webhook_item)` (REST API v3.6) for modifying an existing webhook's name, event, destination URL, and enabled state
- Adds `is_enabled` (`bool`) and `status_change_reason` (`str`) fields to `WebhookItem`, parsed from the `isEnabled` and `statusChangeReason` XML attributes returned by the server
- `RequestFactory.Webhook.update_req()` serializes all updatable fields; `is_enabled` is omitted when `None` to support partial updates; `statusChangeReason` is intentionally not serialized (server-set, read-only)
- Raises `MissingRequiredFieldError` if the webhook item has no ID
- Handles both `webhook-source-event-*` (legacy) and `webhook-event-*` (newer) event name prefixes in the event setter

Closes #1135

## Schema compliance
`isEnabled` and `statusChangeReason` are both defined on `webhookType` in ts-api_3_29.xsd. The `update_req()` child element structure (`webhook-source`, `webhook-destination`) matches the schema. `webhook-event-*` style event names are a live API extension not yet reflected in the published XSD; handling them is consistent with the pre-existing behavior in `create_req()`.

## Test plan
- [x] `python -m pytest test/test_webhook.py -v` -- 22 tests, all pass
- [x] mypy: no issues
- [x] Verify `server.webhooks.update(item)` returns updated `WebhookItem` with correct fields
- [x] Verify updating only `is_enabled=False` leaves name/url/event unchanged on server
- [x] Verify `MissingRequiredFieldError` raised when item has no ID
- [x] Verify `isEnabled` attribute absent from request XML when `is_enabled` is `None`

### DIFF
--- a/tableauserverclient/models/webhook_item.py
+++ b/tableauserverclient/models/webhook_item.py
@@ -44,6 +44,12 @@ class WebhookItem:
 
     owner_id : str | None
         The identifier (luid) of the user who owns the webhook.
+
+    is_enabled : bool | None
+        Whether the webhook is enabled. Disabled webhooks do not fire.
+
+    status_change_reason : str | None
+        The reason the webhook status last changed (e.g. why it was disabled).
     """
 
     def __init__(self):
@@ -52,8 +58,10 @@ class WebhookItem:
         self.url: str | None = None
         self._event: str | None = None
         self.owner_id: str | None = None
+        self.is_enabled: bool | None = None
+        self.status_change_reason: str | None = None
 
-    def _set_values(self, id, name, url, event, owner_id):
+    def _set_values(self, id, name, url, event, owner_id, is_enabled=None, status_change_reason=None):
         if id is not None:
             self._id = id
         if name:
@@ -64,6 +72,10 @@ class WebhookItem:
             self.event = event
         if owner_id:
             self.owner_id = owner_id
+        if is_enabled is not None:
+            self.is_enabled = is_enabled
+        if status_change_reason is not None:
+            self.status_change_reason = status_change_reason
 
     @property
     def id(self) -> str | None:
@@ -116,7 +128,14 @@ class WebhookItem:
         if owner_tag is not None:
             owner_id = owner_tag.get("id", None)
 
-        return id, name, url, event, owner_id
+        is_enabled = None
+        is_enabled_str = webhook_xml.get("isEnabled", None)
+        if is_enabled_str is not None:
+            is_enabled = is_enabled_str.lower() == "true"
+
+        status_change_reason = webhook_xml.get("statusChangeReason", None)
+
+        return id, name, url, event, owner_id, is_enabled, status_change_reason
 
     def __repr__(self) -> str:
-        return f"<Webhook id={self.id} name={self.name} url={self.url} event={self.event}>"
+        return f"<Webhook id={self.id} name={self.name} url={self.url} event={self.event} is_enabled={self.is_enabled}>"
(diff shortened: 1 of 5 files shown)


bcantoni:
## Code Review

Clean, well-structured PR. Here are the issues worth addressing before merge:

### Issues

**Convention deviation — `update` uses `from_response()[0]` instead of `copy.copy` + `_parse_common_tags`**

`webhooks_endpoint.py:150` — The project's documented update pattern (used in `users_endpoint.py:208`, `datasources_endpoint.py:353`) is:
```python
updated = copy.copy(item)
return updated._parse_common_tags(server_response.content, ns)
```
This preserves locally-set fields that the server's partial response omits. Using `from_response()[0]` works here because the server always returns a complete webhook and `WebhookItem` has no `_parse_common_tags`, but it deviates from convention. Either add `_parse_common_tags` to `WebhookItem` for consistency, or leave a short comment explaining why the simpler approach is safe.

**`status_change_reason` is a mutable public attribute**

`webhook_item.py:61` — It's stored as a plain `self.status_change_reason: str | None = None` and can be set freely by callers. Since it's server-assigned and not serialized by `update_req`, a caller who sets it won't get an error but also won't get the intended behavior. Per the project's property pattern for server-assigned/read-only fields (`id`, `created_at`, etc.), this should be a getter-only property (no setter), or the docstring should explicitly say "server-set, do not write."

**`test_update` doesn't assert the request body**

`test/test_webhook.py:184` — The test only validates response parsing. There's no `m.last_request.text` assertion confirming the correct XML was sent in the PUT body. The `test_update_request_factory_*` tests cover serialization independently, so coverage isn't missing, but an integration-level check tying the endpoint to its request body would be more complete.

**Inconsistent `-> None` annotation sweep**

`test_webhook.py:91–120` — The PR adds `-> None` to exactly 4 of the existing test functions while leaving others without it. Either annotate all test functions in the file or none (the file had none before this PR).

**Missing CHANGELOG entry**

Per repo conventions, user-visible additions get a changelog bullet with the PR number. `server.webhooks.update()` is a new public method that qualifies.

### Minor

- `webhook_item.py` — `update_req` accesses `webhook_item._event` directly (intentional, since the property strips the prefix), but a brief inline comment would help readers understand why the private field is used instead of the public property.
- `webhooks_endpoint.py:131` — The docstring `Returns` description could match the phrasing used in other update methods for consistency.

jacalata:
Valid. Landed the convention fix in `3f4b08a`:

- Added `WebhookItem._parse_common_tags` matching the pattern in `UserItem` and `WorkbookItem` (name refers to XML common tags, not Tableau item tags).
- Reworked `Webhooks.update()` to `copy.copy(item) + updated._parse_common_tags(...)`, matching `users_endpoint.py:208` and `datasources_endpoint.py:353`.
- New test `test_update_preserves_locally_set_fields_omitted_by_server` — sets local `url`, `owner_id`, `is_enabled=False`, mocks a partial server response omitting those three, asserts they survive.

jarhun88 (approved):

PR #1813, opened 2026-07-10
feat: add operator parameter to vf() for range filters

### LINKED ISSUE #1752: [Type 2] Support range filter operators (GreaterThan, LessThan, Between) in PDFRequestOptions.vf() for view filters
## Summary
Add support for range filter operators (GreaterThan, LessThan, Between) in PDF view filters using `.vf()` method

## Request Type
**Type 2: add a REST API and support it in tsc.**

This is functionality that exists in Tableau dashboards (continuous range filters) but cannot be properly controlled through the PDF generation API.

### Current Limitation
When using `PDFRequestOptions.vf()` to apply view filters for PDF generation, there is no way to specify filter **operators** such as `GreaterThan`, `LessThan`, or `Between`. The current implementation only supports:
- Single discrete values: `pdf_opts.vf('Region', 'West')`
- Multiple discrete values (IN operator): `pdf_opts.vf('Category', 'Furniture,Technology')`
- Ambiguous comma-separated syntax: `pdf_opts.vf('Score', '10,50')`

### Problem
For **continuous range filters** configured in Tableau dashboards, the comma-separated syntax `'10,50'` is ambiguous and may be interpreted as:
- `IN (10, 50)` - only values exactly 10 or 50
- `BETWEEN 10 AND 50` - range from 10 to 50

Currently, there is no way to explicitly specify that a filter should use `BETWEEN`, `GreaterThan`, or other range operators, even though these filters work perfectly when using the dashboard interactively.

### Proposed Solution
Add operator support to `.vf()` similar to how `RequestOptions.filter.add()` works for API queries:

**Current API (RequestOptions filtering):**
```python
req_option = TSC.RequestOptions()
req_option.filter.add(TSC.Filter(
    TSC.RequestOptions.Field.Score,
    TSC.RequestOptions.Operator.GreaterThanOrEqual,
    '10'
))
```

**Proposed enhancement for PDF view filters:**
```python
pdf_opts = TSC.PDFRequestOptions()
pdf_opts.vf('SCORE', '10', operator=TSC.RequestOptions.Operator.GreaterThanOrEqual)
# OR
pdf_opts.vf('SCORE', min_value='10', max_value='50')  # For BETWEEN
```

### Use Case & Value
- **Frequency**: This is needed any time programmatic PDF generation requires filtering on continuous numeric or date fields
- **Current workaround**: None - users must either accept ambiguous behavior or modify dashboard filters, which affects all users
- **Work saved**: Eliminates need to create separate dashboards with hardcoded filters just for PDF generation
- **Business impact**: Critical for automated reporting where PDFs need to show data for specific ranges (e.g., "Show only high-risk scores 80-100", "Show sales data > $1M")

### Example Scenario
A Tableau dashboard has a continuous slider filter for `SCORE` (range: 0-100). When generating automated weekly PDFs:
- **Needed**: Filter to show only scores BETWEEN 10 AND 50
- **Current behavior**: Passing `[10, 50]` may show only exact values 10 and 50
- **Expected behavior**: Should show all values in the range 10-50, just like moving the slider in the web UI

This feature would bring view filter capabilities in line with the existing `RequestOptions.filter` API and match the functionality available in the Tableau web interface.

jacalata:
## Summary
- Adds optional `operator` parameter to `RequestOptions.vf()` (and by inheritance `PDFRequestOptions`, `ImageRequestOptions`, etc.)
- When provided, encodes the filter as `operator:value` (e.g. `gt:1000`) matching the REST API's range filter syntax
- Backwards-compatible: omitting `operator` preserves existing exact-match behaviour

Closes #1752

## Test plan
- [ ] `test_vf_with_operator` in `test/test_request_option.py` verifies `gt:1000` encoding
- [ ] Existing `vf` tests continue to pass (exact-match unchanged)

### DIFF
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -349,7 +349,7 @@ class _DataExportOptions(RequestOptionsBase):
         self._append_view_filters(params)
         return params
 
-    def vf(self, name: str, value: str) -> Self:
+    def vf(self, name: str, value: str, operator: str | None = None) -> Self:
         """Apply a filter based on a column within the view.
         Note that when filtering on a boolean type field, the only valid values are 'true' and 'false'
 
@@ -363,12 +363,18 @@ class _DataExportOptions(RequestOptionsBase):
         value: str
             The value to filter on
 
+        operator: str, optional
+            A comparison operator from RequestOptions.Operator (e.g. 'gt', 'lte').
+            When provided the filter is encoded as ``operator:value``.
+            Defaults to an exact-match filter when omitted.
+
         Returns
         -------
         Self
             The current object
         """
-        self.view_filters.append((name, value))
+        encoded_value = f"{operator}:{value}" if operator else value
+        self.view_filters.append((name, encoded_value))
         return self
 
     def parameter(self, name: str, value: str) -> Self:
--- a/test/test_request_option.py
+++ b/test/test_request_option.py
@@ -238,6 +238,23 @@ def test_vf(server: TSC.Server) -> None:
     assert "tabloid" in query_string["type"]
 
 
+def test_vf_with_operator(server: TSC.Server) -> None:
+    server.version = "3.23"
+    with requests_mock.mock() as m:
+        m.get(requests_mock.ANY)
+        url = server.workbooks.baseurl + "/456/data"
+        opts = TSC.PDFRequestOptions()
+        opts.vf("Sales", "1000", operator=TSC.RequestOptions.Operator.GreaterThan)
+        opts.vf("Region", "West")
+
+        resp = server.workbooks.get_request(url, request_object=opts)
+        query_string = parse_qs(resp.request.query)
+    assert "vf_sales" in query_string
+    assert "gt:1000" in query_string["vf_sales"]
+    assert "vf_region" in query_string
+    assert "west" in query_string["vf_region"]
+
+
 # Test req_options for versions below 3.7
 def test_vf_legacy(server: TSC.Server) -> None:
     server.version = "3.6"


bcantoni:
## Code Review

**Summary:** Adds an optional `operator` parameter to `_DataExportOptions.vf()` (inherited by `PDFRequestOptions`, `ImageRequestOptions`, etc.), encoding filters as `operator:value` (e.g. `gt:1000`) instead of a bare value. Backwards compatible when omitted. Closes #1752.

**Mechanical checks (on the PR branch):** tests pass (42/42), `black --check` clean, `mypy` clean.

### 🔴 Critical: the feature doesn't do what it claims

Tableau's REST API documentation for view filters (`vf_*`) explicitly states:

> "Wild cards, value ranges, and non-equivalencies, like contains, less than, or greater than, aren't supported."

View filters (`vf_field=value`) only ever do **exact string match** server-side — unlike the separate `filter=field:operator:value` endpoint-filtering syntax (used elsewhere in this codebase via `Filter`/`RequestOptions.filter`), which *does* support operators. This PR conflates the two mechanisms. Sending `vf_sales=gt:1000` will not filter for Sales > 1000 — the server will look for rows where the field literally equals the string `"gt:1000"`, silently returning an empty/wrong result set rather than erroring. The linked issue (#1752) proposes this same approach, so the misunderstanding originates there.

Given this, I don't think this should merge as-is — it adds a plausible-looking but non-functional API surface that would mislead callers into believing range filtering works when it doesn't. It'd be worth checking with the API/docs team (or re-triaging #1752) on whether there's a different mechanism (e.g. a different endpoint, or dashboard-level filter configuration) that actually supports range filters for PDF/image exports, since `.vf()` on view filters isn't it.

### Other findings

- **Test coverage gap**: the new `test_vf_with_operator` test only asserts the client-side string built (`gt:1000` appears in the outgoing query param) via `requests_mock` — it can't and doesn't catch that the real server ignores/misinterprets this syntax.
- **No input validation**: unlike `query.py`'s `_parse_shorthand_filter` (which raises `ValueError` for an unrecognized operator), `vf()` accepts any string as `operator` with no validation against `RequestOptions.Operator`.

bcantoni:
@jacalata not sure how to interpret claude's review findings. Is it missing the fix or do the docs need to updated?

jacalata:
Claude was right. Verified `vf_` is exact-match only server-side; `vf_sales=gt:1000` fails to parse and the filter is silently dropped rather than applied as `Sales > 1000`. The docs are accurate.

I looked at the broader question of "any way to range-filter a view via a URL/API." Summary:

- REST `vf_*` (PDF/image/CSV export): exact-match only, documented server-side.
- View URL `?field=value` embed params: exact-match only.
- Embedding API v2/v3 `Worksheet.applyRangeFilterAsync({min, max})`: yes, but runs only in a browser context.
- Workbook-author-side workaround: expose two Tableau parameters (e.g. `MinSales`, `MaxSales`) + a calculated boolean `[Sales] >= [MinSales] AND [Sales] <= [MaxSales]` on the filter shelf, then hit `?MinSales=1000&MaxSales=5000`. This is a workbook-side design.

So I'll close this PR. The TSC docs on `vf()` are silent on operator support — they just show `vf('Region', 'South')` examples — so I'll add a note that view filters are exact-match only, pointing at the server-side REST docs for detail.

jacalata:
Closing per the discussion above — vf_ is exact-match only server-side, no client-side change can add range support. Filing a docs PR to note the limitation on vf().

PR #1814, opened 2026-07-15
Update several actions to the latest stable versions

bcantoni:
Third-party action versions not updated in this PR. The [simonw/actions-latest](https://github.com/simonw/actions-latest) list only tracks actions/* org actions, so these three third-party actions were left untouched but should be revisited separately:

- pypa/gh-action-pypi-publish@release/v1 — pinned to the maintainer's documented floating branch ref (their README explicitly shows this usage). Latest concrete tag is v1.14.0. Per pypa's own "pro tip," pinning to an exact tag/SHA instead of the branch is more reproducible — worth considering.
- MishaKav/pytest-coverage-comment@main — currently tracks the main branch directly (not a version tag). Latest release is v1.10.0. Should pin to a tagged version instead of main to avoid unreviewed changes landing silently.
- archive/github-actions-slack@v2.8.0 — real version bump available to v3.1.0 (major version jump). Needs a changelog/breaking-change check before upgrading.


jacalata (approved):

jacalata:
created https://github.com/tableau/server-client-python/issues/1818 to track the notes

PR #1816, opened 2026-07-20
fix: normalize CRLF line endings to LF in six Python files

bcantoni:
Several files had CRLF line endings baked in from a prior black-version bump commit, causing ^M noise in diffs.

I think these came in accidentally in #1780


bcantoni:
@jacalata @jorwoods I'm not 100% sure about this one -- it started because those 2 /models/ files were being tagged as needing to be reformatted for me, and I noticed the line ending were off. This PR works but everyone would need to be at least on Python 3.13 for Black to work correctly. Maybe we should just leave it as is for now.

jorwoods:
Could adding an [editorconfig](https://editorconfig.org/) file to the repo permit users to likely keep a lower version locally and keep the higher version in the pipeline?

jorwoods:
Also, 3.10 deprecates and 3.15 releases in 3 months.

jacalata:
You could a `.gitattributes` file to this PR. That's apparently the standard way to enforce line endings repo-wide and removes the dependency on each contributor's local `core.autocrlf` setting. Something like:

```
* text=auto
*.py text eol=lf
```

That would fix the root cause for everyone rather than just normalising the files once.

bcantoni:
@jacalata good call - Claude had suggested that to me earlier but I kinda ignored it. Added to this branch now.

jacalata:
The CRLF-vs-LF diff noise is hiding the actual conflicts from the GitHub UI. Locally, you can force git's merge machinery to see through the line-ending diff by using `-Xrenormalize`:

```sh
git fetch origin
git checkout bcantoni/fix-black-line-endings
git merge origin/development -Xrenormalize
```

jacalata:
@copilot resolve the merge conflicts on this branch.

jacalata:
@copilot resolve the merge conflicts in this pull request

jacalata (approved):

stephendeoca (approved):

PR #1817, opened 2026-07-20
fix: unify streaming download chunk size with dedicated DOWNLOAD_CHUNK_SIZE_MB knob

### LINKED ISSUE #1374: Issue with Downloading Large Views from Tableau Server to CSV
**Describe the bug**
I'm encountering an issue when trying to download a view that is around 200 MB in size. My script works perfectly for smaller files (5 MB or 10 MB), but it keeps running indefinitely and doesn't write data into the CSV file for the larger view.

The script successfully logs "CSV populated," indicating that server.views.populate_csv completes, but the file is not written for larger views.

Here's the code I'm using:

```
def download_file(client, server, content_url, output_file, region):

    # NOTE: Monkey patched TSC.RequestOptions.Field class

    # since the contentUrl filter parameter was not supported

    setattr(TSC.RequestOptions.Field, 'ContentURL', 'contentUrl')

    req_option = TSC.RequestOptions()

    req_option.filter.add(TSC.Filter(TSC.RequestOptions.Field.ContentURL,

                     content_url))

    logger.info('Fetching views...')

    views, pagination_item = server.views.get(req_option)

    logger.info(f'Views fetched: {len(views)}')

    if len(views) == 1:

      only_matched_index = 0

      view = views[only_matched_index]

      logger.info('Populating CSV...')

      server.views.populate_csv(view, TSC.CSVRequestOptions(maxage = 5))

      logger.info('CSV populated.')

      with open(output_file, 'wb') as f:

        f.write(b''.join(view.csv))

      logger.info(f"CSV DATA WRITTEN TO the {output_file}")

 
```

OUTPUT:

INFO:__main__:Fetching views...

INFO:__main__:Views fetched: 1

INFO:__main__:Populating CSV...

INFO:__main__:CSV populated.

Could anyone suggest what might be going wrong with handling larger views? Are there any additional steps or considerations when dealing with larger files in Tableau Server that I might be missing?

Any insights or suggestions would be greatly appreciated!

Thank you!

**Versions**
 - Python version:3.9
 - TSC library version:0.30

jacalata:
Closes #1374.

## Summary

Unifies streaming chunk size across all download paths in `Endpoint`. Previously:

- `DownloadableMixin._download_content` (workbook / datasource / flow) used a hard-coded 1024-byte chunk since the mixin was introduced.
- `Views._get_view_csv` / `_get_view_excel` and `CustomViews._get_custom_view_csv` also used hard-coded 1024 bytes.
- A `TSC_CHUNK_SIZE_MB` env var / `config.CHUNK_SIZE_MB` knob existed for uploads (`fileuploads_endpoint`) but was never applied to downloads.

This PR:

- Adds `DownloadableMixin._stream_content(url, req_options) -> Iterator[bytes]` for endpoints whose callers want to lazily iterate the response (Views, CustomViews).
- Adds a dedicated `DOWNLOAD_CHUNK_SIZE_MB` config value (default **1 MB**, overridable via `TSC_DOWNLOAD_CHUNK_SIZE_MB`). Routes both `_download_content` and `_stream_content` through it.
- Migrates Views and CustomViews CSV/Excel to use `_stream_content`.

## Why a separate knob

The upload chunk (`CHUNK_SIZE_MB` = 50 MB) and the download chunk have different tradeoffs. On upload, larger chunks reduce per-request overhead. On download, `requests.iter_content(N)` buffers up to `N` bytes before yielding the first chunk -- on a 1 Mbps link a 50 MB chunk means ~7 minutes of apparent hang before the caller sees any bytes, exactly the class of symptom this PR is trying to fix. 1 MB is empirically a reasonable balance; callers with different needs can tune via the env var.

## Behavior change

The 1 KB default (which required ~200,000 generator iterations for a 200 MB CSV, causing apparent hangs when callers materialize the full stream via `b''.join(view.csv)`) is replaced by the tunable 1 MB default. Callers that previously deliberately streamed 1 KB at a time will now hold up to 1 MB resident per chunk -- memory-constrained callers can drop this via `TSC_DOWNLOAD_CHUNK_SIZE_MB=<smaller-value>`.

Upload / chunked-publish continues to use `CHUNK_SIZE_MB` / `TSC_CHUNK_SIZE_MB` (default 50 MB) unchanged.

## Test plan

- [x] `test_view.py`: `test_stream_content_uses_configured_chunk_size` asserts `DOWNLOAD_CHUNK_SIZE_MB * BYTES_PER_MB` reaches `iter_content` (was `CHUNK_SIZE_MB`)
- [x] `test_workbook.py`: same guard for `DownloadableMixin._download_content`
- [x] `test_fileuploads.py`: upload path still uses `CHUNK_SIZE_MB`
- [x] 104 tests pass across the three affected files


jorwoods:
Should this use the `config.CHUNK_SIZE_MB` as the download size rather than a fixed 64mb?

bcantoni:
## Code Review

**Summary:** Raises the `iter_content` chunk size in `ViewsEndpoint._get_view_csv` / `_get_view_excel` from a hardcoded 1024 bytes to `config.CHUNK_SIZE_MB * BYTES_PER_MB` (50 MB default), fixing the effective hang reported in #1374 for large (~200 MB) view downloads.

### Strengths
- Correct root-cause fix — 1 KB chunks meant ~200k generator iterations for a 200 MB download; a much larger chunk size removes that overhead.
- Reuses existing config infrastructure (`config.CHUNK_SIZE_MB`, `BYTES_PER_MB`) rather than a new magic number, matching the pattern already used in `fileuploads_endpoint.py`, `datasources_endpoint.py`, and `custom_views_endpoint.py`. This also gives users the existing `TSC_CHUNK_SIZE_MB` env var as a tuning knob.
- Small, surgical change with no other behavior differences — low risk.
- `black --check` passes on the changed file.

### Issues / Gaps

**Inconsistent chunk size left in place elsewhere.** The same hardcoded `1024`-byte pattern still exists, unfixed, in:
- `custom_views_endpoint.py:234` (`_get_custom_view_csv`) — directly parallel to `_get_view_csv`, the exact method being fixed here, and likely has the same large-download-hang risk.
- `flows_endpoint.py:235,242`
- `datasources_endpoint.py:1069,1076`
- `workbooks_endpoint.py:1131,1138`

Worth confirming whether `_get_custom_view_csv` is intentionally out of scope for this PR, since it's the closest analog to the code being changed.

**No test asserts the new chunk size.** `test_populate_csv`, `test_populate_csv_default_maxage`, `test_populate_excel`, and `test_filter_excel` all mock small fixture files and only assert the *reassembled* content is correct — they'd pass identically whether the chunk size were 1024 bytes or 50 MB, so they don't protect this change from regressing. `test_fileuploads.py:85-90` has a good existing pattern for this (sets `TSC_CHUNK_SIZE_MB` via env and asserts the actual chunk length) that could be adapted here, or alternatively mock/patch `iter_content` and assert it's called with the expected size.

The PR description's test plan checkbox is also still unchecked.

### Correctness
`config.CHUNK_SIZE_MB * BYTES_PER_MB` = 50 * 1024 * 1024 ≈ 50 MB by default — consistent with the existing upload chunk-size default. No correctness or security concerns with the change itself.

---

bcantoni:
@jacalata kind of funny how claude code keeps finding issues with a claude code aided PR :) Consider doing those consistency points if they make sense.

jacalata:
Yea, clearly I need to run a more adversarial set of claudes! Turned out the consistency points were important: since I opened this PR, #1803 landed on development and extracted the streaming-to-file logic used by workbooks/datasources/flows into `DownloadableMixin._download_content` — but still using hardcoded 1024-byte chunks. Same performance bug applies to those endpoints, and `custom_views_endpoint._get_custom_view_csv` had the same 1024 too.

Expanded the PR scope:

- `DownloadableMixin._download_content` — chunk size now `config.CHUNK_SIZE_MB * BYTES_PER_MB`. Benefits workbooks, datasources, flows.
- New sibling `DownloadableMixin._stream_content` — for endpoints that yield `Iterator[bytes]` instead of writing to a file. Same chunk-size config.
- Views and CustomViews migrate to `_stream_content`, removing their duplicate streaming loops.
- Regression tests added for both mixin methods.

Callers using `b"".join(view.csv)` or iterating lazily continue to work identically; only the internal chunk size changes.

tableauserverclient/server/endpoint/endpoint.py
```diff
@@ -368,18 +374,36 @@ def _download_content(
             m = Message()
             m["Content-Disposition"] = server_response.headers["Content-Disposition"]
             filename = m.get_filename(failobj="")
+            chunk_size = config.CHUNK_SIZE_MB * BYTES_PER_MB
             if isinstance(filepath, _io_types_w):
```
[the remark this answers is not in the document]
jacalata: Correct that the code uses `config.CHUNK_SIZE_MB * BYTES_PER_MB` (default 50 MB), not the 65,536 bytes the PR body still claims. Sequence of events: the PR started as a bump from 1024 to 65,536 in the two view helpers, but grew into unifying chunk sizing across all download paths via `DownloadableMixin._stream_content` + `_download_content` using the same `config.CHUNK_SIZE_MB` knob. The PR body is out of date; updating it now.

Two things worth separating:

1. **Consistency**: this PR makes view/custom-view streaming use the same chunk size as workbook/datasource/flow downloads (pre-existing use of `config.CHUNK_SIZE_MB` on line 378). That's intentional and correct — one knob for everything, tunable via `TSC_CHUNK_SIZE_MB`.

2. **Whether 50 MB is the right default**: legitimate concern for interactive streaming of small CSVs (`iter_content(50 MB)` on a 500 KB response yields one chunk with the whole body — functionally fine but defeats lazy processing). This affects both the new `_stream_content` and the pre-existing `_download_content`. Retuning the default is a separate change, worth doing but not scoped to this PR.

CHANGELOG.md
```diff
@@ -5,6 +5,21 @@
   hierarchy path (e.g. `"Marketing/Q1 Reports"`). The walk is performed level by
   level using the REST API name filter, so a path with *n* components issues *n*
   requests. Returns the matching `ProjectItem` or `None` if no project is found.
+* Unified streaming download chunk size for the file-backed download paths --
+  `views.populate_csv` and `views.populate_excel`, the CSV/Excel branches of
+  `custom_views.*`, and `workbooks.download` / `datasources.download` /
+  `flows.download`. Previously the mixed 1024-byte and 10240-byte chunks caused
```
jarhun88, line 11: Minor: this says the pre-PR code used "mixed 1024-byte and 10240-byte chunks," but the diff (and the pre-PR `endpoint.py`/`views_endpoint.py`/`custom_views_endpoint.py`) shows every download path hardcoded `iter_content(1024)` -- there's no 10240-byte chunk anywhere in the codebase (checked via clone + grep). Worth fixing the changelog wording so it doesn't overstate the prior inconsistency; doesn't affect the code itself.

— Jaehun Bot

tableauserverclient/config.py
```diff
@@ -14,10 +14,38 @@ def FILESIZE_LIMIT_MB(self):
...
+    # back to the 1 MB default; 0 or a negative chunk size would silently
+    # corrupt the output because iter_content interprets it as "read all".
+    @property
+    def DOWNLOAD_CHUNK_SIZE_MB(self) -> int:
+        raw = os.getenv("TSC_DOWNLOAD_CHUNK_SIZE_MB", "1")
+        try:
+            value = int(raw)
+        except ValueError:
+            # invalid env value; fall back to default
+            value = 1
```
jarhun88, line 46: is default hardcoded anywhere

jarhun88 (approved):

PR #1826, opened 2026-07-22
fix: resolve mypy 2.3 type errors (unblocks dependabot #1821)

jacalata:
## Summary

Two type errors surfaced by mypy 2.3 (dependabot PR #1821):

**`tableauserverclient/helpers/strings.py:42`**
`tostring()` returns `bytes` regardless of the `T` type variable constraint. mypy 2.3 correctly rejects this. Fix: `cast(T, tostring(...))` since the two call sites pass `encoding="unicode"` for `str` and `None` for `bytes`, making the cast always correct at runtime.

**`test/test_datasource.py:580`**
`unittest.mock.patch.object` was used without `import unittest.mock` — only `import unittest` was present. mypy 2.3 no longer resolves submodule attributes through the parent import. Fix: add explicit `import unittest.mock`.

## Test plan

- [x] `mypy tableauserverclient/helpers/strings.py test/test_datasource.py` passes
- [x] `pytest test/test_datasource.py` passes (55 tests)
- [x] CI green on this PR unblocks merging #1821

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
 repository = "https://github.com/tableau/server-client-python"
 
 [project.optional-dependencies]
-test = ["black==26.3.1", "build", "mypy==1.4", "pytest>=7.0", "pytest-cov", "pytest-subtests",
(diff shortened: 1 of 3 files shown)


bcantoni (approved):
@jacalata approved. Looks like the conflict is a mix of mypy and black versions in your branch versus development.

PR #1835, opened 2026-07-28
docs: note vf() view filters are exact-match only

jacalata:
Adds a note near the shared `ImageRequestOptions` / `PDFRequestOptions` / `CSVRequestOptions` section explaining that `vf()` view filters are exact-match only, with a pointer to the Tableau REST API docs and mentioning the workbook-parameter workaround for range filtering.

Followup to closed PR #1813, which had tried to add operator support (`gt:`, `lt:` etc.) to `vf()` client-side — verified against server code that the `vf_*` REST endpoint parses values by data type with no operator recognition, silently dropping filters like `vf_sales=gt:1000`. The client can't fix this; a note is the right response.

## Test plan

- [x] visual review of the rendered section — sits between the "You can use `ImageRequestOptions` and `PDFRequestOptions`..." intro and the "For more information about filtering and sorting..." link
- [ ] once merged, links resolve on tableau.github.io/server-client-python

### DIFF
--- a/docs/api-ref.md
+++ b/docs/api-ref.md
@@ -5580,6 +5580,8 @@ You can use the `Sort` and `RequestOptions` classes to filter and sort the follo
 
 You can use the `ImageRequestOptions` and `PDFRequestOptions` to set options for views returned as images and PDF files.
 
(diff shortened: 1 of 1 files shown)


jacalata:
Merging without review as this is docs-only

PR #1840, opened 2026-07-29
docs: expand RequestOptions.vf() with vf_ wire-format rules

jacalata:
## Summary

The `vf_` REST view filter has several server-side quirks that aren't obvious from the current one-line `vf()` docstring. This expands it to document what values actually mean on the wire: exact-match by default, `,` as OR-list, `\,` for a literal comma, the empty-value override behavior, no wildcards, no operators, boolean `'true'`/`'false'` only.

Callers who build `vf_` requests through TSC no longer have to reverse-engineer server behavior to figure out why their filter returned zero rows.

## Why now

Working through `--filter` parsing bugs in tabcmd, I verified the wire-format behavior end-to-end against Tableau Cloud. Recording those findings in the library the tabcmd (and other callers) build on top of, so the next person doesn't have to redo the experiment.

Notable server-side behaviors documented:

- `Rock\, Paper\, Scissors` — backslash-escape works for a literal comma. `%2C` does NOT.
- `vf_Region=` (empty value) — overrides any workbook-embedded filter on that column. Undocumented but stable and useful.
- `*` — NOT a wildcard on `vf_`. It's a wildcard on the `filter=` list-endpoint syntax, which is a different query parameter.

## Test plan

- [x] `help(PDFRequestOptions.vf)` renders the expanded docstring cleanly
- [x] mypy passes (pre-commit hook)
- [x] No behavior change; docstring-only

### DIFF
--- a/tableauserverclient/server/request_options.py
+++ b/tableauserverclient/server/request_options.py
@@ -351,9 +351,37 @@ class _DataExportOptions(RequestOptionsBase):
 
     def vf(self, name: str, value: str) -> Self:
         """Apply a filter based on a column within the view.
-        Note that when filtering on a boolean type field, the only valid values are 'true' and 'false'
 
-        For more detail see: https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_filtering_and_sorting.htm#Filter-query-views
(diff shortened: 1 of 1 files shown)


jacalata:
docs only change, merging with AI review only

Issue #1658, opened 2025-09-09
Unabled to create Subscriptions with schedule as "On Extract Refresh"


rajkotha77:
Subscriptions can be created on specific schedules. However, it is not possible to create subscriptions based on extract refresh. The SubscriptionItem class (subject, schedule_id, user_id, target) requires schedule_id as a mandatory parameter, so it cannot be ignored or set to None.

I have read some articles to ignore/ dont pass "schedule_id" but it doesn't seems to be working.

Env:
 - Tableau Server: On Prem: 2025.1.1 (20251.25.0415.1018) 64-bit Windows
 - Python version: 3.12.0
 - TSC library version: 0.38.

**To Reproduce**
Try to create a subcription that should have schedule as  "On Extract Refresh"

**Results**
**Error When None is passed to Scheduled ID**:     new_sub = TSC.SubscriptionItem(email_subject, schedule_id=**None**, user_id, target)
 INFO Creating a subscription (<Subscription subject(Test_User_TestSubscriptions2_withoutschedule) schedule_id(None) user_id(f57ca8cb-4a14-4b01-ac7b-721ffe260537)                 target(<Target#761209c1-decd-4626-8bbb-7be632e2efe8, view>))
 Failed to create subscription. below is the error message
INFO cannot serialize None (type NoneType)

**Error When Schedule_Id parameter is skipped**:  new_sub = TSC.SubscriptionItem(email_subject,  user_id, target)
SubscriptionItem.__init__() missing 1 required positional argument: 'schedule_id'

please suggest if I'm missing any

Issue #1537, opened 2024-11-23
Flow image unavailable when using flows publish method


CodingOnWindowsOS:
**Describe the bug**
When using the flows publish method, the flow image is not available on Tableau Cloud. This occurs when using a .tfl and .tflx file. The image is available when manually published through Tableau prep using either file type. A flow image is a diagram of the steps created to produce a desired output. This is what the user sees when creating the flow.

**Versions**
Details of your environment, including:
 - Tableau Cloud
 - Python 3.12.3
 - TSC library version 3.24

**To Reproduce**
`server.flows.publish(
	flow_item=new_flow,
	file=Path('some_flow.tfl'),
	mode='CreateNew'
)`

**Results**
Flow image unavailable

CodingOnWindowsOS:
After downloading the flow from the server manually and inspecting the unzipped contents, I noticed an additional image called flowGraphImage. This image file seems to be produced by the Server / Cloud instance and is not produced when creating and publishing a flow from a local machine, so when I attempt to publish, the image is not available.
![flowGraphImage](https://github.com/user-attachments/assets/a76acdfb-aa31-4957-b3ed-9f6f67306fd8)

This lines up with another observation I had while editing the flow "in-place" on the server. I noticed that if I edited the flow with the flow image not available issue and clicked republish, then the flow image would be produced. Seems like something that is done server side if I have connected to dots correctly.

Is there anyway to trigger this when using the flows endpoint publish method?

jacalata:
It looks like the image can be generated, but when published through the UI it is done separately and this step doesn't happen with the API call. 

I have logged a bug internally for the relevant team to look into, and hopefully they can make these behave the same way.

CodingOnWindowsOS:
Thank you for the support!

jacalata:
Quick update: I reproduced this against Tableau Cloud (2026-07-30, server v3.29) and filed an internal bug against the Tableau Prep Engine team. Confirmed on the current release: publishing `test/assets/SampleFlow.tfl` via `server.flows.publish()` and downloading it back shows `displaySettings, maestroMetadata, flow, flowGraphThumbnail.svg` — no server-generated thumbnail. A UI-published flow on the same site downloads with an additional `flowGraphImage.png` in the bundle, which is the file the flow's overview page actually renders.

One correction from the original report: the missing file is `flowGraphImage.png`, not `flowGraphThumbnail.svg` (the SVG is a local file that's always present). Bug filed with that correction.

[...]

Issue #1508, opened 2024-10-23
Extract Refresh task with a Daily schedule don't work as expected


chroche:
**Describe the bug**

When creating an Extract Refresh task with a Daily schedule following the provided [example](https://github.com/tableau/server-client-python/blob/v0.33/samples/create_extract_task.py), I either get an "Error 500" from the server or a task that's running every hour instead of once a day.

**Versions**
Details of your environment, including:
 - Tableau Server version: Tableau Cloud 2024.3.0 (20243.24.1011.1935) 64-bit Linux
 - Python version: 3.11.10
 - TSC library version: 0.33

**To Reproduce**
When following the example code above with a Daily interval as described in [this sample](https://github.com/tableau/server-client-python/blob/v0.33/samples/create_schedules.py#L64)
```python
  tableau_auth   = TSC.PersonalAccessTokenAuth(TOKEN_NAME, api_token, SERVER_SITE)
  tableau_server = TSC.Server(SERVER_URL, use_server_version=False)
  tableau_server.version = '3.24'

  with tableau_server.auth.sign_in(tableau_auth):
    workbook = TSC.WorkbookItem(name=WORKBOOK_NAME, project_id=project)

    daily_interval = TSC.DailyInterval(start_time=datetime.time(hour=5, minute=40, tzinfo=datetime.UTC))
    daily_schedule = TSC.ScheduleItem(None, None, None, None, daily_interval)
    target_item    = TSC.Target(workbook.id, 'workbook')
    extract_item   = TSC.TaskItem(None, 'FullRefresh', None, None, None, daily_schedule, None, target_item)
    tableau_server.tasks.create(extract_item)
```
I'm getting he following error:
```
Internal error 500 at https://eu-west-1a.online.tableau.com/api/3.24/sites/288c88d2-8f35-4faa-b71c-bc082f0ded3c/tasks/extractRefreshes
b'<?xml version=\'1.0\' encoding=\'UTF-8\'?><tsResponse xmlns="http://tableau.com/api" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://tableau.com/api https://help.tableau.com/samples/en-us/rest_api/ts-api_3_24.xsd"><error code="500000"><summary>Internal Server Error</summary><detail>The server encountered an error and cannot complete your request. Contact your server administrator.</detail></error></tsResponse>'
```

Now if instead I declare the list of weekdays in the Daily interval:
```python
[...]

Issue #1431, opened 2024-07-29
Docs clarification


jorwoods:
In [the docs for filtering the REST API](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_filtering_and_sorting.htm) is the following phrase:

> Wildcard searches for characters using starts with, ends with, and contains-based searches are supported starting in Tableau Cloud May 2023 and Tableau Server 2022.1.14.

No field on any object shows support for `contains`, `startswith`, or `endswith`.

Can we get some clarification on what objects and fields support these operators and the exact syntax used?

kjmassey:
What Jordan said! 😁

jacalata:
Filing this against the Tableau REST API docs team internally rather than keeping it open here — the docs are owned outside this repo. Closing as "not planned" for TSC; the documentation fix is being tracked by the docs team.

PR #1842, opened 2026-07-30
docs: add Known server-side limitations section to api-ref.md

jacalata:
## Motivation

Users hitting a REST API limitation often assume TSC itself is the
problem and file an issue here. `api-ref.md` had no place to list gaps
that are actually server-side. Add a "Known server-side limitations"
section that points readers at the tracking issue for each gap and at
the `Server-Side Enhancement` label as the live index of the same.

Paired with #1841 which stops the stale bot from closing these
tracker issues.

## Behavior change

None to the runtime library. Adds ~43 lines to `docs/api-ref.md` on the
`gh-pages` docs branch:

- New "Known server-side limitations" section at the end of the file,
  seeded with five current gaps
- One-sentence cross-link from the top-of-page Note

Seed entries (each with its tracking issue where one exists):

- `vf_` view filter query parameters support exact match only -- no
  operator prefixes, wildcards, or ranges. Tracked in #1431.
- `vf_` view filter is silently dropped when the server-side view
  already has an operator-based filter on the same field. No tracking
  issue yet.
- Daily extract-refresh schedules cannot be created via the REST API on
  Tableau Cloud (returns 500). Tracked in #1508.
- Subscriptions cannot use "On Extract Refresh" schedule via the REST
  API. Tracked in #1658.
- Flow thumbnail image is not generated when a flow is published via
  the REST API. Tracked in #1537.

## Test plan

- [ ] Preview the rendered `api-ref.md` on the gh-pages Jekyll site
- [ ] TOC entry for the new section renders (file uses `* TOC` auto-generator)
- [ ] Cross-link `#known-server-side-limitations` from top-of-page note resolves
- [ ] Each of the five GitHub issue links opens the expected issue
- [ ] `Server-Side Enhancement` label link resolves

## Relationship to #1835

#1835 also touches `api-ref.md` and adds inline caveats near the
request-options docs. This PR adds a new section at the end plus one
cross-link in the top-of-page note. The two edits target different
regions of the file so should not conflict; a maintainer reviewing
both should confirm merge order.

### DIFF
--- a/docs/api-ref.md
+++ b/docs/api-ref.md
@@ -10965,32 +10965,32 @@ Some behaviors you may hit while using TSC are not bugs in the TSC library. They
 
 For the live, maintainer-updated view of these issues, filter open issues in this repository by the [`Server-Side Enhancement`](https://github.com/tableau/server-client-python/labels/Server-Side%20Enhancement) label.
 
-Each entry below notes where the fix is expected to land (server-side vs. client-side). Items marked *server-side* require a change in Tableau Server / Tableau Cloud before TSC can expose the behavior. Internal Salesforce work-item numbers (`W-...`) are included as breadcrumbs for Tableau maintainers; external users are not expected to be able to resolve them.
+Every item in this list requires a change in Tableau Server / Tableau Cloud before TSC can expose the behavior.
 
 <br>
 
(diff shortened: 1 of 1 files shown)


jacalata:
docs only change, merging with AI review only

PR #1853, opened 2026-08-05
Improve the stale actions config now that it's been running for a bit

bcantoni:
- increase operations-per-run beyond the default 30
- add permission for action to write its cache

Taken together, these should let the action run on the oldest issues and PRs we have.


bcantoni:
@jacalata for your review. I noticed the action is running, but isn't reaching very far back into the Issues backlog. I think these changes should help/fix that.

jacalata (approved):

PR #1862, opened 2026-08-16
Redact password column from CSV import logging and error output

### LINKED ISSUE #1829: security: _validate_import_line_or_throw logs credential fields at DEBUG level
## Problem

`UserItem.CSVImport._validate_import_line_or_throw` (and by extension `validate_file_for_import`) accepts a caller-supplied `logger` and logs validation details at DEBUG level. The raw CSV values passed in include the password column (column index 1) from user-import files.

If a caller passes a logger with DEBUG enabled — common in development or verbose CI environments — passwords from the import CSV are written to the log output.

## Proposed fix

Before any logging in `_validate_import_line_or_throw`, mask or omit the password field. For example, replace it with `"***"` in any debug-logged representation of the line:

```python
def _safe_log_line(values):
    masked = list(values)
    if len(masked) > UserItem.CSVImport.ColumnType.PASSWORD:
        masked[UserItem.CSVImport.ColumnType.PASSWORD] = "***"
    return masked
```

This is a narrow fix — it does not change validation logic, only what gets emitted to the log.

jacalata:
## Summary

Fixes two paths in `UserItem.CSVImport` that could leak the password column from a user-import CSV file:

1. `_validate_import_line_or_throw` logged `logger.info(f"Reading user {line[:4]}")`. The intent was to avoid printing the password, but slicing four characters only obscures usernames shorter than four characters -- longer usernames leak part of the row, and if the layout ever changed the password itself could reach the log.
2. `validate_file_for_import` returned raw lines in its `invalid_lines` list when a row failed validation. Any caller that logged or surfaced those lines would expose the password unmasked.
3. The per-column debug log inside `_validate_import_line_or_throw` printed the PASS column value verbatim.

Fixes #1829.

## What changes

- Row-level logs now include only the username (parsed by partition on the first comma) and are demoted to DEBUG.
- `invalid_lines` entries pass through a new `_redact_password_column` helper that replaces column 1 with `***`, preserving the line's original line ending.
- The per-column DEBUG log masks the PASS column with `***`.

The redaction helper uses the same naive `str.split(",")` parsing already used throughout the file. A password value containing commas is misaligned across columns; only the fragment landing in column 1 is masked. A test documents this limitation -- a proper CSV parser would be a larger change and is out of scope here.

## Test plan

- [x] Four new tests in `test/test_user_model.py` covering: DEBUG log on a valid row (positive-and-negative assertion so a fix that only deletes the log line would fail), invalid-row path (both the returned `invalid_lines` and the logger output), embedded-comma limitation, and unit-level edges on `_redact_password_column` (LF, CRLF, no newline, empty password, trailing comma, single column).
- [x] Full test suite passes locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,15 @@
 ## Unreleased
 
 * Security: `UserItem.CSVImport` no longer logs the password column when
-  validating a user-import CSV file. The password field was previously written
-  to any caller-supplied logger at DEBUG level, and the raw row was returned in
-  `validate_file_for_import`'s `invalid_lines` list unmasked. Fixes #1829.
+  validating a user-import CSV file. Previously, `validate_file_for_import`
+  emitted the first four characters of each raw row at INFO (which could
+  include the beginning of the password when the username was short), and the
+  full raw row was pushed into the returned `invalid_lines` list unmasked; the
+  per-column log inside `_validate_import_line_or_throw` also wrote the
+  password value at DEBUG. Row-level logging is now DEBUG-only and logs the
+  username instead of a raw slice, and the password column is replaced with
+  `***` before any line reaches a log handler or the invalid-lines list.
+  Fixes #1829.
 * Added `Projects.get_by_path(path)` to look up a project by its slash-separated
   hierarchy path (e.g. `"Marketing/Q1 Reports"`). The walk is performed level by
   level using the REST API name filter, so a path with *n* components issues *n*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


jacalata:
22934d4 addresses both Copilot findings:

- `_redact_password_column` now preserves the exact trailing sequence (`\r\n`, `\n`, or none) after redaction. Split-and-rejoin no longer downgrades CRLF to LF when the password is the last field.
- Regression test `"user,pass\r\n" -> "user,***\r\n"` added to `test_redact_password_column_helper`.

CHANGELOG updated separately: the pre-fix code logged the first four characters of each raw row at INFO, not DEBUG, so the entry now describes both the row-level INFO→DEBUG downgrade and the password-column redaction.

PR #1873, opened 2026-08-25
Add Schedules.get_subscriptions_by_schedule to list a schedule's subscriptions

### LINKED ISSUE #1250: [FUNCTION] [List Tasks in a Schedule]
## Summary
Would like to see the ability to list all tasks within a schedule on Tableau Server.

## Description
I have a use-case where I want to monitor tasks in specific schedules. This is so I can compare tasks in those schedules with a whitelist I maintain to ensure users aren't joining these schedules without getting approval first.

ChrisJr404:
## Motivation

Closes #1250. TSC wraps extract refresh tasks per schedule with `get_extract_refresh_tasks(schedule_id)`, but there was no way to ask "which subscriptions run on this schedule?", which is what the issue asks for (monitoring the tasks in a schedule against a whitelist).

I went looking for a `/schedules/{id}/subscriptions` endpoint to mirror the extracts one, but the REST API doesn't have it. Subscriptions are only listable at the site level via [Query Subscriptions](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_subscriptions.htm#query_subscriptions), and each subscription already carries its schedule id. So this filters that list client side.

## Behavior change

Adds one public method:

```python
subs = server.schedules.get_subscriptions_by_schedule(schedule_id)  # -> list[SubscriptionItem]
```

It pages through the site subscriptions and returns the ones whose `schedule_id` matches, so on a busy site it costs one request per page. Nothing else changes. I put it on `Schedules` next to `get_extract_refresh_tasks` so the two task-listing calls live together. Because there's no server-side filter, this is a plain `list`, not a `(list, PaginationItem)` tuple like `get()`. Documented the per-page cost and the "no per-schedule endpoint" caveat in the docstring, and added a CHANGELOG entry.

## Test plan

- `test_get_subscriptions_by_schedule` reuses the existing `subscription_get.xml` asset (its two subscriptions are on different schedules) and asserts only the matching one comes back.
- `test_get_subscriptions_by_schedule_no_match` asserts an unknown schedule id yields `[]`.
- `test_get_subscriptions_by_schedule_empty_id` asserts an empty id raises `ValueError`, matching the other schedule methods.
- Full suite: 908 passed, 1 skipped. `black --check` and `mypy` clean.

One thing on trust rather than a regression test: I couldn't hit a live server, so the client-side-filter approach rests on the documented fact that Query Subscriptions returns every subscription with its `<schedule id=...>`, which the existing `SubscriptionItem` parsing already relies on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 
 ## Unreleased
 
+* Added `Schedules.get_subscriptions_by_schedule(schedule_id)` to list the
+  subscriptions attached to a server schedule. The REST API has no per-schedule
+  subscriptions endpoint (only extract refresh tasks have one), so this pages
+  through the site's subscriptions and returns those whose `schedule_id`
+  matches, mirroring `get_extract_refresh_tasks`. Closes #1250.
 * Added `Projects.get_by_path(path)` to look up a project by its slash-separated
   hierarchy path (e.g. `"Marketing/Q1 Reports"`). The walk is performed level by
   level using the REST API name filter, so a path with *n* components issues *n*
--- a/tableauserverclient/server/endpoint/schedules_endpoint.py
+++ b/tableauserverclient/server/endpoint/schedules_endpoint.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any, Callable, Literal, overload
 from .endpoint import Endpoint, api, parameter_added_in
 from .exceptions import MissingRequiredFieldError
 from tableauserverclient.server import RequestFactory
+from tableauserverclient.server.pager import Pager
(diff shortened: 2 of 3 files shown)


jacalata:
Unfortunately subscriptions look different in Tableau Server (onprem) and Tableau Cloud, and don't have a schedule id on Cloud - that just means your method won't work for Cloud. I have a fix to at least handle the different shapes cleanly in #1875, but if you could add a docstring mention that this is Server only that would probably do. I can merge in my fix afterwards and check that it responds nicely to both kinds. 
Tests look good, just need CLA signed.

jacalata (changes requested):
fix merge conflict (just keep both) and add a docstring saying this is Server only.

Issue #1336, opened 2024-01-13
Replace custom time handling with datetime/timedelta


jacalata:
From #1299: 

The use of a custom datetime module caught me off guard in the project.
I'd suggest using datetime and timedelta as the source for seconds and minutes instead of manually keeping track of what the seconds and minutes ought to be, but didn't want to make any more changes than necessary to fix the problem.

jacalata:
Would be nice to have — PRs welcome to swap the raw `time.time()` / `time.sleep()` in `exponential_backoff.py` for `datetime.timedelta`.

jacalata:
Adding a datapoint from the recent cloud-schedule parsing work in #1875: `datetime_helpers.parse_datetime` had to grow a second `strptime` branch to accept the Cloud form `2026-08-29T16:55:00-0700` alongside Server's `...Z`.

`datetime.fromisoformat` would collapse both into a one-liner — but not until we drop 3.10:
- Python **3.11+** accepts both `...Z` and unpunctuated offsets like `-0700` natively.
- Python **3.10** rejects `Z` and requires punctuated offsets (`-07:00`), so the swap would still need a preprocessing shim — no cleaner than the current two-strptime pattern.

Once `requires-python` bumps past 3.10 (EOL 2026-10), `parse_datetime` becomes `datetime.fromisoformat(s)` and `TABLEAU_CLOUD_DATE_FORMAT` / the `try/except ValueError` fallback can be deleted. Flagging so it doesn't get lost.

jacalata:
Closing as won't-fix rather than stale. On closer inspection (see #1876) the refactor turns out to be cosmetic — `(datetime.now() - self.start_time).total_seconds()` works the same way as `time.time() - self.start_time`, nothing downstream uses timedelta types, both APIs are equally subject to wall-clock jumps. Sorry for the earlier "PRs welcome" mixed signal.

PR #1876, opened 2026-08-27
refactor: use datetime/timedelta for exponential backoff elapsed time

TrueFurina:
## Summary

Per #1336 (and the maintainer's suggestion in the thread): swaps the raw `time.time()` / monotonic-elapsed arithmetic in `tableauserverclient/exponential_backoff.py` for `datetime`/`timedelta`.

## Changes

- `tableauserverclient/exponential_backoff.py`:
  - `self.start_time = time.time()` → `self.start_time = datetime.now()`
  - `elapsed = time.time() - self.start_time` → `elapsed = (datetime.now() - self.start_time).total_seconds()`
  - `time.sleep(...)` retained (sleep still needs seconds)

Behavior is unchanged: `elapsed` is still a float number of seconds, timeout comparison and error message are identical.

## Verification

- Syntax check passed
- Timer still sleeps with backoff and raises `TimeoutError` when the deadline passes (verified locally with a 0.01s timeout)

### DIFF
--- a/tableauserverclient/exponential_backoff.py
+++ b/tableauserverclient/exponential_backoff.py
@@ -1,3 +1,4 @@
+from datetime import datetime, timedelta
 import time
 
 # Polling for server-side events (such as job completion) uses exponential backoff for the sleep intervals between polls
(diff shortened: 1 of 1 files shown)


jacalata:
Hi @TrueFurina — apologies, my earlier "PRs welcome" note on #1336 was too optimistic. Looking closely: `(datetime.now() - self.start_time).total_seconds()` works the same way as `time.time() - self.start_time`; nothing downstream uses timedelta types; both are equally subject to wall-clock jumps. Closing this and updating the issue. Really sorry for the wasted effort.

